PR #17, opened 2017-08-01
Run scripts

### LINKED ISSUE #7: Run scripts
Generate pre and post build run script build phases

yonaskolb:
Resolves #7 

This adds **prebuildScripts** and **postbuildScripts** to targets

### DIFF
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -42,15 +42,6 @@ public class PBXProjGenerator {
         projectReference = generateUUID(PBXProject.self, spec.name)
     }
 
-    func getPath(_ path: String) -> Path {
-        let currentPath = Path(path)
-        if currentPath.isRelative {
-            return basePath + currentPath
-        } else {
-            return currentPath
-        }
-    }
-
     public func generateUUID<T: ProjectElement>(_ element: T.Type, _ id: String) -> String {
         var uuid: String = ""
         var counter: UInt = 0
@@ -141,7 +132,7 @@ public class PBXProjGenerator {
 
     func generateTarget(_ target: Target) throws -> PBXNativeTarget {
 
-        let sourcePaths = target.sources.map { getPath($0) }
+        let sourcePaths = target.sources.map { basePath + $0 }
         var sourceFilePaths: [Path] = []
 
         for source in sourcePaths {
@@ -154,7 +145,7 @@ public class PBXProjGenerator {
             var baseConfigurationReference: String?
 
             if let configPath = target.configFiles[config.name] {
-                let path = getPath(configPath)
+                let path = basePath + configPath
                 baseConfigurationReference = fileReferencesByPath[path]
             }
             return XCBuildConfiguration(reference: generateUUID(XCBuildConfiguration.self, config.name), name: config.name, baseConfigurationReference: baseConfigurationReference, buildSettings: buildSettings)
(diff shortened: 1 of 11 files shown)


Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -223,6 +232,32 @@ public class PBXProjGenerator {
...
+                shellScript = script
+            }
+            let escapedScript = shellScript.replacingOccurrences(of: "\"", with: "\\\"").replacingOccurrences(of: "\n", with: "\\n")
+            let shellScriptPhase = PBXShellScriptBuildPhase(
+                reference: generateUUID(PBXShellScriptBuildPhase.self, String(describing: runScript.name) + shellScript),
+                files: [],
+                name: runScript.name ?? "Run Script",
+                inputPaths: Set(runScript.inputFiles),
+                outputPaths: Set(runScript.outputFiles),
+                shellPath: runScript.shell ?? "/bin/sh",
```
pepibumur: What about making `shell` non-optional and using a default value in the `RunScript` constructor with `/bin/sh`. Since the shell is required for creating a `PBXShellScriptBuildPhase` I'd make it non-optional here as well. What do you think?

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -223,6 +232,32 @@ public class PBXProjGenerator {
...
+        func getRunScript(runScript: RunScript) throws -> PBXShellScriptBuildPhase {
+
+            let shellScript: String
+            switch runScript.script {
+            case let .path(path):
+                shellScript = try getPath(path).read()
+            case let .script(script):
+                shellScript = script
+            }
+            let escapedScript = shellScript.replacingOccurrences(of: "\"", with: "\\\"").replacingOccurrences(of: "\n", with: "\\n")
```
pepibumur: Wouldn't you extract this into an extension or a method that is more explicit about why this is necessary?

Sources/XcodeGenKit/ProjectGenerator.swift
```diff
@@ -33,7 +33,16 @@ public class ProjectGenerator {
         return spec.configs.first { $0.type == .release }!
     }
 
-    func validate() throws {
+    func getPath(_ path: String) -> Path {
```
pepibumur: There's a similar method in `PBXProjGenerator`. Have you consider extracting it somewhere to use it from both places?

docs/ProjectSpec.md
```diff
@@ -146,8 +146,42 @@ targets:
       Release: config_files/release.xcconfig
 ```
 
+#### Run Scripts
```
pepibumur: It's great that you document all these things, very useful 👏

Sources/XcodeGenKit/PBXProjGenerator.swift
yonaskolb: Yeah that would be an option. I've been trying to keep any specific knowledge out of ProjectSpec though, and just keep it dumb. 
Also I think that `/bin/sh` should be the default value in the `xcodeproj` constructor, and I didn't want to double up work there. It's something we still need to add to xcodeproj though

Sources/XcodeGenKit/PBXProjGenerator.swift
yonaskolb: You're right, this should go into PXBWriter, but I just put it here for now to get it working

Sources/XcodeGenKit/ProjectGenerator.swift
yonaskolb: Yeah, it felt weird to add it twice. Maybe an extension on Path

yonaskolb: I just started to add an extension of `Path`, but found out that appending paths already checks for absolute paths, so this was all unnecessary :)

PR #37, opened 2017-08-26
Dependency embed options

### LINKED ISSUE #27: Add Frameworks as Embedded Binaries
I often use Swift dynamic framework and specify them in `xcodegen.yml` like below:

```
name: Foo
targets:
  - name: Bar
    type: application
    platform: iOS
    sources: 
      - Sources
    dependencies:
      - framework: Frameworks/Foo.framework
      - framework: Frameworks/Bar.framework

```

XcodeGen correctly handle this and add those frameworks to `Linked Frameworks and Libraries`, but it seems we do not have an option to have them as `Embedded Binaries`?

yonaskolb:
Fixes #27

This adds a `embed` `codeSign` and `removeHeaders` properties on Dependency. 
`embed` has different defaults depending if it's set on a app target or not. See ProjectSpec.MD for details

### DIFF
--- a/Sources/ProjectSpec/Target.swift
+++ b/Sources/ProjectSpec/Target.swift
@@ -194,31 +194,44 @@ extension Target: JSONObjectConvertible {
     }
 }
 
-public enum Dependency: Equatable {
+public struct Dependency: Equatable {
 
-    case target(String)
-    case framework(String)
-    case carthage(String)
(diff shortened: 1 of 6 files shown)


ImLaufderZeit:
`codeSign` doesn't seem to be working, even with manually specifying `codeSign: true`.

PR #44, opened 2017-08-30
Project Spec includes

### LINKED ISSUE #22: Spec Includes
Ability to support spec includes. This would allow you to seperate the spec into multiple files, to for example seperate all the setting presets for sharing amongst multiple specs

yonaskolb:
Resolves #22 

allows for include property

```yaml
name: MyCoolSpec
include: [base.yml, settings.yml]
```


pepibumur:
Hey @yonaskolb. I'm not sure if I get the purpose of this property. What's the use case?

yonaskolb:
@pepibumur, it means you can reference other spec files within a spec, and they will be loaded and merged into the spec.

PR #48, opened 2017-09-08
Add support to C and C++ files in ‘Compile sources’

antoniocasero:
Previously the Compile Sources section was empty because it was not able
to recognize the file extension.


yonaskolb:
Fantastic! Thanks @antoniocasero, good catch!

PR #50, opened 2017-09-11
Filter out NSNull elements from the json dictionary.

antoniocasero:
This change allows overriding default BuildSettings.

For example:

If I want to keep the attribute SWIFT_OBJC_INTERFACE_HEADER_NAME empty,
(by default it contains some value). I just need to override this value
in the settings spec.

Previously, If I keep property empty, the generated settings value
appear as <null>.

### DIFF
--- a/Sources/XcodeGenKit/SpecLoader.swift
+++ b/Sources/XcodeGenKit/SpecLoader.swift
@@ -16,7 +16,8 @@ public struct SpecLoader {
 
     public static func loadSpec(path: Path) throws -> ProjectSpec {
         let dictionary = try loadDictionary(path: path)
-        return try ProjectSpec(jsonDictionary: dictionary)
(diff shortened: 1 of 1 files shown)


yonaskolb:
Thanks for this @antoniocasero! Yeah there is a known issue with YAMS, and it decoding `""` as Null. I've actually filed a bug here https://github.com/jpsim/Yams/issues/53. If they are open to it I'd like to patch it there.

In terms of fixing it in XcodeGen in the meantime, I'd like to address it at a higher level, perhaps in `SpecLoader`. Are you able to move the fix there, and make it recursively replace all the null values with empty strings?

yonaskolb:
@antoniocasero, if you're interested, I've also just invited you to our `swift-xcode` slack group for this project and xcodeproj

antoniocasero:
Done. Let me know what you think.

yonaskolb:
Looks like this will do the trick, thanks @antoniocasero!

PR #54, opened 2017-09-14
Change targets from list to map

### LINKED ISSUE #53: Change targets from list to map
At the moment `targets` is list. That was done to preserve the order of generated targets in the xcode project.
There would be some upsides to making it a map of names to targets however:
- the `include` feature could edit and merge targets by name
- more natural yaml styl
- enforce the target name
- better visual seperation

The points are very minor other than the first one which is a real consideration.
Is having the order of targets in the project worth it? Thoughts anyone?

Here is a visual difference

**list - Array**
```yaml
targets:
  - name: MyTarget
    platform: iOS
    type: application
  - name: MyFramework
    platform: iOS
    type: framework
```
**map - Dictionary**
```yaml
targets:
  MyTarget:
    platform: iOS
    type: application
  MyFramework:
    platform: iOS
    type: framework 
```

In either case `targets` in the `ProjectSpec` struct would still be an array, this just affects yaml

yonaskolb:
Resolves #53

### DIFF
--- a/Fixtures/TestProject/GeneratedProject.xcodeproj/project.pbxproj
+++ b/Fixtures/TestProject/GeneratedProject.xcodeproj/project.pbxproj
@@ -83,7 +83,7 @@
 			isa = PBXGroup;
 			children = (
 				FR7078510801 /* FrameworkFile.swift */,
-				FR1345298502 /* Info.plist */,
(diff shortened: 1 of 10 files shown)


yonaskolb:
I'll merge this just before the next release, so people don't get confused if they installed the old version and are looking for documentation

Issue #51, opened 2017-09-13
Brew install does not currently work


daniel-beard:
Using the instructions from the README I currently get this: 

```
brew install xcodegen
==> Installing xcodegen from yonaskolb/xcodegen
==> Downloading https://github.com/yonaskolb/XcodeGen/archive/0.6.1.tar.gz
Already downloaded: /Users/dbeard/Library/Caches/Homebrew/xcodegen-0.6.1.tar.gz
==> Building XcodeGen
==> swift build -c release -Xlinker -rpath -Xlinker @executable_path -Xswiftc -static-stdlib
Last 15 lines from /Users/dbeard/Library/Logs/Homebrew/xcodegen/01.swift:
2017-09-13 12:02:33 -0700

swift build -c release -Xlinker -rpath -Xlinker @executable_path -Xswiftc -static-stdlib

error: manifest parse error(s):
sandbox-exec: sandbox_apply_container: Operation not permitted

If reporting this issue please do so to (not Homebrew/brew or Homebrew/core):
yonaskolb/xcodegen

/usr/local/Homebrew/Library/Homebrew/utils/github.rb:226:in `raise_api_error': curl failed!  (GitHub::Error)
curl: (22) The requested URL returned error: 422 Unprocessable Entity
curl: (3) <url> malformed
	from /usr/local/Homebrew/Library/Homebrew/utils/github.rb:184:in `open'
	from /usr/local/Homebrew/Library/Homebrew/utils/github.rb:284:in `search'
	from /usr/local/Homebrew/Library/Homebrew/utils/github.rb:231:in `search_issues'
	from /usr/local/Homebrew/Library/Homebrew/utils/github.rb:244:in `issues_for_formula'
	from /usr/local/Homebrew/Library/Homebrew/exceptions.rb:369:in `fetch_issues'
	from /usr/local/Homebrew/Library/Homebrew/exceptions.rb:365:in `issues'
	from /usr/local/Homebrew/Library/Homebrew/exceptions.rb:419:in `dump'
	from /usr/local/Homebrew/Library/Homebrew/brew.rb:133:in `rescue in <main>'
	from /usr/local/Homebrew/Library/Homebrew/brew.rb:26:in `<main>'
```

daniel-beard:
May be related to having Xcode 9 GM installed.

jalopezcar:
I had the same problem with xcode9, it works with xcode8

```
sudo xcode-select -s .../Xcode.app
```

[...]

PR #55, opened 2017-09-14
Fix the build issues when using the Xcode 9 GM.

ehyche:
This change fixes this issue: https://github.com/yonaskolb/XcodeGen/issues/51

The build issues include:

- KeyPath class was added in Swift 4, which collides with the KeyPath defined in JSONUtilies.
  This was fixed by disambiguating the type by specifying "JSONUtilities.KeyPath" instead of just "KeyPath".
- DecodingError was added in Swift 4, which collides with DecodingError defined in JSONUtilities.
  This was fixed by disambiguating the type by specifying "JSONUtilities.DecodingError" instead of just "DecodingError".
- There were several places where xcodeproj was expecint [String] and the code was calling it with Set<String>. Apparently this worked in Swift 3.2, but no longer works in Swift 4. So these places were corrected by:
  - Not creating a Set() when an array was needed
  - Calling .sorted() on the Set to produce an array

### DIFF
--- a/Sources/ProjectSpec/Decoding.swift
+++ b/Sources/ProjectSpec/Decoding.swift
@@ -13,7 +13,7 @@ import Yams
 
 extension Dictionary where Key: JSONKey {
 
-    public func json<T: NamedJSONDictionaryConvertible>(atKeyPath keyPath: KeyPath, invalidItemBehaviour: InvalidItemBehaviour<T> = .remove) throws -> [T] {
+    public func json<T: NamedJSONDictionaryConvertible>(atKeyPath keyPath: JSONUtilities.KeyPath, invalidItemBehaviour: InvalidItemBehaviour<T> = .remove) throws -> [T] {
         guard let dictionary = json(atKeyPath: keyPath) as JSONDictionary? else {
             return []
         }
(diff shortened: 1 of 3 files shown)


yonaskolb:
Thanks for the PR @ehyche! Unfortunately you're about an hour late! 😄 
The Swift 4 #52 has already been merged. 

In terms of the xcodeproj changes, that's because you must have linked to a new version as Swift 4 doesn't read `Package.pins` anymore so it upgraded itself. I noticed that as well and actually locked it to the previous version in that PR. I'm now going to do the xcodeproj update

ehyche:
@yonaskolb : Wow - I looked at the open PRs before I started mine, but I guess you had already merged by then!

By the way, I also noticed that Spectre also did not build on Swift4, so I submitted a PR against that project to fix it: https://github.com/kylef/Spectre/pull/29. However, it looks like there was already a PR open since June to fix the same issues. So I'm not sure if that project is still being maintained or not.

yonaskolb:
Spectre builds ok for me. There are some warnings in Swift 4 which is what I think you're referring to, but that's not a breaking issue for this repo 👍

yonaskolb:
The upgrade to xcodeproj 0.1.2 has also now been merged in #56

Issue #59, opened 2017-09-21
Support xcconfig


toshi0383:
# Motivation
Auto apply xcconfig configurations like this.
<img width="741" alt="xcode-configuration" src="https://user-images.githubusercontent.com/6007952/30695680-f4cd2ec8-9f13-11e7-922a-cc43b2d8c736.png">

# Current Problems
I notice documentations do mention about xcconfig but looks like it's not working as expected.

- Setting `configFiles` does not affect anything. (Bug? or maybe my spec is wrong)
- There is no way to set base xcconfigs for PBXProject.
- xcconfigs not being added to project

# Proposal
So this is my proposal.
- Support `baseConfiguration` for PBXProj
- Rename target's `configFiles` with `baseConfigurations` for consistency.
- (Optional) Rename `configs` to `configurations` since it looks like `xcconfig` and is misleading.

```project.yml
name: MyProject
configurations:
  Debug:
    baseConfiguration: configs/Debug.xcconfig
  OTA:
    baseConfiguration: configs/OTA.xcconfig
  Release:
    baseConfiguration: configs/Release.xcconfig
targets:
  MyApp:
    type: application
    platform: iOS
    sources: Sources/MyApp
    settings:
      PRODUCT_BUNDLE_IDENTIFIER: jp.toshi0383.hello
    baseConfigurations:
      Debug: configs/MyApp-Debug.xcconfig
      OTA: configs/MyApp-OTA.xcconfig
      Release: configs/MyApp-Release.xcconfig
```

If you guys feel happy with this outline, I can start working on a PR. 💪

yonaskolb:
Hi @toshi0383, thanks for the suggestions!

Yeah, you're right there's no way to specify a project's xcconfig file. Setting `configFiles` on a target should be working though. Is it not?

In terms of adding them to the project, I'd like to keep defining configurations and applying xcconfig files in seperate dictionaries. While it's not hugely documented at the moment the `include` feature is highly used by myself and others, and seperating them makes the feature work better. Also it's nice to keep consistency between a project and a target in how things are defined.

About the renaming, I agree maybe `configs` to `configurations` would be better. I'd like to keep `configFiles` though, but would be open to `configurationFiles`. In either case, both the project and target properties should match. 
Obviously both of these are breaking changes, so would need a major version bump, or still read the old values.

Feel free to get a PR up and going! 👍

toshi0383:
[...]

PR #60, opened 2017-09-21
WIP [cC]onfig => [cC]onfiguration

toshi0383:
- [x] Replace config to configuration. (Let me know if I'm going too far.😅)
- [x] Add deprecation message for old value usage.

Related: https://github.com/yonaskolb/XcodeGen/issues/59


Sources/XcodeGenKit/SettingsPresetFile.swift
```diff
@@ -11,15 +11,15 @@ import xcodeproj
...
+    case configuration(ConfigurationType)
     case platform(Platform)
     case product(PBXProductType)
     case productPlatform(PBXProductType,Platform)
     case base
 
     var path: String {
         switch self {
-        case let .config(config): return "Configs/\(config.rawValue)"
+        case let .configuration(configuration): return "Configurations/\(configuration.rawValue)"
```
yonaskolb: This would break unless the presets `config` folder was renamed

CHANGELOG.md
```diff
@@ -87,7 +87,7 @@
 - Added `carthageBuildPath` option [PR#34](https://github.com/yonaskolb/XcodeGen/pull/34)
 
 ### Fixed
-- Fixed installations of XcodeGen not applying build setting presets for configs, products, and platforms, due to missing resources
```
yonaskolb: references in the changelog shouldn't really be changed

yonaskolb:
Oh wow, you literally renamed every single instance of config in the project :)
I must say, I think I prefer `config` in code. It's just much easier to read and type.

At the moment in terms of user facing properties in the spec we have following:
- `project.configs`
- `project.configFiles` (not added yet)
- `target.configs`
- `target.configFiles`
- `settings.configs` (in project, target and setting groups)

These should be the only things that change (as well as the documentation for them).

In addition, I'm actually still on the fence about this change. Does anyone else here have any opinions?

toshi0383:
IMO `config` is very troublesome. New user might think `configFiles` is related to `config`.  Xcode or xcodebuild does not use `config` for shorthand name for `Configuration`. I think it's more user friendly to avoid potential confusion with xcconfig. Some user even don't know what `Configuration` is either, so we should be very very kind as possible.

> `Usage: xcodebuild [-project <projectname>] [[-target <targetname>]...|-alltargets] [-configuration <configurationname>] [-arch <architecture>]... [-sdk [<sdkname>|<sdkpath>]] [-showBuildSettings] [<buildsetting>=<value>]... [<buildaction>]...`

toshi0383:
I'm not planning to work on this one for now. (I got used to the code-base and I feel good.👌)

PR #64, opened 2017-09-23
XCConfig files

### LINKED ISSUE #59: Support xcconfig

yonaskolb:
Fixes #59 

- [x] fix target `configFiles`
- [x] add project `configFiles`
- [x] read xcconfig files while doing build setting lookups
- [x] add `fileGroups` to project for including non build files (like configs that don't exist in any target sources)
- [x] documentation


yonaskolb:
@toshi0383 this should fix your xcconfig issues. Note that this is still using the old names

yonaskolb:
`project.groups` and `project.settingGroups` might clash. They might be renamed as well as configs. Any suggestions? `settingGroups` used to be called `settingPresets` actually

Sources/XcodeGenKit/SettingsBuilder.swift
```diff
@@ -57,15 +57,19 @@ extension ProjectSpec {
...
     }
 
-    // combines all levels of a target's settings
+    // combines all levels of a target's settings: target, target config, project, project config
     public func getCombinedBuildSettings(basePath: Path, target: Target, config: Config, includeProject: Bool) -> BuildSettings {
         var buildSettings: BuildSettings = [:]
         if includeProject {
+            if let configFilePath = configFiles[config.name] {
+                if let configFile = try? XCConfig(path: basePath + configFilePath) {
+                    buildSettings += configFile.flattenedBuildSettings()
```
toshi0383: Maybe including xcconfig should just `#include` the included xcconfig, not adding to buildSettings section.🤔
What do you think?

toshi0383: Oh and what if included xcconfig `#include` other xcconfigs?

yonaskolb: Not sure what you mean by the first point? Maybe these will answer your question.
- this function is just use for getting all the combined settings for the purposes of checking what build settings will be applied - similar to Xcode's `combined` settings view
- xcconfigs aren't added using #include they are just set as a baseconfiguration for the target or project. The rest of the build settings are applied directly on the build setting or project and don't use an xcconfig

For the second point `configFile. flattenedBuildSettings()` goes through all #include files and merges them 👍

toshi0383: Understood. I'm going to take a look at the details.👌

Fixtures/TestProject/TestProject.xcodeproj/project.pbxproj
```diff
@@ -102,9 +104,20 @@
 			path = TestProjectTests;
 			sourceTree = "<group>";
 		};
+		CC9392641F76EDE600C1934A /* Configs */ = {
```
toshi0383: How did you create this `TestProject.xcodeproj/project.pbxproj` ?
On Xcode, `Configs` group is missing reference (appears red).

Sources/XcodeGenKit/SettingsBuilder.swift
toshi0383: > For the second point configFile. flattenedBuildSettings() goes through all #include files and merges them 👍

What I meant was `XcodeGen:Include`, I but I see `XcodeGen:Include` just ignores `XcodeGen:Include`ed spec's `configFile`, so it's no problem.

```
TestProject $ head spec.yml
include: included.yml
name: GeneratedProject
groups:
  - Configs
targets:
  TestProject:
    type: application
    platform: iOS
    sources: TestProject
    settings:
TestProject $ head included.yml
configFiles:
  Debug: Included/Configs/included.xcconfig  # This is completely ignored when `XcodeGen:included`.
```
This behavior is perfectly 🙆‍♂️ to me. Maybe it worth it to warn user when `XcodeGen:included` spec has configFile attribute, because it's ignored.

Maybe we should really rename `XcodeGen:include` to something else to avoid confusion. 🙄

toshi0383:
> project.groups and project.settingGroups might clash. They might be renamed as well as configs. Any suggestions? settingGroups used to be called settingPresets actually

I've read the documentation.
Looks like it's useful feature, too. I think it should be `buildSettings` and `buildSettingsGroup` to be explicit.

Sources/XcodeGenKit/SettingsBuilder.swift
yonaskolb: That's because include is a list not just a string. I should change this though to allow a simple string for a single value, just like a target's sources:
```
include: [included.yml] #works, using json style
include:
  - included.yml #works, using yaml style
include: included.yml #does work (yet, will change)
```

Fixtures/TestProject/TestProject.xcodeproj/project.pbxproj
yonaskolb: The `TestProject.xcodeproj` was created and edited in Xcode. `GeneratedProject.xcodeproj` is the one generated by XcodeGen.

I'm not sure what you mean by missing configs group. Both projects appear fine for me

docs/ProjectSpec.md
```diff
@@ -29,9 +29,11 @@ Required properties are marked 🔵 and optional properties with ⚪️.
 - ⚪️ **options**: [Options](#options) - Various options to override default behaviour
 - ⚪️ **attributes**: `map` - The PBXProject attributes. This is for advanced use. Defaults to ``{"LastUpgradeCheck": "0900"}``
 - ⚪️ **configs**: [Configs](#configs) - Project build configurations. Defaults to `Debug` and `Release` configs
+- ⚪️ **configFiles**: [Config Files](#config-files) - `.xcconfig` files per config
 - ⚪️ **settings**: [Settings](#settings) - Project specific settings. Default base and config type settings will be applied first before any settings defined here
 - ⚪️ **settingGroups**: [Setting Groups](#setting-groups) - Setting groups mapped by name
 - ⚪️ **targets**: [Target](#target) - The list of targets in the project mapped by name
+- ⚪️ **fileGroups**: `[String]` - A list of paths to add to the top level groups. These are files that aren't build files but that you'd like in the project heirachy. For example a folder xcconfig files that aren't already added by any target sources.
```
pepibumur: there's a typo `heirachy` ~> `hierarchy`

Sources/XcodeGenKit/SettingsBuilder.swift
yonaskolb: `include` can now be a single string https://github.com/yonaskolb/XcodeGen/commit/39c8af324c88f0e43f7131c5791f718a1eb29de6

Fixtures/TestProject/TestProject.xcodeproj/project.pbxproj
yonaskolb: I just saw what you mean by the red group, I've fixed this in master. This project isn't really used in any tests and is more just for a comparison between a project created in xcode versus created in XcodeGen

PR #67, opened 2017-09-24
Add bundleIdPrefix

### LINKED ISSUE #62: Generate bundle identifier
Hello,
Thanks for your this great solution that I really wanted.

Today I tried to generate my project using XcodeGen.
I faced the problem that Xcode shows the message `This app was unable to be installed` after build the project for the simulator.

I reproduced it on TestProject in this repository.

1. `xcodegen --spec spec.yml`
2. Open `GeneratedProject.xcodeproj`
3. Select the simulator like iPhone 8 Plus
4. Run

It can be fixed by removing frameworks from Copy Files in Build Phases or remove the line https://github.com/yonaskolb/XcodeGen/blob/master/Sources/XcodeGenKit/PBXProjGenerator.swift#L236

I don't know this is correct way to fix it. But do frameworks need to be copied when these add as Embedded Framework?

yonaskolb:
Fixes #62 
This is used for automatically generating PRODUCT_BUNDLE_IDENTIFIER if it's not specified


yonaskolb:
I just realized this could be solved without this feature
```
bundleIdPrefix: com.company
```
vs
```
settings: 
  PRODUCT_BUNDLE_IDENTIFIER: com.company.$(PRODUCT_NAME)
```

Is this feature, worth it then? Maybe not

kixswift:
@yonaskolb, I think it worth. This is something I wanted, too.
I'm currently using `variables` to achieve this like below:

```
projectName: &projectName ProjectName
bundleIdentifier: &bundleIdentifier jp.co.ClientName.ProjectName

name: *projectName
targets:
  *projectName:
    settings:
      base:
        PRODUCT_BUNDLE_IDENTIFIER: *bundleIdentifier
```

I always wanted to do something like this: `*bundleIdentifier.*projectName` but yaml doesn't seem to be able to do it.

`bundleIdPrefix: com.company`

This would be some 'unique' nice feature of XcodeGen. More clearer when you look at it.

yonaskolb:
@kixswift thanks for the feedback. You don't think this is simple enough?
```
settings: 
  PRODUCT_BUNDLE_IDENTIFIER: com.company.$(PRODUCT_NAME)
targets:
  MyTarget:
    platform: iOS
    type: application
  MyFramework:
    platform: iOS
    type: framework
```
MyApplication would get a PRODUCT_BUNDLE_IDENTIFIER of `com.company.MyTarget` and MyFramework would get `com.company.MyFramework`

yonaskolb:
Also maybe this should be moved to `project.options` like carthageBuildPath

kixswift:
> @kixswift thanks for the feedback. You don't think this is simple enough?

Um... well, actually you have a point and this might be just my personal preference. And I totally agree with you that this is an option.

yonaskolb:
Are there any other reasons why a bundleIdPrefix would be useful? Xcode makes you set one when creating a project and calls it the `Organisation Identifier`, but I don't think they use it for anything else then to prepend to bundle ids

kixswift:
> I like the idea of using the bundleID prefix. Having the prefix as an option will ensure that all the bundle identifiers are consistent across the targets.

@pepibumur, that's good point 🎉

Issue #69, opened 2017-09-25
Add localized resources to a variant group


ryohey:
If a storyboard localized for some language, it should be grouped into one variant group.

![2017-09-25 22 43 05](https://user-images.githubusercontent.com/5355966/30811447-30a55c72-a243-11e7-9cfd-fa5ea3cb47f9.png)

XcodeGen generates references same level. 

![2017-09-25 22 40 49](https://user-images.githubusercontent.com/5355966/30811462-391c8916-a243-11e7-9a49-692c01275d2f.png)

expectation

![2017-09-24 18 02 29](https://user-images.githubusercontent.com/5355966/30811559-877d08ba-a243-11e7-9c44-acc5afdf03a7.png)

```js
VG1473702401 /* Main.storyboard */ = {
	isa = PBXVariantGroup;
	children = (
		FR3676338402 /* Base */,
	);
	name = Main.storyboard;
	sourceTree = "<group>";
};
VG2365450901 /* Main.strings */ = {
	isa = PBXVariantGroup;
	children = (
		FR7070045401 /* en */,
	);
	name = Main.strings;
	sourceTree = "<group>";
};
```

should be

```js
VG1473702401 /* Main.storyboard */ = {
	isa = PBXVariantGroup;
	children = (
		FR3676338402 /* Base */,
		FR7070045401 /* en */,
	);
	name = Main.storyboard;
	sourceTree = "<group>";
};
```

ryohey:
fixed in #70

PR #70, opened 2017-09-25
Fix localisation variant groups

ryohey:
Add localized resources are added to one variant group. 
described in #69

1. Create variant groups of the base localization
2. Add references to localized resources into base localization variant groups


yonaskolb:
Thanks @ryohey, this is great!
Before I have a proper look at this, are you able to run `swift test` to generate the fixture project again and commit the changes (if there are any).

I actually just added a commit to master that fails the travis test if the generated fixture changes and wasn't committed, but it's after you created your branch

ryohey:
OK thanks!
By the way I want to write some tests for this changes. 
How do you think I can do that properly? (Is FixtureTests.swift good for that?)

yonaskolb:
Yeah, FixtureTests would be a good place. You can make a new Fixture project, but I think it would just be easier to edit the current `TestProject` and localise the storyboard

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -436,40 +436,60 @@ public class PBXProjGenerator {
...
+            allSourceFiles += subGroups.sourceFiles
             groupChildren.append(subGroups.groups.first!.reference)
             groups += subGroups.groups
         }
 
-        for filePath in filePaths {
-            let fileReference = getFileReference(path: filePath, inPath: path)
-            groupChildren.append(fileReference)
+        // create variant groups of the base localisation first
+        var baseLocalisationVariantGroups = [PBXVariantGroup]()
```
yonaskolb: minor: the rest of the codebase uses `var baseLocalisationVariantGroups: [String] = []` syntax

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -436,40 +436,60 @@ public class PBXProjGenerator {
...
+        if let baseLocalisedDirectory = localisedDirectories.first(where: { $0.lastComponent == "Base.lproj" }) {
+            for filePath in try baseLocalisedDirectory.children() {
+                let variantGroup = PBXVariantGroup(reference: generateUUID(PBXVariantGroup.self, filePath.lastComponent),
```
yonaskolb: You can pass the whole `path.string` into generateUUID so it's more deterministic

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -436,40 +436,60 @@ public class PBXProjGenerator {
...
+                                                   children: Set(),
```
yonaskolb: children can use []. This will also be changing to an array in the next xcodeproj version anyway https://github.com/swift-xcode/xcodeproj/issues/87

Sources/XcodeGenKit/PBXProjGenerator.swift
yonaskolb: What happens if more than 1 target has these files in it's sources? Do the variant groups need to be cached like the `groupsByPath` are?

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -436,40 +436,60 @@ public class PBXProjGenerator {
...
-                let variantGroup = PBXVariantGroup(reference: generateUUID(PBXVariantGroup.self, path.lastComponent), children: Set([fileReference.reference]), name: path.lastComponent, sourceTree: .group)
-                addObject(variantGroup)
+            for filePath in try localisedDirectory.children() {
 
-                fileReferencesByPath[path] = variantGroup.reference
-                groupChildren.append(variantGroup.reference)
-                allFilePaths.append(path)
+                // find base localisation variant group
+                let name = filePath.lastComponentWithoutExtension
+                guard let variantGroup = baseLocalisationVariantGroups.first(where: { Path($0.name).lastComponentWithoutExtension == name }) else {
```
yonaskolb: what if a localised file isn't inside a `Base.lproj`? Does this then not add the file reference anymore?

Sources/XcodeGenKit/PBXProjGenerator.swift
ryohey: OK!

Sources/XcodeGenKit/PBXProjGenerator.swift
ryohey: OK!

Sources/XcodeGenKit/PBXProjGenerator.swift
ryohey: Yes It will do not add the reference. It can be a problem in the case that resources placed in en.lproj but not in Base.lproj. For example a case an user removed resource in Base.lproj but not one in en.lproj.
I will fix it.

ryohey: If the project was created in the usual way, it looks like Base.lproj is included.

> The project folder should contain one folder named Base.lproj and other language-specific folders with the .lproj extension. The prefix for the language folders is the language ID, as described in Language and Locale IDs.

https://developer.apple.com/library/content/documentation/MacOSX/Conceptual/BPInternational/MaintaingYourOwnStringsFiles/MaintaingYourOwnStringsFiles.html

Sources/XcodeGenKit/PBXProjGenerator.swift
ryohey: That's right. I will add the cache for variant groups.

ryohey:
@yonaskolb I fixed issues you reviewed and I ran swift test then commit the updated project, but CI was failure. What should I do?

yonaskolb:
Thats interesting. If you've really committed the current version of the generated project I'm not sure why it's generating a different on on travis. 
One thing to make sure of is to not have the GeneratedProject open in Xcode, as Xcode will resave the project with some small formatting differences. If it's open close it and then run `swift test` again

ryohey:
@yonaskolb Thanks I resolved. I used full path to generate UUID. I changed it to remove basePath. Could you review it?

ryohey:
rebased onto master and resolve conflicts

ryohey:
@yonaskolb are there any problems?

yonaskolb:
Hi @ryohey, sorry I haven't had time to look at this in depth. I'll review it as soon as I have some time

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -471,34 +472,76 @@ public class PBXProjGenerator {
...
+                let buildFile = PBXBuildFile(reference: generateUUID(PBXBuildFile.self, variantGroup.reference), fileRef: variantGroup.reference, settings: nil)
+                allSourceFiles.append(SourceFile(path: path, fileReference: variantGroup.reference, buildFile: buildFile))
+            }
         }
 
+        // add references to localised resources into base localisation variant groups
         for localisedDirectory in localisedDirectories {
+            let localisationName = localisedDirectory.lastComponentWithoutExtension
             for path in try localisedDirectory.children().sorted { $0.lastComponent < $1.lastComponent } {
+                guard fileReferencesByPath[path] == nil else {
```
yonaskolb: What happens when multiple targets share localised sources that aren't in a `Base.lproj` directory? It looks like the second target won't get a build file, as this path is skipped

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -471,34 +472,76 @@ public class PBXProjGenerator {
...
+
                 let filePath = "\(localisedDirectory.lastComponent)/\(path.lastComponent)"
-                let fileReference = PBXFileReference(reference: generateUUID(PBXFileReference.self, localisedDirectory.lastComponent), sourceTree: .group, name: localisedDirectory.lastComponentWithoutExtension, path: filePath)
-                addObject(fileReference)
 
-                let variantGroup = PBXVariantGroup(reference: generateUUID(PBXVariantGroup.self, path.lastComponent), children: [fileReference.reference], name: path.lastComponent, sourceTree: .group)
-                addObject(variantGroup)
+                // find base localisation variant group
+                let name = path.lastComponentWithoutExtension
+                let variantGroup = baseLocalisationVariantGroups.first(where: { Path($0.name).lastComponentWithoutExtension == name })
```
yonaskolb: nitpick: could use trailing closure syntax here

Tests/XcodeGenKitTests/FixtureTests.swift
```diff
@@ -17,13 +17,47 @@ func generate(specPath: Path, projectPath: Path) throws {
...
+                return project.pbxproj.variantGroups.filter { $0.name == name }
+            }
+
+            let resourceName = "LocalizedStoryboard.storyboard"
+            let baseResource = "Base.lproj/LocalizedStoryboard.storyboard"
+            let localizedResource = "en.lproj/LocalizedStoryboard.strings"
+
+            guard let variableGroup = getVariableGroups(resourceName).first else { throw failure("Couldn't find the variable group") }
+
+            do {
```
yonaskolb: Is this just for visual grouping? Otherwise the `it` function is throwing so you don't need to wrap in a do

Sources/XcodeGenKit/PBXProjGenerator.swift
ryohey: You’re right. I will fix it.

Tests/XcodeGenKitTests/FixtureTests.swift
ryohey: Yes it's just for visual grouping.

Sources/XcodeGenKit/PBXProjGenerator.swift
ryohey: @yonaskolb I fixed this. https://github.com/yonaskolb/XcodeGen/pull/70/commits/859609158cbbe94930962c9c13848e3e7e46232c#diff-2f3c621620623f55e7ff42cc32198a10R540

ryohey:
I rebased this branch, but the code change is only 8596091, which fixed the problem pointed out in the review.

yonaskolb:
Nice one @ryohey! 🎉

keith:
I noticed this only seems to work if the default localization is in `Base.lproj`, not if you just have one in `en.lproj`. I'm not really sure if that's expected, or if it's worth solving for `*.lproj`, but it might be.

keith:
Actually looks like there might be another issue with this, I get this error on CI:

```
2017-10-17 14:19:02.197 xcodebuild[1357:7330] warning:  The file reference for "Base.lproj/Localizable.strings" is a member of multiple groups ("Localizable.strings" and "Localizable.strings"); this indicates a malformed project.  Only the membership in one of the groups will be preserved (but membership in targets will be unaffected).  If you want a reference to the same file in more than one group, please add another reference to the same path.```

yonaskolb:
@ryohey, are you able to take a look at this?

ryohey:
@keith If only one * .lproj other than Base.lproj is included in the project, It does not create a Variant Group but add a reference to the file directly, so I think it's not a problem.

However, if there are only en.lproj and ja.lproj, for example, they are added directly, so there is a problem.  It's necessary to disable Base Localization to fix the issue but the project currently output by XcodeGen is unconditionally enabled Base Internationalization by specifing Base in `knownRegions`. 
https://github.com/yonaskolb/XcodeGen/blob/09209a4e2c2cae8d57f6e43627260e8e667cd211/Sources/XcodeGenKit/PBXProjGenerator.swift#L126
(I just noticed that it might necessary to add the name of * .proj to knownRegions now)

If there is no base.lproj, it seems correct to use PBXProject's developmentRegion as Base Language, but developmentRegion is fixed to English, It may be necessary to adding syntax so that it can be specified developmentRegion from spec.
https://github.com/yonaskolb/XcodeGen/blob/09209a4e2c2cae8d57f6e43627260e8e667cd211/Sources/XcodeGenKit/PBXProjGenerator.swift#L133

As for the CI warnings please let me know your project's spec because I could not reproduced.

keith:
Here's an example project with the warning. You need to build with the newest commit of XcodeGen for it to work at all. Run `make` to see the warning output in the console.

[XcodeGenLocalizationWarnings.zip](https://github.com/yonaskolb/XcodeGen/files/1405163/XcodeGenLocalizationWarnings.zip)

ryohey:
@keith Thanks I got the warning.
fixed in https://github.com/yonaskolb/XcodeGen/pull/102

keith:
Awesome thanks so much!

PR #74, opened 2017-09-29
Fixed wrong directory path for macOS

toshi0383:
# before: Carthage/Build/macOS
screen shot 2017-09-30 at 0 05 30

# after: Carthage/Build/Mac
screen shot 2017-09-30 at 0 04 28

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Master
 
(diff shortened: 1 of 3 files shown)


yonaskolb:
Thanks @toshi0383! Good catch

toshi0383:
🎉

PR #76, opened 2017-09-29
No need carthage copy-frameworks for macOS app

toshi0383:


yonaskolb:
Interesting. I don't have much experience with macOS development so I missed this. It looks like we'd need to also have a script that copies the `dYSM` files as well for macOS targets. 
Do you feel up to adding that?

toshi0383:
I've never uploaded dSYMs for my Mac apps actually, so maybe not now. 🙃
I can investigate later, though. I'm going to make a new release for Swift4 anyways.(later this year..)

yonaskolb:
Ok, I'll merge this for now. About to do a hotfix release

PR #77, opened 2017-09-29
Avoid overwriting xcconfig's value by preset's value

toshi0383:
# Motivation

In this case below, preset's `SWIFT_VERSION` overwrites xcconfig's `SWIFT_VERSION` on Xcode.
Presets value shouldn't overwrite user-defined xcconfig's value.

### user-defined xcconfig
```
SWIFT_VERSION = 3.0
```

### SettingPresets/base.yml
```
SWIFT_VERSION: '4.0'
```

# Proposed Solution
Check if preset's value duplicates with xcconfig's and remove it from `buildSettings` if matches.


yonaskolb:
Hi @toshi0383 
I'd prefer not to go with this approach, as it differs from what Xcode does and deviates from expected behaviour. xcconfig files should not overwrite what is set in the project, as they are applied one inheritance level above.

This same issue would occur if you created a new project in Xcode and then tried to set the `SWIFT_VERSION` setting in an xcconfig (no matter if it's a target or project xcconfig).

If you want to change from the defaults then this can be set in the project build settings in the spec:
```
name: MyProject
settings:
  SWIFT_VERSION: '3.0'
targets: ...
```

yonaskolb:
Just to note, there is no "overwriting" currently happening. The current behaviour is just how Xcode handles the different levels of build settings. The order is

- application
- application xcconfig
- project
- project xcconfig
- sdk

The settings in `SettingPresets/base.yml` are applied to the project, before any custom settings you apply via the spec

toshi0383:
Thanks for review!

> This same issue would occur if you created a new project in Xcode and then tried to set the SWIFT_VERSION setting in an xcconfig (no matter if it's a target or project xcconfig).

Oh I see, you are trying to respect Xcode's behavior.
But I think XcodeGen is handling different situation. Let me explain.

### XcodeGen is handling different situations than "creating new project"
Having presets value like `SWIFT_VERSION = 4.0;` is default behavior of "creating new project", and it's okay to follow as long as you don't use xcconfigs, because newly created project don't have xcconfig files.
But when you use xcconfig files, you will have to remove buildSettings from pbxproj anyways. So IMO, spec with `configFiles` is not the same situation as "creating new project".

### Allow to define every buildSettings in xcconfig.
Other than that, I prefer to have every buildSettings defined in xcconfig, not in specs. It's easier to share with other projects.

### Summary
So those are the reasons why we should go with this PR's approach.
That's said, that `SettingPresets/base.yml` should be treated as somewhere between `project xcconfig` and `sdk`, not `project` level.

Please think about this again when you have time. 😉

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -70,10 +70,18 @@ public class PBXProjGenerator {
         }
 
         let buildConfigs: [XCBuildConfiguration] = spec.configs.map { config in
-            let buildSettings = spec.getProjectBuildSettings(config: config)
+            var buildSettings = spec.getProjectBuildSettings(config: config)
             var baseConfigurationReference: String?
             if let configPath = spec.configFiles[config.name] {
                 baseConfigurationReference = getFileReference(path: basePath + configPath, inPath: basePath)
+                // Do not overwrite xcconfig's value.
```
yonaskolb: Can this be moved into SettingsBuilder

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -173,6 +181,26 @@ public class PBXProjGenerator {
...
+                if let configFile = try? XCConfig(path: basePath + configPath) {
+                    for (k, _) in configFile.flattenedBuildSettings() {
+                        // FIXME: Catch platform specifier. e.g. LD_RUNPATH_SEARCH_PATHS[sdk=iphone*]
+                        buildSettings.removeValue(forKey: k)
+                        buildSettings.removeValue(forKey: "\"\(k)\"")
+                    }
+                }
+            }
+            // Do not overwrite base xcconfig values.
+            if let configPath = spec.configFiles[config.name] {
```
yonaskolb: I don't think we have to remove target build settings because of a project config file

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -173,6 +181,26 @@ public class PBXProjGenerator {
 
         let configs: [XCBuildConfiguration] = spec.configs.map { config in
             var buildSettings = spec.getTargetBuildSettings(target: target, config: config)
+            // Do not overwrite target's xcconfig values.
+            if let configPath = spec.targets.flatMap({ $0 == target ? $0.configFiles[config.name] : nil }).first {
```
yonaskolb: Can this be moved into SettingsBuilder. 
Also is there a reason this can't just be `if let configPath = target.configFiles[config.name]`. Maybe I'm missing something

Sources/XcodeGenKit/SettingsBuilder.swift
```diff
@@ -67,14 +67,22 @@ extension ProjectSpec {
                     buildSettings += configFile.flattenedBuildSettings()
                 }
             }
-            buildSettings += getProjectBuildSettings(config: config)
+            for (k, v) in getProjectBuildSettings(config: config) {
```
yonaskolb: Once the replacement logic is moved into getProjectBuildSettings, these can be removed

Sources/XcodeGenKit/SettingsBuilder.swift
```diff
@@ -67,14 +67,22 @@ extension ProjectSpec {
...
+                }
+            }
         }
         if let configFilePath = target.configFiles[config.name] {
             if let configFile = try? XCConfig(path: basePath + configFilePath) {
                 buildSettings += configFile.flattenedBuildSettings()
             }
         }
-        buildSettings += getTargetBuildSettings(target: target, config: config)
+        for (k, v) in getTargetBuildSettings(target: target, config: config) {
```
yonaskolb: Once the replacement logic is moved into getTargetBuildSettings, this can be removed

toshi0383:
Thanks for review again!
Fixed regarding your comments! 44b2710

> I don't think we have to remove target build settings because of a project config file

For example, normally I have `MACOSX_DEPLOYMENT_TARGET` in project config file, so I can apply this in each target. (e.g. LGTMKit and LGTM) But currently `SettingPresets/Platforms/macOS.yml` overwrites this.😕
So I thought in general, any presets should never overwrite xcconfig value.🤔
What's your thoughts on this?

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -71,7 +71,7 @@ public class PBXProjGenerator {
         }
 
         let buildConfigs: [XCBuildConfiguration] = spec.configs.map { config in
-            let buildSettings = spec.getProjectBuildSettings(config: config)
+            let buildSettings = spec.getProjectBuildSettings(basePath: basePath, config: config)
```
yonaskolb: I'm thinking of moving the basePath into the spec itself, so we won't need things like this 👍

Sources/XcodeGenKit/SettingsBuilder.swift
```diff
@@ -24,18 +24,47 @@ extension ProjectSpec {
         }
 
         buildSettings += getBuildSettings(settings: settings, config: config)
```
yonaskolb: I think the clearing should happen before this line, so custom build settings don't get removed, just the setting presets

Sources/XcodeGenKit/SettingsBuilder.swift
```diff
@@ -24,18 +24,47 @@ extension ProjectSpec {
...
         buildSettings += SettingsPresetFile.product(target.type).getBuildSettings()
         buildSettings += SettingsPresetFile.productPlatform(target.type, target.platform).getBuildSettings()
         buildSettings += getBuildSettings(settings: target.settings, config: config)
-
+        // Do not overwrite target xcconfig's values.
+        if let configPath = target.configFiles[config.name] {
+            if let configFile = try? XCConfig(path: basePath + configPath) {
+                for (k, _) in configFile.flattenedBuildSettings() {
+                    // FIXME: Catch platform specifier. e.g. LD_RUNPATH_SEARCH_PATHS[sdk=iphone*]
+                    buildSettings.removeValue(forKey: k)
```
yonaskolb: There's quite a bit of code duplication here, as the thing thing is done in 3 places. Can we move to a common function here to clear build settings from a configFile

Sources/XcodeGenKit/SettingsBuilder.swift
```diff
@@ -24,18 +24,47 @@ extension ProjectSpec {
...
     }
 
-    public func getTargetBuildSettings(target: Target, config: Config) -> BuildSettings {
+    public func getTargetBuildSettings(basePath: Path, target: Target, config: Config) -> BuildSettings {
         var buildSettings = BuildSettings()
 
         buildSettings += SettingsPresetFile.platform(target.platform).getBuildSettings()
         buildSettings += SettingsPresetFile.product(target.type).getBuildSettings()
         buildSettings += SettingsPresetFile.productPlatform(target.type, target.platform).getBuildSettings()
         buildSettings += getBuildSettings(settings: target.settings, config: config)
```
yonaskolb: I think the clearing should happen before this line, so custom build settings don't get removed, just the setting presets

yonaskolb:
Yeah ok, we can remove targets settings from a project config file as well

yonaskolb:
By the way does #101 solve your use case, or do you still rely on some of the setting preset values?

Sources/XcodeGenKit/PBXProjGenerator.swift
yonaskolb: I've done this in #103

toshi0383:
You know what.. I just came up with this idea, but rather than filtering presets values by configFiles values, how about just treating presets as `configFile` ?

- If a user doesn't have project level `configFile` specified in spec, then presets values go into `presets.xcconfig` and set as `project xcconfig`.
- If a user does have project level `configFile` specified, then `presets.xcconfig` is ignored. (can be generated)
- Apply those to target level `configFile` and presets too.

yonaskolb:
Yeah I thought about that, but then you can't mix setting presets and xcconfig files. You might have a few target settings in an xcconfig for example, but then as soon as you set that, you lose all the platform and product type presets.
Also I'd prefer not to generate any files in someones project.

toshi0383:
Fixed!

toshi0383:
> By the way does #101 solve your use case, or do you still rely on some of the setting preset values?

For now yes. Because that way spec is shorter.

Sources/XcodeGenKit/SettingsBuilder.swift
```diff
@@ -23,6 +23,13 @@ extension ProjectSpec {
             buildSettings += SettingsPresetFile.config(type).getBuildSettings()
         }
 
+        if let configPath = configFiles[config.name] {
+            // Do not overwrite project xcconfig's value.
+            if let configFile = try? XCConfig(path: basePath + configPath) {
```
yonaskolb: Can we move these config file lines into the common function as well?
```
if let configPath = configFiles[config.name] {
   buildSettings = removeConfigFileSettings(from: buildSettings, configPath: configPath)
}
```

That way it's all kept together, removing duplication. It then also lets us potentially cache the loaded and flattened file settings in one place, because if you have many targets each one of them could be loading and flattening a LOT of files

toshi0383: Should I let you refactor those after merge? I think you have solid image.
I wouldn't mind about my code being modified at all.😉

toshi0383:
Merged master.

yonaskolb:
Refactored in master 👍

yonaskolb:
@keith, would this be something you guys use? Or would you just clear all setting presets?
I wonder if this is worth putting into an option with either a true or false default

keith:
We clear all settings and define every build setting we have in xcconfigs.

PR #78, opened 2017-09-29
Remove CODE_SIGN_IDENTITY: "" from presets

toshi0383:
It turns out empty string is parsed as `NSNull` and it ends-up with buildSettings like this.
```
LGTM.xcodeproj/project.pbxproj:                         CODE_SIGN_IDENTITY = "<null>";
```


toshi0383:
Oh looks like it's been fixed in master already. Thanks!

PR #79, opened 2017-09-29
Always embed frameworks for macOS, but provide link option for ObjC selector calls

toshi0383:

For Objective-C runtime compatibility, `Linked Frameworks and Binaries` is sometimes necessary.

# project.yml
```yml
    dependencies:
      - target: LGTMKit_macOS
      - carthage: APIKit
      - carthage: Alamofire
      - carthage: Himotoki
      - carthage: MASPreferences
        link: true
      - carthage: Nuke
      - carthage: Realm
      - carthage: RealmSwift
      - carthage: Result
      - carthage: RxBlocking
      - carthage: RxCocoa
      - carthage: RxSwift
      - carthage: RxTest
      - carthage: SwiftyUserDefaults
      - carthage: Toast
        link: true
```

# Generated Project
screen shot 2017-09-30 at 6 46 29

e.g. Toggling MASPreferences' tabbar causes `unrecognized selector` runtime error, unless you have `MASPreferences.framework` both embedded and linked.

### DIFF
--- a/docs/ProjectSpec.md
+++ b/docs/ProjectSpec.md
@@ -218,6 +218,12 @@ These only applied to `target` and `framework` dependencies.
 - ⚪️ **codeSign**: `Bool` - Whether the `codeSignOnCopy` setting is applied when embedding framework. Defaults to true
 - ⚪️ **removeHeaders**: `Bool` - Whether the `removeHeadersOnCopy` setting is applied when embedding the framework. Defaults to true
 
+**Carthage link option for macOS**:
+
+For macOS, Carthage frameworks are to be always embedded. But sometime you need to link, too.
+
+- ⚪️ **link**: `Bool` - [macOS only] Whether to link the dependency. Defaults to false.
+
 **Carthage Dependency**
 
 Carthage frameworks are expected to be in `CARTHAGE_BUILD_PATH/PLATFORM/FRAMEWORK.framework` where:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Master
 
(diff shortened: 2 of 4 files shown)


docs/ProjectSpec.md
```diff
@@ -218,6 +218,12 @@ These only applied to `target` and `framework` dependencies.
 - ⚪️ **codeSign**: `Bool` - Whether the `codeSignOnCopy` setting is applied when embedding framework. Defaults to true
 - ⚪️ **removeHeaders**: `Bool` - Whether the `removeHeadersOnCopy` setting is applied when embedding the framework. Defaults to true
 
+**Carthage link option for macOS**:
+
+For macOS, Carthage frameworks are to be always embedded. But sometime you need to link, too.
```
pepibumur: I'm not sure about this. As far as I know, the frameworks generated by Carthage needs to be linked and embedded regardless of the platform.

toshi0383: You are right. That's safer.🤔
Do you want me to fix the embed option to defaults to true for macOS?

pepibumur: I think that would prevent that developers generate the project and it doesn't run because frameworks are missing. I don't see a case where you would link the Carthage frameworks without embedding them.

toshi0383: Well it's working for me, though?
Anyways it's safer to both link and embed, so I'm closing this PR.👌
I will submit another fix for you, because it was only linking Carthage frameworks when I tried.

pepibumur: What do you think @yonaskolb ?

PR #82, opened 2017-10-01
Always embed and link frameworks for macOS

toshi0383:
Replaces #79.
XcodeGen currently don't support embedding Carthage frameworks.

# project.yml
```yml
...
targets:
  LGTM:
    platform: macOS
    type: application
    sources:
      - LGTM
    dependencies:
      - target: LGTMKit_macOS
      - carthage: APIKit
      - carthage: Alamofire
      - carthage: Himotoki
      - carthage: MASPreferences
      - carthage: Nuke
      - carthage: Realm
      - carthage: RealmSwift
      - carthage: Result
      - carthage: RxCocoa
      - carthage: RxSwift
      - carthage: SwiftyUserDefaults
      - carthage: Toast
...
```

# Before
screen shot 2017-10-01 at 9 32 42

# After
screen shot 2017-10-01 at 9 33 20

### DIFF
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -297,6 +297,11 @@ public class PBXProjGenerator {
                 carthageFrameworksByPlatform[target.platform.carthageDirectoryName]?.append(fileReference)
 
                 targetFrameworkBuildFiles.append(buildFile.reference)
+                if target.platform == .macOS {
+                    let embedFile = PBXBuildFile(reference: generateUUID(PBXBuildFile.self, fileReference + target.name), fileRef: fileReference, settings: dependency.buildSettings)
+                    addObject(embedFile)
+                    copyFiles.append(embedFile.reference)
+                }
             }
         }
 
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -297,7 +297,7 @@ public class PBXProjGenerator {
                 carthageFrameworksByPlatform[target.platform.carthageDirectoryName]?.append(fileReference)
 
                 targetFrameworkBuildFiles.append(buildFile.reference)
-                if target.platform == .macOS {
+                if target.platform == .macOS && target.type.isApp {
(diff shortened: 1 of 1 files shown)


yonaskolb:
This would also embed carthage frameworks in framework targets. Is that ok?

toshi0383:
Ohh I'll fix it.

toshi0383:
Fixed in 99e5113 🙏

yonaskolb:
What happens to those frameworks that have carthage dependencies though. They have to be embedded somewhere right? 

Should this be moved down to where the `carthage copy-frameworks` is created, and then embed all the carthage frameworks from all targets?

It would be a good to have create a new macOS target in the TestProject fixture to test that this stuff works

toshi0383:
> Should this be moved down to where the carthage copy-frameworks is created, and then embed all the carthage frameworks from all targets?

No, because user can just add all dependencies flattened in `project.yml`.
I think that's how Carthage works, too. Carthage flattens out all frameworks into `Carthage/Build/$platform`.
(I don't think we should parse `Cartfile.resolved` to know nested dependencies graph either. That's Carthage's job. 👍)
```yml
    dependencies:
      - target: LGTMKit_macOS
      - carthage: APIKit // <- depends on Result
      - carthage: Result
```

yonaskolb:
I don't mean embed **all** carthage frameworks, just looking through the ones that the app target is dependendant one, like the generated script phase does. This means someone can say that a framework has a carthage dependency, without having to also add it to all targets that have that framework as a dependency

yonaskolb:
Can this be moved down to here so it embeds all the relevant carthage frameworks. https://github.com/yonaskolb/XcodeGen/blob/master/Sources/XcodeGenKit/PBXProjGenerator.swift#L392

This is so the target can embed all the carthage dependencies that are linked in all of that target's target dependencies. This is to support the use case where carthage dependencies are defined on the frameworks that need it and not on the application targets.

That whole carthage code will need to be moved up before `copyFiles` so the embeds can be included in that

toshi0383:
Can we do that in separated PR?
I couldn't make it work correctly.

yonaskolb:
What was the issue?

toshi0383:
It correctly embeds nested carthage framework, but doesn't link.
screen shot 2017-10-14 at 6 11 34

Maybe you want to take over ?

Issue #29, opened 2017-08-23
Generate output files for carthage copy frameworks script


yonaskolb:
This will make the  Carthage `copy-frameworks` script run only if the output files are missing. As suggested in the Carthage readme

toshi0383:
>  As suggested in the Carthage readme

I couldn't find this. Could you tell me which section?

yonaskolb:
It's here in point 6 https://github.com/Carthage/Carthage/blob/master/README.md#getting-started

toshi0383:
Thanks 😄

yonaskolb:
Resolved in #84

PR #84, opened 2017-10-04
Add outputpaths to carthage copy-framewors script

mironal:
According to Carthage README, the Output Files should be set to `carthage copy-frameworks` script.

https://github.com/Carthage/Carthage#if-youre-building-for-ios-tvos-or-watchos

### Before

image

### After

image

### DIFF
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -388,7 +388,8 @@ public class PBXProjGenerator {
 
             if target.type.isApp && target.platform != .macOS {
                 let inputPaths = Set(carthageFrameworksToEmbed).map { "$(SRCROOT)/\(carthageBuildPath)/\(target.platform)/\($0)\($0.contains(".") ? "" : ".framework")" }
-                let carthageScript = PBXShellScriptBuildPhase(reference: generateUUID(PBXShellScriptBuildPhase.self, "Carthage" + target.name), files: [], name: "Carthage", inputPaths: inputPaths, outputPaths: [], shellPath: "/bin/sh", shellScript: "/usr/local/bin/carthage copy-frameworks\n")
+                let outputPaths = Set(carthageFrameworksToEmbed).map { "$(BUILT_PRODUCTS_DIR)/$(FRAMEWORKS_FOLDER_PATH)/\($0)\($0.contains(".") ? "" : ".framework")" }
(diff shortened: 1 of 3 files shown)


Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -388,7 +388,8 @@ public class PBXProjGenerator {
 
             if target.type.isApp && target.platform != .macOS {
                 let inputPaths = Set(carthageFrameworksToEmbed).map { "$(SRCROOT)/\(carthageBuildPath)/\(target.platform)/\($0)\($0.contains(".") ? "" : ".framework")" }
-                let carthageScript = PBXShellScriptBuildPhase(reference: generateUUID(PBXShellScriptBuildPhase.self, "Carthage" + target.name), files: [], name: "Carthage", inputPaths: inputPaths, outputPaths: [], shellPath: "/bin/sh", shellScript: "/usr/local/bin/carthage copy-frameworks\n")
+                let outputPaths = Set(carthageFrameworksToEmbed).map { "$(BUILT_PRODUCTS_DIR)/$(FRAMEWORKS_FOLDER_PATH)/\($0)\($0.contains(".") ? "" : ".framework")" }
```
pepibumur: Good catch @mironal. This helps xcode now when it needs to do recompile certain components because any of the frameworks changed. Looks good to me!

yonaskolb:
Thanks @mironal!
This resolves issue #29

Sources/XcodeGenKit/PBXProjGenerator.swift
yonaskolb: @mironal I believe the input and output paths have to line up in terms of their order. The code here SHOULD be fine, but the lists may be out of order. That's because Set doesn't guarantee an order, though I think it would be the same with the same data and same run.
To be safe can you remove the Set here and move it up to the carthageFrameworksToEmbed?

```
let carthageFrameworksToEmbed = Array(Set(carthageDependencies
            .filter { ($0.embed ?? true) }
            .map { $0.reference }))
            .sorted()
```
The Set is used by the way to remove duplicates, as an application target as well it's dependent targets may reference the same carthage dependency

mironal: Thanks for your code review. I will fix it. 😄

mironal:
I fixed 😃

Issue #92, opened 2017-10-17
Frameworks and static library targets should be linked (and sometimes copied) into their dependent targets


keith:
Currently if you create a configuration like this:

```yaml
name: Project
targets:
  Project:
    type: application
    platform: iOS
    sources:
      - Sources
    dependencies:
      - target: Module
  Module:
    type: framework
    platform: iOS
    sources:
      - ModuleSources
```

And generate a project, the `Module.framework` is copied into the `Project.app` target, but isn't linked in the "Link Binary and Libraries" phase (as is happening if you just use a vendored framework.

This is the same if you change `type: framework` to `type: library.static` with the one difference that static libraries should be linked, but shouldn't be copied, into the final `.app`

keith:
Ah it actually looks like this commit is the culprit https://github.com/yonaskolb/XcodeGen/commit/183630159e760c81c725ba10c75a4679e1a5d46a

keith:
I've submitted a possible fix for this: https://github.com/yonaskolb/XcodeGen/pull/93

yonaskolb:
Are you getting build errors without this linking? All the examples I’ve tried work without this

keith:
Yep, in this case with a static library you get this linker error:

```
Undefined symbols for architecture x86_64:
  "__T06Module3FooV3baryyFZ", referenced from:
      __T07Project11AppDelegateC11applicationSbSo13UIApplicationC_s10DictionaryVySC0E16LaunchOptionsKeyVypGSg022didFinishLaunchingWithH0tF in AppDelegate.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

Here's a sample project: [XcodeGenTest6.zip](https://github.com/yonaskolb/XcodeGen/files/1389463/XcodeGenTest6.zip)

yonaskolb:
Ah ok, got it. Do you think there is any advantage to linking frameworks as well, even though Xcode seems to do ok with just pointing at the target?

keith:
We're doing that in our project now, I'm not sure there's any functional advantage, but personally I do like having them there explicitly, since it seems a little strange even that Xcode is smart enough to link them, but not static libraries, without explicitly including them in the link phase.

pepibumur:
[...]

PR #93, opened 2017-10-17
Fix dependency framework/library linking

keith:
This PR updates XcodeGen to link framework and library dependencies of targets, into the dependent targets. This is required for them to be usable at runtime. This also removes libraries from the copy framework build phase, since that just results in a `.a` file ending up in your final package.

This should fix https://github.com/yonaskolb/XcodeGen/issues/92

### DIFF
--- a/Sources/ProjectSpec/ProjectExtensions.swift
+++ b/Sources/ProjectSpec/ProjectExtensions.swift
@@ -46,6 +46,14 @@ extension PBXProductType {
         }
     }
 
+    public var isFramework: Bool {
+        return self == .framework
+    }
+
+    public var isLibrary: Bool {
+        return self == .staticLibrary || self == .dynamicLibrary
+    }
+
     public var isExtension: Bool {
         return fileExtension == "appex"
     }
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -256,7 +256,8 @@ public class PBXProjGenerator {
                     targetFrameworkBuildFiles.append(buildFile.reference)
                 }
 
-                if embed {
+                if embed && !dependencyTarget.type.isLibrary {
+
                     let embedSettings = dependency.buildSettings
                     let embedFile = PBXBuildFile(reference: generateUUID(PBXBuildFile.self, dependencyFileReference + target.name), fileRef: dependencyFileReference, settings: embedSettings)
                     addObject(embedFile)
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -25,7 +25,7 @@ public class PBXProjGenerator {
     var variantGroupsByPath: [Path: PBXVariantGroup] = [:]
 
     var targetNativeReferences: [String: String] = [:]
-    var targetBuildFileReferences: [String: String] = [:]
+    var targetBuildFiles: [String: PBXBuildFile] = [:]
     var targetFileReferences: [String: String] = [:]
     var topLevelGroups: [PBXGroup] = []
     var carthageFrameworksByPlatform: [String: [String]] = [:]
@@ -93,7 +93,7 @@ public class PBXProjGenerator {
 
             let buildFile = PBXBuildFile(reference: generateUUID(PBXBuildFile.self, fileReference.reference), fileRef: fileReference.reference)
             addObject(buildFile)
-            targetBuildFileReferences[target.name] = buildFile.reference
+            targetBuildFiles[target.name] = buildFile
         }
 
         let targets = try spec.targets.map(generateTarget)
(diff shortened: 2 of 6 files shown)


Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -248,9 +248,10 @@ public class PBXProjGenerator {
                 addObject(targetDependency)
                 dependencies.append(targetDependency.reference)
 
-                // don't bother linking a target dependency
-                // let dependencyBuildFile = targetBuildFileReferences[dependencyTargetName]!
-                // targetFrameworkBuildFiles.append(dependencyBuildFile)
+                if dependencyTarget.type.isLibrary || dependencyTarget.type.isFramework {
+                    let dependencyBuildFile = targetBuildFileReferences[dependencyTargetName]!
+                    targetFrameworkBuildFiles.append(dependencyBuildFile)
```
keith: Hmm, this seems to not be the type that Xcode is expecting.

keith: Oops, I was testing with a project without a bundle identifier, which really confuses Xcode

keith:
Updated the project since the changes are legit, the nexet test run should be green.

keith:
Hmm, it looks like there is some issue with this, where right after generating the project, if I do a `pod install`, `xcodeproj` throws this error:

```
RuntimeError - [Xcodeproj] Consistency issue: no parent for object `Module2.framework`: `FrameworksBuildPhase`, `FrameworksBuildPhase`
```

I'll try and look in to why this is happening

keith:
Based on the project diff after opening Xcode, it seems like the issue might be re-using file reference IDs in the "frameworks" build phase. As soon as you open the project in Xcode, it gets re-written to have new IDs, but leaves the original there for 1 of the references.

keith:
Ok I just pushed a commit with a fix for this. So it turns out that the problem is, if you want to include a framework like this into multiple copy/frameworks/whatever phases, you need separate file PBXBuildFiles for each case. But they all need to point back to the original PBXFileReference which represents the original target's product.

This now allows the same framework to be linked to multiple dependent targets

Sources/ProjectSpec/ProjectExtensions.swift
```diff
@@ -46,6 +46,14 @@ extension PBXProductType {
         }
     }
 
+    public var isFramework: Bool {
```
pepibumur: I'd add some tests here.

keith: Done!

keith:
Thanks for the review! Added some tests here

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -248,11 +248,18 @@ public class PBXProjGenerator {
...
-                // targetFrameworkBuildFiles.append(dependencyBuildFile)
+                if dependencyTarget.type.isLibrary || dependencyTarget.type.isFramework {
+                    let dependencyBuildFile = targetBuildFiles[dependencyTargetName]!
+                    let buildFile = PBXBuildFile(reference: generateUUID(PBXBuildFile.self, dependencyBuildFile.reference + target.name), fileRef: dependencyBuildFile.fileRef)
+                    addObject(buildFile)
+                    targetFrameworkBuildFiles.append(buildFile.reference)
+                }
 
                 if embed {
+                    if dependencyTarget.type.isLibrary {
```
yonaskolb: Is a library _never_ embeded, or are there special cases where they are? 
If it can be sometimes, ca we move this logic up into where `let embed` is defined, which gives the embedding a default, and then users can override this.

If it's definitely never, we can also move this if into the if above `if embed && !dependencyTarget.type.isLibrary`

Tests/LinuxMain.swift
```diff
@@ -5,4 +5,5 @@ XCTMain([
     testCase(GeneratorTests.allTests),
     testCase(SpecLoadingTests.allTests),
     testCase(FixtureTests.allTests),
+    testCase(TargetTests.allTests),
```
yonaskolb: Can we make this ProjectSpecTests?

Sources/ProjectSpec/ProjectExtensions.swift
```diff
@@ -46,6 +46,14 @@ extension PBXProductType {
...
+        return fileExtension == "framework"
```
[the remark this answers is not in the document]
yonaskolb: Let's leave the getter here to match the other accessors, but check the case within it 👍

Sources/XcodeGenKit/PBXProjGenerator.swift
keith: They are never embedded. Since that would just mean that you're literally copying the `.a` file into your `.app`, which would never be used at runtime since static libraries are linked into your final binary instead. I've updated the condition!

Sources/ProjectSpec/ProjectExtensions.swift
```diff
@@ -46,6 +46,14 @@ extension PBXProductType {
...
+    }
+
+    public var isLibrary: Bool {
+        return name == "library.static" || name == "library.dynamic"
```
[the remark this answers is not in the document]
keith: done

Sources/ProjectSpec/ProjectExtensions.swift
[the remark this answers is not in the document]
keith: done

Tests/LinuxMain.swift
keith: done

keith:
All feedback addressed!

PR #94, opened 2017-10-17
Exclude more file types from resources

keith:
Entitlements files, gpx files, explicit lproj directories, and apns test files should
never be included in a target's copy resources phase.

Ideally this would also include files with the plist extension, but
there are some outliers such as AppIntentVocabulary.plist.

### DIFF
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -450,7 +450,7 @@ public class PBXProjGenerator {
             switch fileExtension {
             case "swift", "m", "cpp": return .sources
             case "h", "hh", "hpp", "ipp", "tpp", "hxx", "def": return .headers
-            case "xcconfig": return nil
+            case "xcconfig", "entitlements", "gpx", "lproj", "apns": return nil
             default: return .resources
             }
         }


yonaskolb:
Thanks @keith. The `lproj` directories should already be taken care of when the source directories are parsed, but no harm in adding it here.
Would `gpx` always be ignored? Wouldn't someone want to sometimes embed some `gpx` files?

keith:
`gpx` should be ignored because, unless you're actually trying to vend `gpx` files to use at runtime, they are only used by Xcode at build time, they don't serve any purpose by being in your final app's resources directory, besides wasting a bit of space.

yonaskolb:
Yeah, but there might be legitimate cases of bundling gpx files within some sort of mapping application. I agree though that in the common case they should be ignored. I'll merge this now and then we'll remove that filetype once we add the ability to exclude file types.

I've added you as a maintainer of the project by the way! 👍 
There is also an open slack group where we discuss the development of `XcodeGen` as well as `xcproj` and other tools. You can join here if you're interested http://xcodeswift.herokuapp.com/

keith:
Awesome thanks! Yea I think this is less important to have once `exclude` is supported, but for now it's nice to not have the unexpected files in your final product. Awesome I'll join!

daltonclaybrook:
We use GPX files as resources in one of our test bundles to test some routing functionality, so we need these files to be part of the copy resources phase for that target. Is there a workaround I can use to support my use-case?

Issue #91, opened 2017-10-17
Ability to copy bundle product types into other targets


keith:
With this configuration:

```yaml
name: Project
targets:
  Project:
    type: framework
    platform: iOS
    dependencies:
      - target: ProjectResources
  ProjectResources:
    type: bundle
    platform: iOS
```

You can create a framework target, and an associated resource bundle for that target. But it doesn't seem like there's any way for you to add `ProjectResources` in this case to the copy resources phase of `Project`.

This would be very nice to support for the case where you have static libraries and resource bundles, since that's the only way to vend resources and not have conflicts with the targets linking the final library.

keith:
I've submitted a possible fix to this here: https://github.com/yonaskolb/XcodeGen/pull/95

jerrymarino:
@yonaskolb @keith I think I've got a similar issue. I'm trying to add some pre built bundles from various Vendors to my iOS App. It looks like merely including a bundle in a targets `sources` won't work. I'm not sure if pre built bundles should be treated as `sources`, but it is aligned with how other resources are implemented

![prebuiltbundles](https://user-images.githubusercontent.com/1245820/32355291-0d8f0de4-bfeb-11e7-9875-e30a8bad47f5.png)

yonaskolb:
@jerrymarino, your issue should be resolved in #148

PR #95, opened 2017-10-17
Allow copying of resource files from targets

keith:
Previously we were assuming that all targets that were not app
extensions should be added to the copy frameworks build phase, even
though we didn't have any guarantee they were actually frameworks. This
updates that to ensure that things copied in the copy frameworks phase
are actually frameworks, and then falls back to the resources phase
instead. This fixes the ability to embed bundle targets, and copy them
as resources.

Fixes https://github.com/yonaskolb/XcodeGen/issues/91

### DIFF
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -229,7 +229,8 @@ public class PBXProjGenerator {
 
         var dependencies: [String] = []
         var targetFrameworkBuildFiles: [String] = []
-        var copyFiles: [String] = []
+        var copyFrameworksReferences: [String] = []
+        var copyResourcesReferences: [String] = []
         var copyWatchReferences: [String] = []
         var extensions: [String] = []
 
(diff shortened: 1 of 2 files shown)


yonaskolb:
Nice catch @keith!

keith:
Thanks! Rebased

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -261,10 +262,12 @@ public class PBXProjGenerator {
                     if dependencyTarget.type.isExtension {
                         // embed app extension
                         extensions.append(embedFile.reference)
+                    } else if dependencyTarget.type.isFramework {
```
pepibumur: We'd need a bit more check here since the extension is not the only one determining if the framework needs to be copied. Only the dynamic ones need to. You can use [lipo](https://pewpewthespells.com/blog/convert_static_to_dynamic.html) for that.

keith: Yea, I thought about that case, we actually have that case and I'm currently setting `embed: False` on it since we don't currently have a reference to the actual binary.

keith: We can also use `file` which might be easier than `lipo` for this

keith: I thought this was at least an improvement since that's now the only case we're not handling correctly after this and my other PRs

pepibumur: I agree this could be the second iteration to make it more concrete 👍. I didn't know it was possible to get that info using `file`.

PR #96, opened 2017-10-17
Add support for copying Watch apps

keith:
Watch apps aren't copied to the same place as other targets, they need
their own copy files phase with a custom path just for watch targets.


yonaskolb:
Thanks @keith! It would be nice to have a watch target in the test fixture, but I'll merge this for now

keith:
Agreed. Thanks!!

PR #97, opened 2017-10-18
Strip XcodeGen after building

keith:
This removes the symbols from the final binary, making it much smaller.
In my tests it took the binary from 13mbs to 8.6mbs.


keith:
Hmm, closing this, this seems to cause issues

Issue #25, opened 2017-08-22
Add directories as folder references


ImLaufderZeit:
Adding specific directories as **Folder References** would be really nice feature.
I'm referring to the thing below:
<img width="364" alt="folderreference" src="https://user-images.githubusercontent.com/31232499/29550376-691f3232-8748-11e7-8bc9-b08505da84ed.png">

pepibumur:
That's a good one @ImLaufderZeit 😛

yonaskolb:
Yeah, that's a good feature. Given that the groups are synced to disk anyway though, I thought it would be less important, because as far as I know the only downside to using groups as opposed to folder references is that you have to manually manage them, but that isn't an issue with XcodeGen. Would specific problem would this solve for you?

I could imaging it maybe looking like this:
```yaml
targets:
    - name: MyTarget
    - sources:
        - path: MyTarget
          name: MyTarget
          type: group
       - path: ../MyDirectory
          type: folderReference
```
This would add an optional name property (that defaults to the directory name), and an optional type (which would default to `group`). Sources could still be written as an array of strings that use the defaults, or you could specific this more verbose form with more flexibility. And a single source string would still be allow as it is now for simple cases (`sources: MyTarget`)

Using the more verbose form would mean you could potentially define an `excludes: */*.uglyFiles` for group types as well

What do you think?

ImLaufderZeit:
@yonaskolb, that sounds just awesome.

> ... Given that the groups are synced to disk anyway though, ..... Would specific problem would this solve for you?

Well, in fact, if #24 has been solved - there is no problem, exactly as you say.

yonaskolb:
Resolves in #151

PR #105, opened 2017-10-24
Remove sourceExcludes

bkase:
Great project!

~~`sourceExludes` should be `sourceExcludes` instead. I fixed this everywhere I could find it.~~

Removed `sourceExcludes` since it's dead for now (as discussed below)

### DIFF
--- a/Sources/ProjectSpec/Target.swift
+++ b/Sources/ProjectSpec/Target.swift
@@ -16,7 +16,6 @@ public struct Target {
     public var platform: Platform
     public var settings: Settings
     public var sources: [String]
-    public var sourceExludes: [String]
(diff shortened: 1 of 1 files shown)


yonaskolb:
Thanks @bkase, good catch!`sourceExcludes` doesn't actually do anything at the moment, and is left over from when this project started. It still needs to be implemented in #5, but I think will be taking a slightly different approach than this variable, which is hinted at in #25.

Renaming this here doesn't hurt though :)

yonaskolb:
In fact, if you want to remove these references for now, even better 😄 👍

bkase:
Done!

yonaskolb:
Thanks @bkase!

PR #106, opened 2017-10-25
Support file sources

bkase:
The `sources` key of the project spec only supported directories and not
files. Now it supports both!

This commit introduces a `getGroups` overload that doesn't explicitly
invoke `path.children()`, but instead accepts `children` as a parameter.
This allows us to invoke the `children` overload of getGroups with just
the files we want to include (determined by specifying the sources).

Now for sourcePaths that are files, we first group them by common
parent directories and then feed those specific files in the children
parameter of `getGroups`.

Motivation: We have directories where we need to compile a subset of
files in the same folder -- and we already have a nice way to get the exact
list of files we need for an arbitrary target. We need this feature to avoid
extra manual effort grouping files by directories and moving files around.

### DIFF
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -66,7 +66,8 @@ public class PBXProjGenerator {
 
         for group in spec.fileGroups {
             //TODO: call a seperate function that only creates groups not source files
-            _ = try getSourceFiles(path: spec.basePath + group)
+            let path = spec.basePath + group
+            _ = try getSourceFiles(path: path, children: try path.children())
         }
 
         let buildConfigs: [XCBuildConfiguration] = spec.configs.map { config in
@@ -158,7 +159,7 @@ public class PBXProjGenerator {
 
         let carthageDependencies = getAllCarthageDependencies(target: target)
 
-        let sourceFiles = try target.sources.flatMap(getSourceFiles)
+        let sourceFiles = try getAllSourceFiles(sources: target.sources)
 
         // find all Info.plist files
         let infoPlists: [Path] = target.sources.map { spec.basePath + $0.path }.flatMap { (path) -> [Path] in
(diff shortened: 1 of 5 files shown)


Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -160,8 +160,20 @@ public class PBXProjGenerator {
...
         var sourceFiles: [SourceFile] = []
 
-        for source in sourcePaths {
-            let sourceGroups = try getGroups(path: source)
+        // load groups for sourcePaths
+        let (files, dirs) = (sourcePaths.filter{ $0.isFile }, sourcePaths.filter{ $0.isDirectory })
+        let filesByParent: [Path: [Path]] = files.reduce([:]) { acc, file in
+            var mut = acc
+            let group = file.parent()
+            mut[group] = (mut[group] ?? []) + [file]
```
yonaskolb: We can use the cool new swift 4 syntax 😎 
```
mut[group, default: []].append(file)
```

yonaskolb:
Fantastic! Could you add some file sources at different directory hierarchies in the Fixtures/TestProject. Both in the spec.yml and some random source file.

Just for some info the TestProject fixture is a bit of a catch all for random tests at the moment and just acts as a snapshot to make sure things don't change, and can also be checked if it still builds.

You're welcome to create some more functional tests in `ProjectGeneratorTests` if you have time

keith:
It sounds like this might be a better option than using fileGroups for the same purpose? That’s what I’m currently doing.

yonaskolb:
Yeah, I have been using fileGroups for this purpose

yonaskolb:
@bkase could you add at least one source file to the TestProject fixture?

bkase:
Hey sorry for the slow response on these PRs, I got very busy with work, and now I'm at a conference. Is it okay if I get back to these on Wednesday?

yonaskolb:
Yeah no problem, no rush

bkase:
@yonaskolb I managed to do this much more cleanly by sticking it inside the `getSourceFiles` function (thanks for the todo there). I added a single file to the Fixtures, not sure if I need to manually generate a project? Anyway, feel free the takeover the PR to get it landed if there are any more minor changes needed! Thanks!

bkase:
Wait, sorry forgot about the problem I had with groups before. It still has to be handled one level above -- I'll fix it in a moment

bkase:
Okay! Should be good, also the fixture looks good to me! @yonaskolb let me know if there is anything else (or feel free to finish up yourself)

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -518,7 +535,7 @@ public class PBXProjGenerator {
         var groups: [PBXGroup] = []
 
         for path in directories {
-            let subGroups = try getSourceFiles(path: path, depth: depth + 1)
+            let subGroups = try getSourceFiles(path: path, children: children, depth: depth + 1)
```
yonaskolb: Should this be `try path.children()`?

bkase: oops, good catch

bkase:
@yonaskolb okay should be fixed now

PR #108, opened 2017-10-25
Optionally make intermediate filler groups

bkase:
This commit adds a new option `mirrorsFileSystem` that defaults to
false. When it is false, the behavior of XcodeGen is the same as
before. When it is true, we make intermediate groups recursively until
we reach the basePath. In practice that means if you've chosen
`Platform/PINFoundation/Sources` as one of your sourcePaths, you get a
top-level group of `Platform` and under that `PINFoundation` and under
that `Sources`. This is instead of the default behavior of just making
`Sources` a top-level group (which is confusing when your directory is
called `Sources` for example).

I think this would probably be a good place to put a fixture, but I wasn't
sure how I should go about adding one. Please advise, thanks!

`mirrorsFileSystem: false` (or unset):

screen shot 2017-10-25 at 3 27 06 pm

`mirrorsFileSystem: true`:

screen shot 2017-10-25 at 3 36 16 pm


yonaskolb:
Awesome @bkase! I've been thinking of something similar but adding it to a more configurable source struct as hinted at here https://github.com/yonaskolb/XcodeGen/issues/25 where you could define the name or path of group.
These would be a good catch all solution though.

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -578,17 +612,18 @@ public class PBXProjGenerator {
             }
         }
 
-        let groupPath: String = depth == 0 ? path.byRemovingBase(path: spec.basePath).string : path.lastComponent
         let group: PBXGroup
-        if let cachedGroup = groupsByPath[path] {
-            group = cachedGroup
+        if spec.options.mirrorsFileSystem {
```
yonaskolb: Once thing that would need to be checked for is the case where you define a source outside of the base path eg. `sources: ../directory`. Currently this would probably add every directory until the filesystem root

Sources/ProjectSpec/ProjectSpec.swift
```diff
@@ -29,6 +29,7 @@ public struct ProjectSpec {
 
     public struct Options {
         public var carthageBuildPath: String?
+        public var mirrorsFileSystem: Bool
```
yonaskolb: As XcodeGen already somewhat mirrors the file system, can we call this `createIntermediateGroups` or something similar?

yonaskolb:
Could you add documentation to `docs/ProjectSpec.md` as well.

In terms of the fixture, you could add something to the TestProject. You could make a hierarchy of directories and then include one as a source at least one level down to validate this works

Sources/XcodeGenKit/PBXProjGenerator.swift
bkase: Good catch

bkase:
@yonaskolb okay made requested changes + added example to fixture

yonaskolb:
To fix the tests, re-run `swift test` to re-generate the fixture and commit it. And make sure the generated project isn't opened or xcode will make some slight modifications to it, which is what seems to have happened here. (Resolving #114, would fix the need for that)

bkase:
Okay rebased + `swift test` commits @yonaskolb

bkase:
Tests will fail for now, but will pass when #108 lands (after a rebase)

PR #109, opened 2017-10-25
Add option for not linking dependencies

keith:
This adds a new attribute to Dependency that allows consumers to choose
to not link a dependency. This is useful for if you have this dependency
tree with static libraries:

App -> A -> Shared
App -> B -> Shared

Where A and B both share a static library dependency, that is finally
linked into App. If Shared is added to the link phase of A and B, you
end up with duplicate symbols during the link phase. With this change
consumers could set link: False on A and B's dependency on Shared, this
way Shared will get build before A and B, but not linked.

### DIFF
--- a/docs/ProjectSpec.md
+++ b/docs/ProjectSpec.md
@@ -219,6 +219,7 @@ A dependency can be one of a 3 types:
 These only applied to `target` and `framework` dependencies.
 
 - ⚪️ **embed**: `Bool` - Whether to embed the dependency. Defaults to true for application target and false for non application targets.
+- ⚪️ **link**: `Bool` - Whether to link the dependency. Defaults to true but only static library and dynamic frameworks are linked. This only applies for target dependencies.
 - ⚪️ **codeSign**: `Bool` - Whether the `codeSignOnCopy` setting is applied when embedding framework. Defaults to true
 - ⚪️ **removeHeaders**: `Bool` - Whether the `removeHeadersOnCopy` setting is applied when embedding the framework. Defaults to true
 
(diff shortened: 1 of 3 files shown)


yonaskolb:
Nice one @keith! Would this option be applicable to the carthage or framework dependency types as well? Then this check should be added to those cases.

Also could you add some documentation in Docs/ProjectSpec.md

keith:
> Would this option be applicable to the carthage or framework dependency types as well?

Good question, after thinking about this a bit, I think no, because the reason to have a target dependency for something is to make sure it's built first, where in the case of those already built frameworks if you don't want to link them, they just shouldn't be a dependency of your target. Also since we're specifically talking about the dynamic framework case, whether or not you link them at this phase won't affect the final product, where with static libraries (or static frameworks) this will be an issue because you'll end up with duplicate symbols.

keith:
Added documentation!

docs/ProjectSpec.md
```diff
@@ -219,6 +219,7 @@ A dependency can be one of a 3 types:
 These only applied to `target` and `framework` dependencies.
 
 - ⚪️ **embed**: `Bool` - Whether to embed the dependency. Defaults to true for application target and false for non application targets.
+- ⚪️ **link**: `Bool` - Whether to link the dependency. Defaults to true for but only static library and dynamic frameworks are linked.
```
yonaskolb: `for but`
And maybe mention that this is only applicable for `target` dependency types

keith: Updated!

PR #112, opened 2017-10-27
add support --version option

mironal:
I think that it is better for xcodegen to have `--version` option as it is for many commands.

However, I think that this will increase the effort of editing main.swift for version up process. 😟 
If you feel troublesome, you can reject this pull request. 😇

### DIFF
--- a/Sources/XcodeGen/main.swift
+++ b/Sources/XcodeGen/main.swift
@@ -15,6 +15,8 @@ import ProjectSpec
 import JSONUtilities
 import Rainbow
 
+let version = "1.3.0"
+
 func generate(spec: String, project: String) {
 
     let specPath = Path(spec).normalize()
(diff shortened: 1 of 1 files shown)


keith:
This would be fantastic! @yonaskolb thoughts on the extra work for shipping?

yonaskolb:
Yes, adding this is definitely a good idea. Thanks @mironal!
In terms of keeping this updated I think the version should live in one place and then a release script would update it places like this, tag a version, and also use it to update the brew formula. 
The release script can live in the make file and the version could either be in there or in it's own file. Is there a standard way of doing that?

yonaskolb:
Would this `--version` flag or a `xcodegen version` command be preferred?

keith:
I think we should just bake that in to a build script somewhere, to sed out a variable in a swift file with a version number. Personally I prefer `--version` but don't feel strongly

PR #118, opened 2017-10-31
Don't add unnecessary headers build phase

ryohey:
I found that header files are unwantedly copied into the application package.
I checked that only the dynamicLibrary and the framework contained headers from the newly created project from the template.

|type|needs headers|
|:-:|:-:|
|application|NO|
|framework|YES|
|dynamicLibrary|YES|
|staticLibrary|copies by Copy Files build phase|
|bundle|NO|
|unitTestBundle|NO|
|uiTestBundle|NO|
|appExtension|NO|
|commandLineTool|NO|
|watchApp|NO|
|watch2App|NO|
|watchExtension|NO|
|watch2Extension|NO|
|tvExtension|NO|
|messagesApplication|NO|
|messagesExtension|NO|
|stickerPack|NO|
|xpcService|NO|

### DIFF
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -354,9 +354,11 @@ public class PBXProjGenerator {
         addObject(resourcesBuildPhase)
         buildPhases.append(resourcesBuildPhase.reference)
 
-        let headersBuildPhase = PBXHeadersBuildPhase(reference: generateUUID(PBXHeadersBuildPhase.self, target.name), files: getBuildFilesForPhase(.headers))
-        addObject(headersBuildPhase)
-        buildPhases.append(headersBuildPhase.reference)
+        if target.type == .framework || target.type == .dynamicLibrary {
+            let headersBuildPhase = PBXHeadersBuildPhase(reference: generateUUID(PBXHeadersBuildPhase.self, target.name), files: getBuildFilesForPhase(.headers))
+            addObject(headersBuildPhase)
+            buildPhases.append(headersBuildPhase.reference)
+        }
 
         if !targetFrameworkBuildFiles.isEmpty {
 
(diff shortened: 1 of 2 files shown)


yonaskolb:
Thanks @ryohey!
@keith @pepibumur, does this check out to you?

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -354,9 +354,11 @@ public class PBXProjGenerator {
         addObject(resourcesBuildPhase)
         buildPhases.append(resourcesBuildPhase.reference)
 
-        let headersBuildPhase = PBXHeadersBuildPhase(reference: generateUUID(PBXHeadersBuildPhase.self, target.name), files: getBuildFilesForPhase(.headers))
-        addObject(headersBuildPhase)
-        buildPhases.append(headersBuildPhase.reference)
+        if target.type == .framework || target.type == .dynamicLibrary {
+            let headersBuildPhase = PBXHeadersBuildPhase(reference: generateUUID(PBXHeadersBuildPhase.self, target.name), files: getBuildFilesForPhase(.headers))
```
keith: Should we also only create this if `getBuildFilesForPhase` here returns a  non-empty array?

keith:
One thought, but looks good to me!

Sources/XcodeGenKit/PBXProjGenerator.swift
yonaskolb: Yes, this could be added though the same would apply to the sources and resources build phase. I left them in there in case people wanted to easily files add to them in Xcode (that's assuming they are only generating once and then checking in the project file)
In either case it's not something we have to cover in this PR

keith: Ah makes sense, I guess for some of these it's just preference, since in our all swift project we'll _never_ need this build phase.

yonaskolb: Yeah, I think the empty ones could be removed

yonaskolb: This has been done in #149

PR #120, opened 2017-11-01
Add .mm file as a source type

enmiller:
While working on a project, I discovered that .mm files were being added to the project but not to the build sources phase.


keith:
+1

yonaskolb:
Good catch, thanks @enmiller!

PR #121, opened 2017-11-02
Support CompilerFlags in Sources

bkase:
Added support for compilerFlags in source list. If any source file
metadata (like compilerFlags) is attached to a directory the metadata
propagates downwards to all children recursively until the files are
reached.

Since there is an extra bit of metadata we need to carry around during
production of PBXObjects, we can no longer pass a `Path` directly.
`Source` needs `path` to be `String` because we don't know the
basePath until the spec is fully loaded. When we do have the basePath
we can compute the real `Path` and start to use PathKit APIs.

I attempted a first pass where I just used the `Source` everywhere and
forwarded along a basePath at all operations that needed it, but that
was too messy, so I ended up making another struct that enriched the
source with the basePath (called `EnrichedSource`). I don't love the
name, let me know if you come up with a better one.

Fixtures and unit tests are updated as well.

### DIFF
--- a/Fixtures/TestProject/Project.xcodeproj/project.pbxproj
+++ b/Fixtures/TestProject/Project.xcodeproj/project.pbxproj
@@ -6,17 +6,17 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
-		BF1073850101 /* AppDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = FR1332263601 /* AppDelegate.swift */; };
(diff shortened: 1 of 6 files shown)


yonaskolb:
Instead of creating `EnrichedSource` or passing through the `basePath` everywhere, could we just read the `spec.basePath` wherever it's needed?

bkase:
Tests will fail for now, but will pass when #108 lands (after a rebase)

PR #126, opened 2017-11-03
Fix localized files with same name

### LINKED ISSUE #122: Localised files with same name aren't generated properly
If localized files have with the same name but different extensions, they are generated incorrectly

screen shot 2017-11-02 at 1 39 35 pm

ryohey:
- Refactor the code generating variant groups at e24c0e2 ce27af0. These do not change any output.
- Fix localized files with same name at 66a2893. This will add localized files into the variant group that matches the file path with extension first. We use the variant group matched the file name without extension secondary for the case such as storyboard and strings.

close #122

### DIFF
--- a/Fixtures/TestProject/App_iOS/Base.lproj/Localizable.strings
+++ b/Fixtures/TestProject/App_iOS/Base.lproj/Localizable.strings
@@ -1,0 +1,7 @@
+/* 
+  Localizable.strings
+  Project
+
+  Created by ryohey on 2017/11/03.
+  
+*/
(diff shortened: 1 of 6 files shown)


yonaskolb:
Nice one! Thanks @ryohey!

yonaskolb:
By the way @ryohey I'm not sure if you're in the slack group, but you're welcome to join. We discuss XcodeGen and the projects from https://github.com/xcodeswift
http://xcodeswift.herokuapp.com

Anyone else who is reading this is free to join too, it's an open community 👍

PR #128, opened 2017-11-03
Support precompiled bundle files

jerrymarino:
Bundles are a special file type in Xcode: they aren't treated as directories.
This patch adds the ability to support these files.

### DIFF
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -29,6 +29,7 @@ public class PBXProjGenerator {
     var topLevelGroups: [PBXGroup] = []
     var carthageFrameworksByPlatform: [String: Set<String>] = [:]
     var frameworkFiles: [String] = []
+    var bundleFiles: [Path: SourceFile] = [:]
 
     var uuids: Set<String> = []
     var project: PBXProj!
@@ -524,7 +525,7 @@ public class PBXProjGenerator {
 
     // Add groups for all parents recursively
     // ex: path/foo/bar/baz/Hello.swift -> path:[foo:[bar:[baz:[Hello.swift]]]]
-    func getIntermediateGroups(path: Path, group: PBXGroup) -> PBXGroup {
+    func getGroups(path: Path, group: PBXGroup) -> PBXGroup {
         // verify path is a subpath of spec.basePath
         guard Path(components: zip(path.components, spec.basePath.components).map{ $0.0 }) == spec.basePath else {
             return group
@@ -536,7 +537,7 @@ public class PBXProjGenerator {
         }
 
         // recursive case
-        return getIntermediateGroups(
+        return getGroups(
             path: path.parent(),
             group: getSingleGroup(path: path, mergingChildren: [group.reference])
         )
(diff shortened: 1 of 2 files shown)


Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -240,6 +241,32 @@ public class PBXProjGenerator {
         var copyWatchReferences: [String] = []
         var extensions: [String] = []
 
+        // Find bundles within a targets sources
+        let bundlePaths = target.sources
```
yonaskolb: It seems bundles would then only be supported if they are added explicitly as a single source, not if they lived inside a sources directory?

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -564,6 +597,11 @@ public class PBXProjGenerator {
             (try path.children().sorted(), path)
 
         let excludedFiles: [String] = [".DS_Store"]
+        // Don't process the files of a bundle. Bundles are a special file type
+        // in Xcode ( not directories ).
+        if path.extension == "bundle" {
```
yonaskolb: Couldn't all the handling of bundles happen right here, instead of all around the place?

Sources/XcodeGenKit/PBXProjGenerator.swift
jerrymarino: You're right! Bundles should be added to targets, if they are contained within a sources directory.

Sources/XcodeGenKit/PBXProjGenerator.swift
jerrymarino: I think this would be pretty nice

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -557,6 +558,30 @@ public class PBXProjGenerator {
...
+                                             explicitFileType: "wrapper.plug-in",
+                                             path: path.lastComponent, includeInIndex: 1)
+        fileReferencesByPath[path] = fileReference.reference
+        addObject(fileReference)
+        // Add the bundle to the parent's group.
+        let group = getIntermediateGroups(
+            path: path.parent().parent(),
+            group: getSingleGroup(path: path.parent(), mergingChildren: [fileReference.reference], depth: 0)
+        )
+        addObject(group)
```
bkase: `getIntermediateGroups` will recursively do `addObject(group)` for you (so you don't have to do it here)

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -565,6 +590,12 @@ public class PBXProjGenerator {
 
         let excludedFiles: [String] = [".DS_Store"]
 
+        // Don't process the files of a bundle. Bundles are a special file type
+        // in Xcode ( not directories ).
+        if path.extension == "bundle" {
+            return ([getBundle(atPath: path)], [])
```
bkase: Will anything break if we don't return the groups created by `getBundle` here, @yonaskolb ?

Sources/XcodeGenKit/PBXProjGenerator.swift
yonaskolb: - yes
- this group isn't used as a child anywhere
- why is a group being created?
- intermedia groups should only be made if the spec.options.createIntermediateGroups == true

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -557,6 +558,30 @@ public class PBXProjGenerator {
         return variantGroup
     }
 
+    func getBundle(atPath path: Path) -> SourceFile {
+        // It is possible for several targets to depend on the same bundle.
+        // Do not attempt to create a bundle for each target.
+        if let cachedBundle = bundleFiles[path] {
```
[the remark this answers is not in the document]
jerrymarino: @yonaskolb I don't quite follow this. There should be a single build file per bundle, which is added to multiple targets phases.

Sources/XcodeGenKit/PBXProjGenerator.swift
jerrymarino: This code uses `getIntermediateGroup` as a helper to get all the groups - it is unrelated the user facing feature. We need make sure all groups are created - otherwise it isn't created if there is no other files in the directory than the bundle. Does that make sense?

@bkase you are right, we shouldn't add parent group here.


Re

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -557,6 +558,37 @@ public class PBXProjGenerator {
...
+        fileReferencesByPath[path] = fileReference.reference
+        addObject(fileReference)
+        addObject(getGroup(path: path.parent(),mergingChildren: [fileReference.reference], depth: 0))
+        let buildFile =  PBXBuildFile(reference: generateUUID(PBXBuildFile.self, fileReference.reference), fileRef: fileReference.reference)
+        let source = SourceFile(path: path, fileReference: fileReference.reference, buildFile: buildFile)
+        bundleFiles[path] = source
+        return source
+    }
+
+    /// Return the top level group for a given path
```
jerrymarino: @yonaskolb I updated this to consider the `createIntermediateGroups` - what do you think?

Sources/XcodeGenKit/PBXProjGenerator.swift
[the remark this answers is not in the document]
yonaskolb: Xcode creates a unique build file for every target

PR #136, opened 2017-11-05
Create topLevelGroup if not-cache hit. Fixes #129

### LINKED ISSUE #129: createIntermediateGroups results in broken file refs in projects
With this project config:

```yaml
name: Project
options:
targets:
  Project:
    type: application
    platform: iOS
    sources:
      - Sources/Foo/bar.swift
```

The resulting project has broken file refs to `bar.swift`:

with

Example project: [XcodeGenTest.zip](https://github.com/yonaskolb/XcodeGen/files/1442410/XcodeGenTest.zip)

bkase:
We want to add a group to the topLevelGroups array if it was not caused
by a cache hit (at any level recursively -- because this means we
already would have added it), AND only the top-most group (after
all intermediate groups finish.

This regressed in the most recent fix to fileGroups.

Looks like this regressed with the fix to fileGroups. It seems like
the `topLevelGroups` field is a big source of problems, so I put a TODO
to figure out a way to kill it.

It would be nice to capture createIntermediateGroups somehow in the fixture, but then we wouldn't be catching the other case. Maybe we need another fixture?

### DIFF
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -26,6 +26,7 @@ public class PBXProjGenerator {
     var targetNativeReferences: [String: String] = [:]
     var targetBuildFiles: [String: PBXBuildFile] = [:]
     var targetFileReferences: [String: String] = [:]
+    // TODO: This topLevelGroups-field is a source of problems, can we refactor it away somehow
     var topLevelGroups: [PBXGroup] = []
     var carthageFrameworksByPlatform: [String: Set<String>] = [:]
     var frameworkFiles: [String] = []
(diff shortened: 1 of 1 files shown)


yonaskolb:
Thanks for this @bkase, but I'm not sure I'm such a fan of all the required caching params and the complexity this adds.
I think I would need to see a failing test to understand the problem better, but can this be solved by checking if a group is already in topLevelGroups before trying to add it?

bkase:
We would need to check if any parents of this path is in topLevelGroups as well, but that's an interesting idea, I'll see if I can get something better

bkase:
We're heads down at work on getting a proof-of-concept up, so I may not have time for this until the weekend. `fileGroups` should work on master, though, so hopefully this isn't blocking anyone.

yonaskolb:
Tagging @keith, so he's aware

yonaskolb:
Fixed in #144

PR #139, opened 2017-11-08
Support paths directly to xcassets

bkase:
Pretend xcassets are files to allow top-level xcassets in the spec.

I added a new xcasset to the fixture that generates incorrectly before this PR and correctly after it.


yonaskolb:
Nice. Can you add a test case here https://github.com/yonaskolb/XcodeGen/blob/master/Tests/XcodeGenKitTests/ProjectGeneratorTests.swift#L315

bkase:
Done

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -563,7 +563,8 @@ public class PBXProjGenerator {
     
     func getSources(sourceMetadata source: Source, path: Path, depth: Int = 0) throws -> (sourceFiles: [SourceFile], groups: [PBXGroup]) {
         // if we have a file, move it to children and use the parent as the path
-        let (children, path) = path.isFile ?
+        // pretend xcassets are files
+        let (children, path) = path.isFile || path.extension == "xcassets" ?
```
yonaskolb: Would it make sense to handle all directories with extensions like this?

bkase: Good question -- I think some directories with extensions could actually be directories. Maybe we want a whitelist? It looks like the `xcproj` library handles a lot of these special ones, but `xcassets` are not handled by `xcproj` (which is why they have to be handled up here).

What do you think? What would we want in the whitelist?

yonaskolb: Yeah a whitelist would be good. I guess `.bundle` would go in there. Seems a cleaner solution to #128 as well

bkase: I'll test that out

yonaskolb: Merging this in it's current state for now

PR #142, opened 2017-11-09
Fix duplicate sources in groups, fixes #141

### LINKED ISSUE #141: Group children are duplicated when sources are shared across targets
PR #136 seems to have broken the generation of groups and files when multiple targets share the same source directories, even if `createIntermediateGroups` is false.
This issue can be seen the Test project fixture here https://github.com/yonaskolb/XcodeGen/commit/e48045da9e6600df50f6291d02ea4cd78a7bdaa8

bkase:


Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -504,7 +504,7 @@ public class PBXProjGenerator {
     func getSingleGroup(path: Path, mergingChildren children: [String], depth: Int = 0) -> PBXGroup {
         let group: PBXGroup
         if let cachedGroup = groupsByPath[path] {
-            cachedGroup.children += children
+            cachedGroup.children = Array(Set(cachedGroup.children + children))
```
yonaskolb: This means we lose ordering of files

yonaskolb: And if we just sort this we lose alphabetical ordering of files

bkase:
It's less efficient now, but I think it should be okay. There shouldn't ever be too many files in a single group in practice

Sources/XcodeGenKit/PBXProjGenerator.swift
yonaskolb: By the way, I changed things up so that all files and groups are now sorted at the end, so we don't have to worry about this sort of thing anymore

PR #148, opened 2017-11-12
Add support for linking directly to bundle

### LINKED ISSUE #91: Ability to copy bundle product types into other targets

yonaskolb:
Or any directory with an extension
Fixes #91

### DIFF
--- a/Sources/XcodeGenKit/SourceGenerator.swift
+++ b/Sources/XcodeGenKit/SourceGenerator.swift
@@ -183,9 +183,8 @@ class SourceGenerator {
 
     private func getSources(targetSource: TargetSource, path: Path, isRootSource: Bool) throws -> (sourceFiles: [SourceFile], groups: [PBXGroup]) {
 
-        let directoryFiles = ["xcassets"] // these are directories we should treat as files
-        let isDirectoryFile = path.extension != nil && directoryFiles.contains(path.extension!)
-        let isFile = path.isFile || isDirectoryFile
+        // treat all directories with extensions as files
+        let isFile = path.isFile || path.extension != nil
         let fileName = isFile && isRootSource ? targetSource.name : nil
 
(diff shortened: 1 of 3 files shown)


Sources/XcodeGenKit/SourceGenerator.swift
```diff
@@ -183,9 +183,8 @@ class SourceGenerator {
 
     private func getSources(targetSource: TargetSource, path: Path, isRootSource: Bool) throws -> (sourceFiles: [SourceFile], groups: [PBXGroup]) {
 
-        let directoryFiles = ["xcassets"] // these are directories we should treat as files
-        let isDirectoryFile = path.extension != nil && directoryFiles.contains(path.extension!)
-        let isFile = path.isFile || isDirectoryFile
+        // treat all directories with extensions as files
+        let isFile = path.isFile || path.extension != nil
```
jerrymarino: Nice fix! What happens if you have a directory, which is named `Some.other`, which should be treated as a directory? What do you think about whitelisting known directory extensions that Xcode treat as files by conventions? Just a thought!

yonaskolb: Yeah, I was going to do that, but left it out here. It is also needed when checking what is a file or directory down below.
I can add it now.
Can you think of any valid directory extensions other than `xcassets` and `bundle`?

PR #149, opened 2017-11-12
Don't generate empty build phases

yonaskolb:


keith:
🎉

yonaskolb:
It seems this is causing issues. A framework with no Compile Sources phase is not linking properly

rpassis:
We are currently seeing an issue with sticker extensions where an empty compile sources phase is getting generated, and that in turn triggers a "Multiple commands produce" error message when building.

Apologies if this is documented elsewhere but Is there a way to force exclude a specific/empty build phase for a target?

```
  Stickers:
    type: app-extension.messages-sticker-pack
    platform: iOS
    configFiles:
      Debug: ...
      Release: ...
    settings:
      INFOPLIST_FILE: ...
    sources:       
      - path: Stickers.xcassets
```

keith:
It sounds like maybe for this specific target type we should never include this phase since no resources in it should ever have to be compiled?

yonaskolb:
Hmm indeed. @rpassis could you open a new issue and list all the target types this is an issue for

acecilia:
I am experiencing the same issue with a `application.watchapp2` target, where the `application.watchapp2` target has no sources, but a `Compile Sources (0)` build phase is created that results in a `Multiple commands produce` build time error.

```
  Watch:
    platform: watchOS
    type: application.watchapp2
    sources:
      - path: Watch
        includes:
          - "Assets.xcassets"
          - "Icon.xcassets"
          - "Interface.storyboard"
        buildPhase: resources
    dependencies:
      - target: WatchExtension
```

I did not find any workaround yet

PR #151, opened 2017-11-15
Add folder reference support

### LINKED ISSUE #25: Add directories as folder references

yonaskolb:
Resolves #25 
This adds a `type` property to `TargetSource` which can be either
- file
- group
- folder

If the type is left out it will be guessed. (If the path points to a file or the directory has an extension it will be treated as a file reference, otherwise it will be a group)

If it's set to `folder` it will be a folder reference.

This also refactors some more source generation code


Sources/XcodeGenKit/SourceGenerator.swift
```diff
@@ -259,26 +261,73 @@ class SourceGenerator {
...
+            createIntermediaGroups(for: group.reference, at: path)
         }
 
         groups.insert(group, at: 0)
         return (allSourceFiles, groups)
     }
 
+    private func getSourceFiles(targetSource: TargetSource, path: Path) throws -> [SourceFile] {
+
+        let type = targetSource.type ?? (path.isFile || path.extension != nil ? .file : .group)
```
bkase: This is awesome! This will make our generation way faster since we waited on I/O for the `path.isFile` check for every file in our project.

PR #160, opened 2017-11-18
[WIP] Initial idea for testing SourceGenerator

ryohey:
This adds a tree structure of path strings so that we can test more cases easily.

## Problem

There are many variations of project configurations, but only a few cases have been tested. I want to write tests that check if variant groups created correctly. We currently use real files as the fixture but that makes hard to test many cases.

## Proposal 

We can change SourceGenerator to use virtual file tree (`enum File`) which is the snapshot of file paths. That looks like the following.

```swift
public indirect enum File {
    case directory(FilePath, [File]) // path, children
    case file(FilePath) // path
}
```

By doing this, we can write a test as follows for various edge cases.

```swift
// file tree
let dir: File = File("Src", [
    File("main.swift"),
    File("ViewController.swift"),
    File("Base.lproj", [File("Main.storyboard")]),
    File("en.lproj", [File("Main.strings")]),
    ])
// process the file tree
sourceGenerator.getFileGroups(in: dir)
try expect(sourceGenerator.knownRegions) == ["Base", "en"]
// and tests about generating groups, variant groups, SourceFile
```

I implemented it partially as an example in this PR and added the test about knownRegions. 

What do you think of this way?

I think that this kind of change may lead to the reduction of side effect and the introduction of immutability by making it possible to think of the part that generates the project structure from the file tree as a kind of map function.

I feel that name `enum File` as is not good.  Aside from that, scanning the file tree may cause performance problems if it contains very deep directories.

### DIFF
--- a/Tests/XcodeGenKitTests/SourceGeneratorTests.swift
+++ b/Tests/XcodeGenKitTests/SourceGeneratorTests.swift
@@ -1,0 +1,41 @@
+//
+//  SourceGeneratorTests.swift
+//  XcodeGenKit
+//
+//  Created by ryohey on 2017/11/18.
+//
+
+import Spectre
+import XcodeGenKit
+import xcproj
+import PathKit
+import ProjectSpec
+import Yams
+
+
+func sourceGeneratorTests() {
+
+    describe("SourceGenerator") {
+        let application = Target(name: "MyApp", type: .application, platform: .iOS,
+                                 settings: Settings(buildSettings: ["SETTING_1": "VALUE"]),
+                                 dependencies: [])
+        let options = ProjectSpec.Options(bundleIdPrefix: "com.test")
+        let spec = ProjectSpec(basePath: "", name: "test", targets: [application], options: options)
+        var objects: [PBXObject] = []
+        let referenceGenerator = ReferenceGenerator()
+        let proj = PBXProj(objectVersion: 46, rootObject: referenceGenerator.generate(PBXProject.self, "test"))
+        let sourceGenerator = SourceGenerator(spec: spec, proj: proj, referenceGenerator: referenceGenerator, addObject: { objects.append($0) })
+
+        $0.it("_") {
+            let dir: File = File("Src", [
+                File("main.swift"),
+                File("ViewController.swift"),
+                File("Base.lproj", [File("Main.storyboard")]),
+                File("en.lproj", [File("Main.strings")]),
+                ])
+            sourceGenerator.getFileGroups(in: dir)
+            try expect(sourceGenerator.knownRegions.contains("Base")).to.beTrue()
+            try expect(sourceGenerator.knownRegions.contains("en")).to.beTrue()
+        }
+    }
+}
--- a/Sources/XcodeGenKit/File.swift
+++ b/Sources/XcodeGenKit/File.swift
@@ -1,0 +1,51 @@
+//
+//  File.swift
+//  XcodeGenKit
+//
+//  Created by ryohey on 2017/11/18.
+//
+
+import Foundation
+import PathKit
+
+internal struct FilePath {
+    var string: String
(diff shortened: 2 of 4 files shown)


rahul-malik:
@ryohey - Nice! This is actually very similar to an enum I used to model file directory structures at Pinterest. I gave a talk about the benefits of this approach at Mobiconf / Functional Swift. Here is the notes if you're interested:

https://github.com/rahul-malik/writing-dsls/blob/master/Deck/deckset.md#reify-with-recursive-enums

I came up with a similar enum `DirectoryItem` that was sufficient for our needs.
```swift
   enum DirectoryItem {
      case File(name: String, content: String)
      indirect case Folder(name: String, contents: [DirectoryItem])
    }
```

and then its usage was:
```swift
let package = DirectoryItem.Folder(
	name: "MyLibrary",
	contents: [
		.File(name: ".gitignore", content: ""),
		.Folder(
			name: "Sources",
			contents: [.File(name: "MyLibrary.swift", content: "...")]),
		.Folder(
			name: "Tests",
			contents: [
				.File(name: "LinuxMain.swift", content: "..."),
				.Folder(name: "MyLibraryTests",
						contents: [.File(name: "MyLibraryTests.swift", content: "...")])
			]
		)
	]
)
```

ryohey:
@rahul-malik wow thank you very much. I needed such a document. The name DirectoryItem is better. The only difference is the name and full path, and each element must have full path to use in place of PathKit's Path.

ryohey:
@yonaskolb Thank you for reviewing! I missed that! So, there are two things I think.
- I think it's better to write unit tests instead of integrated tests. How do you think about it? (yes this is not related to this PR now)
- SourceGenerator mutates inner state while walking on the file tree. I think its some part are seem to able to be refactored as function that maps the file path tree to some outputs. When that, the only nondeterministic part is Path.children. Although it's just that I like pure functional way.

yonaskolb:
Yeah, I agree unit tests are usually better than integration tests, but that's usually only due to speed of execution, and the granularity of the tests. In this case there's quite a bit of important interaction with the filesystem via PathKit that is important to capture, like the `excludes` and it's use of glob, the way relative paths interact, or the way files and dictionaries are determined. Being able to generate the "fixtures" also gives us the granularity we require.
Having some strict unit tests in addition would be great to have though!

In terms of making the SourceGenerator purely functional, that would be fantastic. If you can find a way not maintaining caches of visited paths, and keeping a history of top level groups, please open a PR as it's a tricky problem.

rahul-malik:
+ @bkase as I'm sure he has some good ideas around making `SourceGenerator` more functional :)

Tests/XcodeGenKitTests/SourceGeneratorTests.swift
```diff
@@ -0,0 +1,41 @@
...
+    describe("SourceGenerator") {
+        let application = Target(name: "MyApp", type: .application, platform: .iOS,
+                                 settings: Settings(buildSettings: ["SETTING_1": "VALUE"]),
+                                 dependencies: [])
+        let options = ProjectSpec.Options(bundleIdPrefix: "com.test")
+        let spec = ProjectSpec(basePath: "", name: "test", targets: [application], options: options)
+        var objects: [PBXObject] = []
+        let referenceGenerator = ReferenceGenerator()
+        let proj = PBXProj(objectVersion: 46, rootObject: referenceGenerator.generate(PBXProject.self, "test"))
+        let sourceGenerator = SourceGenerator(spec: spec, proj: proj, referenceGenerator: referenceGenerator, addObject: { objects.append($0) })
```
yonaskolb: I've now removed the dependency on PBXProj in SourceGenerator 👍

ryohey:
@yonaskolb Thank you for your response! I understand it. I'll try it someday, closing this PR.

PR #161, opened 2017-11-19
Fix TargetSource excludes tests.

peymankh:


yonaskolb:
Great, thanks @peymankh!

yonaskolb:
Hi @peymankh. I just realized this still doesn't support the most common use case of filtering out all filetypes in the source. This would be achieved by `**/*.myIgnoredFile` with a proper pattern matcher, or `*.myIgnoredFile` if we were doing things recursively, but either one of those doesn't work.

I just came across this while reading some documentation for buddybuild and it's a good example of the patterns I'd like to support https://docs.buddybuild.com/builds/selective_builds.html#examples

I had a bit of a play though, and I don't think `Path.glob` supports the ** pattern (any sub directory)

peymankh:
Hey @yonaskolb. You are right, `**` is not a "true" glob pattern and because of this `Path.glob` does not support it. Having said that, `**` is usually a convention for build tools to perform recursive operations. 

I haven't got enough time yet to look at how we can implement this in an optimised way since `PathKit` does not support this. When we add the functionality, it should work exactly like BuddyBuild examples.

If you want I can open an issue for this and address it as soon as we can.

yonaskolb:
Yeah, let's discuss further in an issue. If you could layout your findings in one that would be great 😄 👍

PR #164, opened 2017-11-20
Add ProjectSpec product

soffes:
This exposes ProjectSpec.framework to packages that depend on XcodeGen. I'm using this to do some introspection of the project structure in a command line tool. It would be great if this was public. Thanks for your consideration!


yonaskolb:
Thanks @soffes! Yes, that's a good move 👏

Issue #163, opened 2017-11-20
Link & Embed implicit dependencies


sbarow:
I have multiple `.xcodeproj` generated using `XcodeGen` which are then added to a work space.

Say we have the structure:
- App (depends on UI & Foundation)
- UI (depends on Foundation)
- Foundation (no dependencies)

With there respective `project.yml` files being defined like this:
```
## App/project.yml
...
dependencies:
	- framework: UI.framework 
	- framework: Foundation.framework
...

## UI/project.yml
...
dependencies:
	- framework: Foundation.framework
...
```

This will compile and launch correctly on simulators because the build option for implicit dependencies is checked, however this will crash on launch when running on device because the implicit frameworks are not being copied to the apps frameworks folder.

Example of embedded framework in project:

![screen shot 2017-11-20 at 10 18 59 pm](https://user-images.githubusercontent.com/867796/33042490-59b39fe0-ce42-11e7-99d0-351fc21397bd.png)

- `UI.framework` is embedded correctly - by manually selecting the framework which is being built by the UI project and placed in the `$(BUILT_PRODUCTS_DIR` (note the `... in Build/Debug-iphoneos`)
- `Foundation.framework` is the added dependency  which will cause a crash because its not copied into the app frameworks folder.

Is there a way to specify that the framework being linked is in the `$(BUILT_PRODUCTS_DIR`?

yonaskolb:
Hi @sbarow 
There's currently no way to achieve this in the spec that I know of. I guess this goes back to the question of generating the whole workspace as per #24. I want to open up a proper discussion about that soon.

Does anyone have an idea of a workaround in the meantime?

rahul-malik:
@sbarow - Can you just add `$(BUILT_PRODUCTS_DIR)` to `FRAMEWORK_SEARCH_PATHS` via an xcconfig file / config setting?

sbarow:
I have a script that can do the workspace generation pretty easily depending on the folder structure & `project.yml` files and its dependencies but needs this to be work before I can make it public.

Could we create a new `Dependency` type which would create a `PBXFileReference` with `sourceTree: .buildProductsDir`? Not sure how involved this would be and if its a road I want to go down?

sbarow:
@rahul-malik that doesn't seem to work, the build still fails saying it can't find the frameworks.

yonaskolb:
@sbarow adding something like that to `Dependency` is definitely doable, if you'd like to submit a PR for it.

PR #165, opened 2017-11-20
Add implicit type to Dependency

sbarow:
As a first pass, this seems like all that is needed to add an implicit framework dependency. Are there any other edge cases we need to worry about?

I'll add some tests when I get a 👍 

Fixes: #163

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -7,6 +7,7 @@ let package = Package(
     products: [
         .executable(name: "XcodeGen", targets: ["XcodeGen"]),
         .library(name: "XcodeGenKit", targets: ["XcodeGenKit"]),
+        .library(name: "ProjectSpec", targets: ["ProjectSpec"]),
     ],
     dependencies: [
         .package(url: "https://github.com/kylef/PathKit.git", from: "0.8.0"),


yonaskolb:
Nice! Have you tested it on your own project?

sbarow:
Yeah, tested in a workspace with 5 `.xcodeproj` (workspace and projects are all generated). Works in simulator and on device.

sbarow:
That could work. Just need to be explicit in the docs that it only applies to `framework` type.

yonaskolb:
Yeah, there are already some options that are only applied to certain types, or have different defaults. That reminds me, can you update ProjectSpec.md?

Maybe an `implicit` boolean? Thoughts anyone? As future work may impact this decision.

jerrymarino:
This is interesting, and perhaps adding transitive dependency logic could be useful.

Let's say I've got a project:
```
 name: SimpleProject
  targets:
   App:
     type: application
     dependencies:
      - target: Transitive
   Transitive:
     type: framework
     dependencies: 
      - target: Simple
   Simple:
     type: framework
     platform: iOS
     dependencies: 
       - implicit: Implicit.framework
```

We'd want to propagate all of the dependencies to the Application, so that they are all linked and or embedded. The `App` linker invocation would need `-framework Transitive, -framework Simple, -framework Implicit`, and the frameworks would need to be copied over.


The current way I'm achieving this is adding all transitive deps directly to top level targets, like `Application`, `Extension`, etc..
Here is what the required spec would be:
```
   App:
     type: application
     dependencies:
      - target: Transitive
      - target: Simple
      - implicit: Implicit.framework
```

Is this something that `XcodeGen` should handle internally, instead of doing it in the spec?


Based on #163 it seems like this may be related?

yonaskolb:
@jerrymarino good point. At the moment only carthage dependencies are automatically embedded in app targets even if they aren't defined in that target, but are in a dependent target. (this isn't the case for macOS targets anymore by the way as it hasn't been added back yet after the changes in #76 removed that)

This would be possible for other dependency types as well, but I just thought there would be too many edge cases and special build settings and build scripts that people might have to do this reliably for anything else.
Perhaps another one for `ProjectSpec.options`?

sbarow:
@jerrymarino the goal is to not have to do anything manually 😜

Regarding transitive dependencies, its an interesting one and something I have thought about. At Uber we use [buck](https://buckbuild.com) and do transitive dependency resolution manually (with the help of auditing that lets us know if we are missing a dependency in the BUCK file) I believe its a more sane approach to let the user know they are missing a dependency for a target, the other thing to consider is cyclic dependencies.

I'll take a look at both in another diff?

PR #166, opened 2017-11-21
Add implicit option to framework Dependency

sbarow:
Updated version of [#165](https://github.com/yonaskolb/XcodeGen/pull/165) - I didn't have coffee before trying to rebase this morning 😞

Fixes: #163

### DIFF
--- a/Docs/ProjectSpec.md
+++ b/Docs/ProjectSpec.md
@@ -249,9 +249,9 @@ These only applied to `target` and `framework` dependencies.
 
 **Implicit Framework options**:
 
-This only applies to `framework` dependencies.
+This only applies to `framework` dependencies. Implicit framework dependencies are useful in Xcode Workspaces which have multiple `.xcodeproj` that are not embedded within each other yet have a dependency on a framework built in an adjacent `.xcodeproj`. 
 
-- ⚪️ **implicit**: `Bool` - Whether the framework is an implicit dependency. This can be useful in a Workspace where projects are not nested within each other.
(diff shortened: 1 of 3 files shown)


Sources/ProjectSpec/Dependency.swift
```diff
@@ -17,12 +17,14 @@ public struct Dependency: Equatable {
     public var codeSign: Bool = true
     public var removeHeaders: Bool = true
     public var link: Bool = true
+    public var implicit: Bool?
```
yonaskolb: As a nil value has no real meaning here we can make it default to false. Something like `embed` is an optional as the default changes depending on other factors

Docs/ProjectSpec.md
```diff
@@ -247,6 +247,12 @@ These only applied to `target` and `framework` dependencies.
 - ⚪️ **codeSign**: `Bool` - Whether the `codeSignOnCopy` setting is applied when embedding framework. Defaults to true
 - ⚪️ **removeHeaders**: `Bool` - Whether the `removeHeadersOnCopy` setting is applied when embedding the framework. Defaults to true
 
+**Implicit Framework options**:
+
+This only applies to `framework` dependencies.
+
+- ⚪️ **implicit**: `Bool` - Whether the framework is an implicit dependency. This can be useful in a Workspace where projects are not nested within each other.
```
yonaskolb: We can mention that the default is false, and also how this works - that the framework will look in the build products directory

sbarow:
Great, updated. Let me know if the docs section makes sense.

PR #167, opened 2017-11-22
Add quiet option

soffes:
This option will suppress success messages. A new `Logger` struct is added to the `XcodeGen` target that encapsulates printing, exiting, and the Rainbow logic.

I only added a Commander flag for `--quiet`. If we added a Commander interface, it would be trivial to remove colored output too if that is desired. I didn't do this since I couldn't decide on a name. Also, I figured if you were in a situation where you didn't want colored output, you might also want to suppress emojis.

### DIFF
--- a/Sources/XcodeGen/Logger.swift
+++ b/Sources/XcodeGen/Logger.swift
@@ -1,23 +1,15 @@
 import Foundation
 import Rainbow
 
-/// Mechanism for logging informational, success, and fatal messages.
 struct Logger {
 
     // MARK: - Properties
 
-    /// Should informational and success messages be suppressed
     let isQuiet: Bool
-
-    /// Should the output be colored
     let isColored: Bool
 
     // MARK: - Initializers
 
-    /// Initialize a logger
-    ///
-    /// - parameter isQuiet: should informational and success messages be suppressed
-    /// - parameter isColored: should the output be colored
     init(isQuiet: Bool = false, isColored: Bool = true) {
         self.isQuiet = isQuiet
         self.isColored = isColored
@@ -25,16 +17,10 @@ struct Logger {
 
     // MARK: - Logging
 
-    /// Log a fatal message and exit with status code 1
-    ///
-    /// - parameter message: the message to log
     func fatal(_ message: String) {
         print(isColored ? message.red : message)
     }
 
-    /// Log an informational message
-    ///
-    /// - parameter message: the message to log
     func info(_ message: String) {
         if isQuiet {
             return
@@ -43,9 +29,6 @@ struct Logger {
         print(message)
     }
 
-    /// Log a success message and exit with status code 0
-    ///
-    /// - parameter message: the message to log
     func success(_ message: String) {
         if isQuiet {
             return
--- a/Sources/XcodeGen/Logger.swift
+++ b/Sources/XcodeGen/Logger.swift
@@ -1,0 +1,56 @@
+import Foundation
+import Rainbow
+
+/// Mechanism for logging informational, success, and fatal messages.
+struct Logger {
+
+    // MARK: - Properties
+
+    /// Should informational and success messages be suppressed
+    let isQuiet: Bool
+
+    /// Should the output be colored
+    let isColored: Bool
+
+    // MARK: - Initializers
+
+    /// Initialize a logger
+    ///
+    /// - parameter isQuiet: should informational and success messages be suppressed
+    /// - parameter isColored: should the output be colored
+    init(isQuiet: Bool = false, isColored: Bool = true) {
+        self.isQuiet = isQuiet
+        self.isColored = isColored
+    }
+
+    // MARK: - Logging
+
+    /// Log a fatal message and exit with status code 1
+    ///
+    /// - parameter message: the message to log
+    func fatal(_ message: String) {
+        print(isColored ? message.red : message)
+    }
+
+    /// Log an informational message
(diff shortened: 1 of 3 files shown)


Sources/XcodeGen/Logger.swift
```diff
@@ -0,0 +1,61 @@
...
+
+import Foundation
+import Rainbow
+
+/// Mechanism for logging informational, success, and fatal messages.
+struct Logger {
+
+    // MARK: - Properties
+
+    /// Should success messages be suppressed
```
yonaskolb: Seems it suppresses `info` not `success` messages

Sources/XcodeGen/main.swift
```diff
@@ -13,51 +13,47 @@ import XcodeGenKit
...
+        logger.fatal("Generation failed: \(error.localizedDescription)")
     }
 }
 
 command(
     Option<String>("spec", "project.yml", flag: "s", description: "The path to the spec file"),
     Option<String>("project", "", flag: "p", description: "The path to the folder where the project should be generated"),
-    generate)
-    .run(version)
+    Flag("quiet", flag: "q", description: "Suppress printing of success messages", default: false),
```
yonaskolb: Seems it suppresses info not success messages

Sources/XcodeGen/Logger.swift
```diff
@@ -0,0 +1,61 @@
...
+
+        print(message)
+    }
+
+    /// Log a success message and exit with status code 0
+    ///
+    /// - parameter message: the message to log
+    func success(_ message: String) -> Never {
+        print(isColored ? message.green : message)
+        exit(0)
```
yonaskolb: We might print multiple success messages. No need to exit here

yonaskolb:
Could you add documentation to Readme as well?

Sources/XcodeGen/Logger.swift
```diff
@@ -0,0 +1,61 @@
+//
+//  Logger.swift
+//  XcodeGen
+//
+//  Created by Sam Soffes on 11/21/17.
+//
```
yonaskolb: Could you remove header. I've removed them all in master, and they are all removed automatically now in `make format_code`

soffes:
Made the changes your requested! I tested this locally since there aren't tests for the output of the command line tool (yet!).

I was thinking about `--verbose` too. Maybe this option down the road could list all of the targets, groups, files etc. This may be useful for debugging at some point.

Sources/XcodeGen/Logger.swift
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    // MARK: - Logging
+
+    /// Log a fatal message and exit with status code 1
+    ///
+    /// - parameter message: the message to log
+    func fatal(_ message: String) -> Never {
+        print(isColored ? message.red : message)
+        exit(1)
```
sbarow: Seems a bit weird to put this kind of logic in a `Logger`? You wouldn't really expect the program to `exit` after logging some `fatal` error, seems like that should happen in the catch block?

soffes: Fair. I was just excited to use `-> Never`. I'll remove. I agree it's weird for the logger to `exit`.

Sources/XcodeGen/main.swift
```diff
@@ -5,51 +5,52 @@ import XcodeGenKit
...
     let spec: ProjectSpec
     do {
         spec = try ProjectSpec(path: specPath)
-        print("📋  Loaded spec:\n  \(spec.debugDescription.replacingOccurrences(of: "\n", with: "\n  "))")
+        logger.info("📋  Loaded spec:\n  \(spec.debugDescription.replacingOccurrences(of: "\n", with: "\n  "))")
     } catch let error as JSONUtilities.DecodingError {
-        print("Parsing spec failed: \(error.description)".red)
-        exit(1)
+        logger.fatal("Parsing spec failed: \(error.description)")
+		exit(1)
```
yonaskolb: spacing :trollface:

soffes: Every time I regenerate the Xcode project it changes it back to tabs 🙄

Sources/XcodeGen/Logger.swift
```diff
@@ -0,0 +1,56 @@
...
+    /// - parameter isQuiet: should informational and success messages be suppressed
+    /// - parameter isColored: should the output be colored
+    init(isQuiet: Bool = false, isColored: Bool = true) {
+        self.isQuiet = isQuiet
+        self.isColored = isColored
+    }
+
+    // MARK: - Logging
+
+    /// Log a fatal message and exit with status code 1
```
yonaskolb: This is why I stay away from doc comments as they always get out of date. 😄  In this case, it mentions status code here (and in success). 
If this doesn't exit anymore, we could also rename this to `error`.

Also if we wanted to remove all the duplicate `exit(1)` in the main file, a simple helper function there called `fatalError(_ message: String)` could log an error and then exit

yonaskolb: Sorry for not seeing this before :)

sbarow: `fatalError(_ message: String)` good use case for `-> Never` @soffes 😜

soffes:
@yonaskolb thanks for your time on this! Sorry for all of the back and forth!

yonaskolb:
Thank YOU :)

PR #171, opened 2017-11-23
[WIP] Define the project in a Project.swift file

pepibumur:
## Tasks
- [ ] Dump `Project.swift` into a `TOML` format.
- [ ] Parse `TOML` to get a `Project` description that can be passed to XcodeGen.
- [ ] Figure out how to bundle the dynamic library that contains the project description.
- [ ] Add command to generate a project where you can edit the `Project.swift` file.

---
This change is [Reviewable](https://reviewable.io/reviews/yonaskolb/xcodegen/171)

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -8,6 +8,7 @@ let package = Package(
         .executable(name: "XcodeGen", targets: ["XcodeGen"]),
         .library(name: "XcodeGenKit", targets: ["XcodeGenKit"]),
         .library(name: "ProjectSpec", targets: ["ProjectSpec"]),
+        .library(name: "ProjectDescription", type: .dynamic, targets: ["ProjectDescription"])
     ],
     dependencies: [
         .package(url: "https://github.com/kylef/PathKit.git", from: "0.8.0"),
@@ -35,6 +36,7 @@ let package = Package(
           "xcproj",
           "Yams",
         ]),
+        .target(name: "ProjectDescription", dependencies: []),
         .testTarget(name: "XcodeGenKitTests", dependencies: [
           "XcodeGenKit",
         ])
--- a/Sources/ProjectDescription/Project.swift
+++ b/Sources/ProjectDescription/Project.swift
@@ -1,0 +1,20 @@
+import Foundation
+
+public final class Project {
+    
+    // MARK: - Attributes
+    
+    /// Project name.
+    private let name: String
+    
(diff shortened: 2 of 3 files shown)


pepibumur:
@yonaskolb I started working on this and I have a question. Reading from [this article](http://bhargavg.com/swift/2016/06/11/how-swiftpm-parses-manifest-file.html) Swift Package Manager includes the definition of the project in a library that does not contain other dependencies. I guess they decided to do it this way to avoid having to depend on multiple libraries.

In case of `XcodeGen`, `ProjectSpec`, depends on `JSONUtilities`, `Yams`, and `xcproj` that we'd need to bundle with `XcodeGen` to get the `Project.swift` parsing working. I wonder which one of these two options do you think is the best:

- Find a way to remove these dependencies and keep that library as simple as possible, just including the models that represent the project.
- Creating another library that contains the definition of the project that is used from the `Project.swift`.

The big downside of this second approach is that by doing that we'd need to maintain two specifications. But it also allows us to have a more Swifty representation of the project, and we could leverage things like enum, or even generics! 

What are your thoughts on this @rahul-malik / @yonaskolb ?

yonaskolb:
Nice one @pepibumur!
Interesting question. Either option seems a pretty big sacrifice. Have you looked into how @orta has implemented it in https://github.com/danger/danger-swift? I think that project doesn't have any dependencies though.

In general I'd prefer approach 1. We could move decoding into another module (or XcodeGenKit), and duplicate anything from xcproj we rely on. I'm wary of duplicating the project spec.
Even so, is it all worth it in the end? It's a cool feature, but I'd love to know who would prefer using a Swift file as opposed to YAML/JSON.

@lyptt do you have any stats on how many people use yaml/json vs ruby in https://github.com/lyptt/struct?

pepibumur:
I'm also a bit skeptical about adding support for it. The way @orta does it in Danger is providing a library, `libDanger` that needs to be imported in the `Danger.swift`. The only benefit that I see of using the `.swift` approach is that you could edit the specification file using Xcode and make sure that the structure is correct *(since Xcode will verify the types)*. Besides that, I don't see much benefit since we can get the same using `yaml`.

I'm not a big fan of maintaining two specs either. That comes with maintenance costs and potential inconsistencies between the two specs formats.

Unless we see a lot of value on it I don't think it's worth the effort.

PR #172, opened 2017-11-26
Add commandlineArguments to XcodeGenKit Scheme specifications

rahul-malik:
These are the necessary changes after this PR in `xcproj` lands:
https://github.com/xcodeswift/xcproj/pull/167

### DIFF
--- a/Docs/ProjectSpec.md
+++ b/Docs/ProjectSpec.md
@@ -326,7 +326,7 @@ This is a convenience used to automatically generate schemes for a target based 
 - 🔵 **configVariants**: `[String]` - This generates a scheme for each entry, using configs that contain the name with debug and release variants. This is useful for having different environment schemes.
 - ⚪️ **testTargets**: `[String]` - a list of test targets that should be included in the scheme. These will be added to the build targets and the test entries
 - ⚪️ **gatherCoverageData**: `Bool` - a boolean that indicates if this scheme should gather coverage data
-- ⚪️ **commandLineArguments**: `` - a dictionary from the argument name (`String`) to if it is enabled (`Bool`). These arguments will be added to the Test, Profile and Run scheme actions
+- ⚪️ **commandLineArguments**: `[String:Bool]` - a dictionary from the argument name (`String`) to if it is enabled (`Bool`). These arguments will be added to the Test, Profile and Run scheme actions
 
 For example, the spec below would create 3 schemes called:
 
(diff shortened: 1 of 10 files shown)


yonaskolb:
@rahul-malik I’d rather not merge anything that doesn’t point to a specific depndency version

Sources/ProjectSpec/Target.swift
```diff
@@ -130,11 +130,13 @@ public struct TargetScheme {
     public var testTargets: [String]
     public var configVariants: [String]
     public var gatherCoverageData: Bool
+    public var commandLineArguments: [String: Bool]
```
yonaskolb: Would anyone want to specify disabled arguments? In other words is there a need to make this a dictionary as opposed to an array?

rahul-malik:
@yonaskolb - should i revert then? i'm assuming `xcproj 1.6.0` isn't coming until the weekend

Sources/ProjectSpec/Target.swift
yonaskolb: commandLineArguments or just arguments? What does the Xcode interface use? (on my phone at the moment)

Docs/ProjectSpec.md
```diff
@@ -354,6 +355,9 @@ targets
         - Staging
         - Production
       gatherCoverageData: true
+      commandLineArguments:
```
yonaskolb: Most arguments will use “-myArgument” format, so its good to show quotes are needed

rahul-malik: 👍  I can update with quotes

Sources/ProjectSpec/Target.swift
rahul-malik: - Disabled arguments are actually useful if you have different modes your program can run in but their behavior is conflicting. The advantage is that you could define all the configurations in the schema and let developers toggle them on/off during development
- The element in `xcscheme` is `CommandLineArguments` so it's based off that. The Xcode UI has a more verbose `Arguments passed on launch`

Docs/ProjectSpec.md
rahul-malik: PR here: https://github.com/yonaskolb/XcodeGen/pull/176

Sources/ProjectSpec/Target.swift
yonaskolb: Perhaps `launchArguments`? What do you think?

rahul-malik: I can submit an update for launch arguments

yonaskolb: Ok cool

yonaskolb:
We’ll leave it for now, but a versioned release will need to be referenced before an xcodegen release is made

PR #174, opened 2017-11-27
Print version with -v too

kastiglione:
Would it be reasonable to support `-v` in addition to `--version`?

### DIFF
--- a/Sources/XcodeGen/main.swift
+++ b/Sources/XcodeGen/main.swift
@@ -8,7 +8,12 @@ import JSONUtilities
 
 let version = "1.4.0"
 
-func generate(spec: String, project: String, isQuiet: Bool) {
+func generate(spec: String, project: String, isQuiet: Bool, justVersion: Bool) {
+    if justVersion {
+        print(version)
+        exit(EXIT_SUCCESS)
+    }
+
     let logger = Logger(isQuiet: isQuiet)
 
     func fatalError(_ message: String) -> Never {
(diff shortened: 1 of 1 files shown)


yonaskolb:
It would indeed be reasonable @kastiglione :)
It might clash with a future `--verbose` flag though. And if Commander already makes the decision to restrict to `--version` perhaps that's a good default. What do you think? Are you used to using `-v`?

kastiglione:
Good points. Hopefully Commander wouldn't force that, but I see the concern there.

We have a version of XcodeGen that I want to upgrade to use the latest upstream release. Our XcodeGen has a `-v` flag (but not `--version`), and that flag is used by other tools to tell whether the local copy is out of date and needs an update. I could work around the `-v`/`--version` discrepancy in other ways, but if XcodeGen officially supports `-v` then that makes it easier.

Personally, I never like it when I run `-v` on some command and find out it does nothing and then try `--version` and see that it works. I can see if `-v` had some other meaning, but having no meaning seems unfortunate. For verbose, that could be `-V`, and since CAPS usually means loud, that could be suitable for verbose. I could try to add `-v` to Commander, as another approach.

Do you think XcodeGen is going to have a verbose mode? It already seems verbose/detailed as a default.

yonaskolb:
Some good counter points :)

kastiglione:
Thanks!

kastiglione:
What do you think, should I open a pull request in Commander to see if they'll make `-v` an official alias for `--version`?

yonaskolb:
Yeah, I think it's worth a PR. It should probably be done in a way in that if anyone manually adds a `-v` flag it will use that instead, and only print version as a fallback

PR #175, opened 2017-11-29
Support Legacy Targets

bkase:
Blocked on https://github.com/xcodeswift/xcproj/pull/171

XcodeGen now supports "External Build Tool" target type (internally called
PBXLegacyTarget in Xcode). This is implied when the target type is .none

PBXLegacyTargets are just like normal targets but the `isa` is different
and they have a notion of a buildTool. Most of the target logic doesn't
have to change.

### DIFF
--- a/Docs/ProjectSpec.md
+++ b/Docs/ProjectSpec.md
@@ -141,6 +141,7 @@ Settings are merged in the following order: groups, base, configs.
 - ⚪️ **postbuildScripts**: [[Build Script](#build-script)] - Build scripts that run *after* any other build phases
 - ⚪️ **dependencies**: [[Dependency](#dependency)] - Dependencies for the target
 - ⚪️ **scheme**: [Target Scheme](#target-scheme) - Generated scheme with tests or config variants
+- ⚪️ **legacy**: [Legacy Target](#legacy-target) - When present, opt-in to make an Xcode "External Build System" legacy target instead.
 
 ### Product Type
 This will provide default build settings for a certain product type. It can be any of the following:
@@ -163,6 +164,7 @@ This will provide default build settings for a certain product type. It can be a
 - app-extension.messages
 - app-extension.messages-sticker-pack
 - xpc-service
+- "" (used for legacy targets)
 
 ### Platform
 This will provide default build settings for a certain platform. It can be any of the following:
@@ -361,3 +363,12 @@ targets
   MyUnitTests:
     sources: Tests
 ```
+
+###  Legacy Target
+By providing a legacy target, you are opting in to the "Legacy Target" mode. This is the "External Build Tool" from the Xcode GUI. This is useful for scripts that you want to run as dependencies of other targets, but you want to make sure that it only runs once even if it is specified as a dependency from multiple other targets.
+
+- 🔵 **toolPath**: String - Path to the build tool used in the legacy target.
+- ⚪️ **arguments**: String - Build arguments used for the build tool in the legacy target
+- ⚪️ **passSettings**: Bool - Whether or not to pass build settings down to the build tool in the legacy target.
+- ⚪️ **workingDirectory**: String - The working directory under which the build tool will be invoked in the legacy target.
+
--- a/Sources/ProjectSpec/Target.swift
+++ b/Sources/ProjectSpec/Target.swift
@@ -2,6 +2,22 @@ import Foundation
 import xcproj
 import JSONUtilities
 
+public struct LegacyTarget {
+    public var toolPath: String
+    public var arguments: String?
+    public var passSettings: Bool
+    public var workingDirectory: String?
+}
+
+extension LegacyTarget: Equatable {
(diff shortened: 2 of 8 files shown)


Sources/ProjectSpec/Target.swift
```diff
@@ -13,6 +13,11 @@ public struct Target {
     public var postbuildScripts: [BuildScript]
     public var configFiles: [String: String]
     public var scheme: TargetScheme?
+    
+    public var buildToolPath: String?
```
yonaskolb: As these won't be used most of the time, can we extract them into a nullable `legacy: LegacyTarget` struct

Docs/ProjectSpec.md
```diff
@@ -164,6 +171,8 @@ This will provide default build settings for a certain product type. It can be a
 - app-extension.messages-sticker-pack
 - xpc-service
 
+Note: Passing "" for the product type implies you want a LegacyTarget
```
yonaskolb: What about if legacy targets need "legacy" as the type? Or alternatively that a "legacy" or "buildTool" dictionary must be defined? Thoughts?

rahul-malik: I like the idea of legacy as the type here. Feels more explicit than the presence of a build tool dictionary

bkase: Unfortunately the product type is defined in `xcproj` not here in `XcodeGen` so we can't add a new variant easily (without building another enum on top of the existing one).

Sources/ProjectSpec/Target.swift
bkase: I like that!

Package.resolved
```diff
@@ -78,8 +78,8 @@
         "repositoryURL": "https://github.com/xcodeswift/xcproj.git",
         "state": {
           "branch": null,
-          "revision": "0a14d97c0a2b3a4bbfc2edf6a0027091aa722d57",
-          "version": "1.6.1"
+          "revision": "2b20ccc8ccaa136c619e62d2d9f8b15029f63a94",
```
bkase: Do not merge until next xcproj release

Docs/ProjectSpec.md
```diff
@@ -141,10 +141,12 @@ Settings are merged in the following order: groups, base, configs.
 - ⚪️ **postbuildScripts**: [[Build Script](#build-script)] - Build scripts that run *after* any other build phases
 - ⚪️ **dependencies**: [[Dependency](#dependency)] - Dependencies for the target
 - ⚪️ **scheme**: [Target Scheme](#target-scheme) - Generated scheme with tests or config variants
+- ⚪️ **legacy**: [Legacy Target](#legacy-target) - When present, opt-in to make an Xcode "External Build System" legacy target instead.
 
 ### Product Type
 This will provide default build settings for a certain product type. It can be any of the following:
 
+- ""
```
yonaskolb: maybe add a `(used for legacy targets)` and move it to the bottom of the list

yonaskolb:
Does an empty target type require the legacy dictionary or vice versa? If so we can add something to the Spec validator

Sources/XcodeGenKit/ProjectGenerator.swift
```diff
@@ -41,9 +41,10 @@ public class ProjectGenerator {
 
         func getBuildEntry(_ buildTarget: Scheme.BuildTarget) -> XCScheme.BuildAction.Entry {
 
-            let targetReference = pbxProject.objects.nativeTargets.referenceValues.first { $0.name == buildTarget.target }!
+            let predicate: (PBXTarget) -> Bool = { $0.name == buildTarget.target }
+            let targetReference = pbxProject.objects.nativeTargets.referenceValues.first{ predicate($0 as PBXTarget) } ?? pbxProject.objects.legacyTargets.referenceValues.first{ predicate($0 as PBXTarget) }!
```
yonaskolb: the next version of xcproj will have a helper function for returning a target by name

bkase:
@yonaskolb with the new legacy struct approach, the type is just ignored completely (so we don't have to put anything in the validator). I just kept the `""` in the docs since it's nice for the legacy target. We can make it stricter if you want, but we might as well accept anything. I'll update the docs on that bit though

Sources/ProjectSpec/Target.swift
```diff
@@ -2,6 +2,22 @@ import Foundation
 import xcproj
 import JSONUtilities
 
+public struct LegacyTarget {
+    public var buildToolPath: String
+    public var buildArgumentsString: String?
```
yonaskolb: Does this need a `String` appendix?

yonaskolb: Another way of naming this could be `LegacyBuild` with the following properties:

- toolPath
- arguments
- passSettings
- workingDirectory

What do you think?

yonaskolb:
Master now points to xcproj 1.7.0

rahul-malik:
@yonaskolb - Saw you already signed off and CI passed so i just went ahead and merged this PR.

PR #178, opened 2017-11-30
Add carthageCopyScript as an option.

peymankh:
Hi @yonaskolb,

This is a good option to have since not all Carthage frameworks such as Fabric, Crashlytics, should be copied.

Let me know your thoughts.

### DIFF
--- a/Sources/ProjectSpec/ProjectSpec.swift
+++ b/Sources/ProjectSpec/ProjectSpec.swift
@@ -21,6 +21,7 @@ public struct ProjectSpec {
 
     public struct Options {
         public var carthageBuildPath: String?
+        public var carthageCopyScript: Bool
         public var createIntermediateGroups: Bool
         public var bundleIdPrefix: String?
         public var settingPresets: SettingPresets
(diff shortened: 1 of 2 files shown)


yonaskolb:
Interesting. Some Carthage frameworks shouldn't be passed to the Carthage copy-frameworks script? Is there any documentation around that?
If so, we could utilize the `embed` property of `Dependency` to control this on a more fine grained per dependency basis

peymankh:
There you go https://github.com/Carthage/Carthage/blob/master/Documentation/StaticFrameworks.md

It is basically when you use use-binaries to include framework statically. 

I think you're right, it should probably go as a property of Dependency itself. I provided this as a quick workaround.

yonaskolb:
Ah I see, so it's about not copying static frameworks. 

Yeah, if you could base this off the dependency `embed` property that would be great. I don't want to clutter up `spec.options` to much if we don't have to.

peymankh:
Fair enough. Will sort it out soon.

Issue #179, opened 2017-12-02
Update current Xcode version


peymankh:
Currently the Xcode version is 0900 where as the latest Xcode should be 0910.

Question is whether this should be supported with as an option or should we continue using the hard coded version.

yonaskolb:
Yeah this could be updated. Does that fix any practical issues?
We could add an option as well

peymankh:
It's really just to avoid Xcode warnings to `Update Project Settings`. 

OK I'll add this as an option with `0910` as default.

rahul-malik:
@peymankh @yonaskolb - Would it make sense to infer this based on the currently selected path for the Xcode toolchain? The advantage there is that you won't have to update this every time there is an update

yonaskolb:
@rahul-malik that’s a good idea. Though sometimes a new Xcode may have breaking changes that have we to accomodate, so we can’t automatically support all future versions. So it might be good to still manually update to newer versions as the fallback

peymankh:
@rahul-malik I think I have to agree with @yonaskolb on this.

PR #182, opened 2017-12-03
Issue  #179. Add currentXcodeVersion as an option. 

peymankh:


Sources/ProjectSpec/ProjectSpec.swift
```diff
@@ -47,7 +50,8 @@ public struct ProjectSpec {
             }
         }
 
-        public init(carthageBuildPath: String? = nil, createIntermediateGroups: Bool = false, bundleIdPrefix: String? = nil, settingPresets: SettingPresets = .all, developmentLanguage: String? = nil) {
+        public init(currentXcodeVersion: String? = nil, carthageBuildPath: String? = nil, createIntermediateGroups: Bool = false, bundleIdPrefix: String? = nil, settingPresets: SettingPresets = .all, developmentLanguage: String? = nil) {
+            self.currentXcodeVersion = currentXcodeVersion ?? Options.defaultXcodeVersion
```
rahul-malik: Do you think most users of this option will know to use the `XCODE_VERSION_ACTUAL` format (`0910`) vs just specifying `9.1`? Should we have some logic to accept multiple formats

peymankh: I personally think that’d be a matter of documentation, most users shouldn’t even need to set this, having multiple formats could be overdoing it IMHO.

Sources/ProjectSpec/ProjectSpec.swift
```diff
@@ -20,6 +20,9 @@ public struct ProjectSpec {
     public var include: [String] = []
 
     public struct Options {
+        fileprivate static let defaultXcodeVersion = "0910"
+        
+        public var currentXcodeVersion: String
```
yonaskolb: Can we name this just `xcodeVersion`. You probably just went with the old name, sure why the old property was `currentXcodeVersion`

yonaskolb: Can we also make this an optional and move the `defaultXcodeVersion` into XcodeGenKit. I'd prefer to keep any hardcoded values out of this module, as some people use it independently and it's really the generator that supports a certain version or not.

Sources/ProjectSpec/ProjectSpec.swift
yonaskolb: I think I agree with @rahul-malik here. If someone defines a string with a single `.` in it we could just join the parts and pad it out to the expected 4 digit number. This can happen in the `XcodeGenKit` through using an extension on `ProjectSpec` that also uses the default fall back as mentioned above.

yonaskolb:
Thanks for working on this @peymankh. I've created a new PR here addressing the comments 👍  #197

peymankh:
Thanks @yonaskolb. Sorry, been having a hectic end of the year, was hoping to get it sorted during the holidays.

PR #184, opened 2017-12-06
Transform input spec path to absolute path

### LINKED ISSUE #183: createIntermediateGroups ends up with broken file refs
With this project configuration:

```yaml
name: Project
options:
targets:
  Project:
    type: application
    platform: iOS
    sources:
      - Modules/Sources
      - Foo/Sources
```

And this folder structure:

```
.
├── Foo
│   └── Sources
│       └── baz.swift
├── Modules
│   └── Sources
│       └── bar.swift
└── project.yml
```

Running XcodeGen results in a broken project file:

screen shot 2017-12-05 at 16 32 41

Example setup: [XcodeGenTest.zip](https://github.com/yonaskolb/XcodeGen/files/1533449/XcodeGenTest.zip)

kastiglione:
Fixes #183

As mentioned in #183, `createIntermediateGroups` fails to work as expected under certain conditions. Specifically, project generation works when the `--spec` path is absolute, but breaks when using a relative path. The solution in this change is to convert the spec path to be absolute, at the beginning of project generation.

### DIFF
--- a/Sources/XcodeGen/main.swift
+++ b/Sources/XcodeGen/main.swift
@@ -21,7 +21,7 @@ func generate(spec: String, project: String, isQuiet: Bool, justVersion: Bool) {
         exit(1)
     }
 
-    let specPath = Path(URL(fileURLWithPath: spec).path).normalize()
+    let specPath = Path(spec).absolute()
     let projectPath = Path(project).normalize()
 
     if !specPath.exists {
--- a/Sources/XcodeGen/main.swift
+++ b/Sources/XcodeGen/main.swift
@@ -21,7 +21,7 @@ func generate(spec: String, project: String, isQuiet: Bool, justVersion: Bool) {
         exit(1)
     }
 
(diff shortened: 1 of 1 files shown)


Sources/XcodeGen/main.swift
```diff
@@ -21,7 +21,7 @@ func generate(spec: String, project: String, isQuiet: Bool, justVersion: Bool) {
         exit(1)
     }
 
-    let specPath = Path(spec).normalize()
+    let specPath = Path(URL(fileURLWithPath: spec).path).normalize()
```
kastiglione: I didn't see any code in PathKit to do this.

kastiglione: I am mistaken, I can change this to `absolute()`.

yonaskolb: Yep, absolute 👍

yonaskolb:
Thanks @kastiglione! 👏

Issue #152, opened 2017-11-15
Add TargetSource.buildPhase


yonaskolb:
This would allow you to override the guessed build phase of a source file or group of source files

PR #187, opened 2017-12-07
Possibility to change the visibility of headres in case of a framework

yageek:
Add the possibility to specify the visibility of the header in case of a framework target.

You can use the `headers` property for that:
```yaml
name: TestProj
targets:
  MyTarget:
    type: framework
    platform: iOS
    sources:
      - path: some_path
    headers:
      public:
        - generated-src/objc/*.h
      private:
        - somefile.hpp
```
If a non framework target uses this property, a `invalidHeaderMapForTarget` error is thrown during the validation pass.
If one of the provided files does not exist, an `invalidHeaderPathForTarget` error is thrown during the validation pass.

Some basics tests have been added for the parsing.

The default behaviour for headers has changed: all headers will be considerated as `Project` by default, unless `headers` tells something else.

### DIFF
--- a/Sources/XcodeGenKit/SourceGenerator.swift
+++ b/Sources/XcodeGenKit/SourceGenerator.swift
@@ -45,10 +45,7 @@ class SourceGenerator {
         let fileReference = fileReferencesByPath[path]!
         var settings: [String: Any] = [:]
         let buildPhase = buildPhase ?? getDefaultBuildPhase(for: path)
-
-        if buildPhase == .headers {
-            settings = ["ATTRIBUTES": ["Public"]]
-        }
+  
         if targetSource.compilerFlags.count > 0 {
             settings["COMPILER_FLAGS"] = targetSource.compilerFlags.joined(separator: " ")
         }
--- a/Tests/XcodeGenKitTests/ProjectSpecTests.swift
+++ b/Tests/XcodeGenKitTests/ProjectSpecTests.swift
@@ -8,7 +8,7 @@ func projectSpecTests() {
     describe("ProjectSpec") {
 
         let framework = Target(name: "MyFramework", type: .framework, platform: .iOS,
-                               settings: Settings(buildSettings: ["SETTING_2": "VALUE"]))
+                               settings: Settings(buildSettings: ["SETTING_2": "VALUE"]), headerMap: Target.HeaderVisibilityMap(publicHeaders: ["Some"]))
         let staticLibrary = Target(name: "MyStaticLibrary", type: .staticLibrary, platform: .iOS,
                                    settings: Settings(buildSettings: ["SETTING_2": "VALUE"]))
         let dynamicLibrary = Target(name: "MyDynamicLibrary", type: .dynamicLibrary, platform: .iOS,
@@ -71,7 +71,8 @@ func projectSpecTests() {
                                        dependencies: [Dependency(type: .target, reference: "invalidDependency")],
                                        prebuildScripts: [BuildScript(script: .path("invalidPrebuildScript"), name: "prebuildScript1")],
                                        postbuildScripts: [BuildScript(script: .path("invalidPostbuildScript"))],
-                                       scheme: TargetScheme(testTargets: ["invalidTarget"])
+                                       scheme: TargetScheme(testTargets: ["invalidTarget"]),
+                                       headerMap: Target.HeaderVisibilityMap(publicHeaders: ["Some"])
                 )]
 
                 try expectValidationError(spec, .invalidTargetDependency(target: "target1", dependency: "invalidDependency"))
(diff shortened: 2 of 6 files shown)


Tests/XcodeGenKitTests/ProjectSpecTests.swift
```diff
@@ -85,7 +86,8 @@ func projectSpecTests() {
 
                 try expectValidationError(spec, .missingConfigTypeForGeneratedTargetScheme(target: "target1", configType: .debug))
                 try expectValidationError(spec, .missingConfigTypeForGeneratedTargetScheme(target: "target1", configType: .release))
-
```
pepibumur: I think it'd be great to also test the other error that you throw when the file doesn't exist. You could do that by injecting a function to the `SpecValidation` with the signature `exists: (Path) -> Bool = { $0.exists}` and updating your code to be:

```swift
if !exists(path) {
  errors.append(.invalidHeaderPathForTarget(target: target.name, path: header))
}
```
Then you could inject that function in the test returning `false`.

Sources/XcodeGenKit/SourceGenerator.swift
```diff
@@ -45,10 +45,7 @@ class SourceGenerator {
         let fileReference = fileReferencesByPath[path]!
         var settings: [String: Any] = [:]
         let buildPhase = buildPhase ?? getDefaultBuildPhase(for: path)
-
-        if buildPhase == .headers {
```
pepibumur: This change will require a major version update because you are changing the behavior of the generation. What do you think about keeping it public by default and discuss further if it makes sense to default it to project in a future major version?

rahul-malik:
Is there a reason to have another section for headers vs an attribute in sources? We already have support for compilation flags, etc that are applied on a subset of files in sources which could be leveraged here.

I think we could make this change by adding a "visibility" option (public, private) to sources. The default option value would be public.

yageek:
I wanted to be as closed as possible to Xcode. Visually, you change it inside the "Build Phases" by choosing which header is either"Project", "Public" or "Private". I wanted to be closed to this.

rahul-malik:
Currently sources represent a lot of things that aren't in the Build Sources section of build phases. It can be a bit ambiguous but if we're going down this route we should probably separate other things like resources to be consistent. 

The current schema seems friendly for users at the expense of not matching Xcode explicitly.

@yonaskolb - thoughts here?

yonaskolb:
Hi guys. Sorry I haven't had time to look at this.

Yes I agree with @rahul-malik, in that keeping everything in our `TargetSources` is the best. It could even be a simple `headerVisibility` enum that defaults to `public`.

@yageek I can see the advantage of the current way you've implemented it though in that you can seperate the types via glob patterns, and are not limited to a single visibility type per source declaration. To achieve something similar, the path itself could handle glob patterns (quite a bit of work), or the whole `headers` structure could be moved down into `TargetSources` and the patterns be relative to the path. The later is probably the best approach for now. Thoughts?

yageek:
Do you mean having something like this?:


```yaml
name: TestProj
targets:
  MyTarget:
    type: framework
    platform: iOS
    sources:
      headers:
        public:
          - generated-src/objc/*.h
        private:
          - somefile.hpp
      - path: some_path
```

yonaskolb:
Yeah. What do you think?

yageek:
From the user points of view, I think it is equivalent to the previous implementation :)
By relative to the path,  Are you still speaking about the path of the `project.yaml`?

Regarding the default visibility to public, I guess teh answer depends on the habbits/type of projects: In general, I have less public headers than "project" or "private" headers. This means, in my case, I would have to write more rules to hide elements than rules to make elements public.

I guess some people encounter the opposite case 🤔 

In case we stay with a default visibility to "public", I would suggest to add also the "project" type inside the enum to give the possibility to the user to fully configured his project.

yonaskolb:
@yageek I've had a think about this.

I think at the moment it would be best to just have a single `headerVisibility: HeaderVisibility` property on `TargetSource` which defaults to `.public`.

As @rahul-malik points out all our source options are currently within a single controlled place with options for the different types. Maybe in the future we could break this out into a `sourceFiles`, `headers`, `resources`, (any maybe `infoPlist`) which have dedicated options, but that would be a larger piece of breaking work and would come later. For now we should adapt to the structure that is already there. 
This means you would create a seperate source entry for each header visibility type, but with `excludes` and #152 (which I want to add in now) it should give you the same control.

```
sources:
   - path: sources/obj
      headerVisibility: private
```

In terms of the default, no matter the structure, I think it has to be public. The aim is for the spec to be as simple as possible for simple cases. That means if you don't touch anything to do with headers, things should "just work" and things will compile (unless you import multiple headers with the same name). It also avoids breaking many current projects.

Let me know if you'd like to work on this. If so, I would say leave this PR here and perhaps create a new one.

yonaskolb:
@yageek I've made an alternative PR #252. Let me know what you think there. Can something that simple in addition to the `exclude` feature meet your needs?

yageek:
I would say that this is a step to the right direction :) It will be possible to configure the different headers as private or public in my case.

yonaskolb:
Having worked on this now and seen it in action, I do also like the approach you've taken here

PR #190, opened 2017-12-13
Support indentation options from xcproj

### LINKED ISSUE #189: Add options for indentWidth and tabWidth
Hi!

If https://github.com/xcodeswift/xcproj/pull/183 lands, xcproj will be able to support encoding the indentWidth and tabWidth directly inside a PBXGroup. This overrides the default settings that a user has specified in Xcode (so it's a way to enforce code consistency in a project).

I think it would be great for XcodeGen to support this, but I'm not sure the best way to introduce it into the ProjectSpec. I feel like we don't need per-group settings (but maybe we should support that?). There should definitely be some sort of setting that can apply to the `topLevelGroup`. I tested this by hardcoding the `indentWidth` and `tabWidth` and those fields do cascade, so setting them on the `topLevelGroup` sets them for the entire project.

The top-level settings right now directly map to Xcconfigs, though, is the right place for this in `options`?

cc @yonaskolb

bkase:
Fixes #189

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -16,7 +16,7 @@ let package = Package(
         .package(url: "https://github.com/yonaskolb/JSONUtilities.git", from: "3.3.0"),
         .package(url: "https://github.com/kylef/Spectre.git", from: "0.8.0"),
         .package(url: "https://github.com/onevcat/Rainbow.git", from: "3.0.0"),
-        .package(url: "https://github.com/xcodeswift/xcproj.git", from: "1.6.0")
+        .package(url: "https://github.com/xcodeswift/xcproj.git", from: "1.7.0")
     ],
     targets: [
         .target(name: "XcodeGen", dependencies: [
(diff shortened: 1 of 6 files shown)


Package.swift
```diff
@@ -16,7 +16,7 @@ let package = Package(
         .package(url: "https://github.com/yonaskolb/JSONUtilities.git", from: "3.3.0"),
         .package(url: "https://github.com/kylef/Spectre.git", from: "0.7.0"),
         .package(url: "https://github.com/onevcat/Rainbow.git", from: "2.1.0"),
-        .package(url: "https://github.com/xcodeswift/xcproj.git", from: "1.6.0")
+        .package(url: "https://github.com/xcodeswift/xcproj.git", .revision("2b20ccc8ccaa136c619e62d2d9f8b15029f63a94"))
```
bkase: Do not merge until xcproj has a new release and we stick it here

yonaskolb:
Master now points to xcproj 1.7.0

rahul-malik:
@yonaskolb - Saw you already signed off and CI passed so i just went ahead and merged this PR.

PR #191, opened 2017-12-16
[WIP] Bump xcproj to 2.0

### LINKED ISSUE #223: Update to xcproj 2.0 or later
Are there plans to upgrade to xcproj 2.0? Are there any known issues/blockers? I want to add XcodeGen support for scheme pre- and post- actions, which will depend on xcodeswift/xcproj#217, which will presumably be in a 2.x.y release.

pepibumur:
Before merging https://github.com/xcodeswift/xcproj/pull/185 this branch will update XcodeGen to use xcproj 2.0.

Resolves #223


pepibumur:
It's great that the integration tests when the integration breaks https://travis-ci.org/xcodeswift/xcproj/builds/317297483?utm_source=github_status&utm_medium=notification

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -37,8 +37,8 @@ public class PBXProjGenerator {
         }
     }
 
-    func addObject(_ object: PBXObject) {
-        proj.objects.addObject(object)
+    func addObject(_ object: PBXObject, reference: String) {
```
yonaskolb: As we already have the instance of the object we can create the reference within this function using the dynamic type. Then we only have to pass in the id for the reference generator. This also lets us easily swap out the generator (as you are doing in this PR)

yonaskolb: In fact because a PBXObject doesn't have a reference anymore, we can just pass in the instance to the generator in xcproj instead of a class type. Then the class name could be looked up in the generator. You could make this change in https://github.com/xcodeswift/xcproj/pull/185.
Then later on the id param in the generator can be become optional, as we can create a hash from all the instance properties

pepibumur: Good idea @yonaskolb 😛. It'll make easier to use hashes when we automate the conformance with Sourcery. Looking forward to it!

pepibumur: @yonaskolb I came across this [part of](https://github.com/yonaskolb/XcodeGen/blob/11e1813474f0e18730a7c45328f79c57ece77f8c/Sources/XcodeGenKit/PBXProjGenerator.swift#L68) the generation where the targets references are generated in advance using the class type. Do you think I should also expose the class-based method or is there any other way you can think of to solve this issue?

yonaskolb: Oh, tricky! Yeah I guess both function would need to be there to support that. One with a `type` param and one with an `object` param.
It also means that if the `addObject` function creates the references, that the targets can't use that function and have to manually add the objects to the project.

BUT another way this could be solved is creating the actual targets instead of simple references and caching them instead, and then late modifying them. We can do that now as the objects are mutable, which they weren't when this was written

yonaskolb:
Closed in favour of #227

PR #194, opened 2017-12-20
Scan for Info.plist lazily

kastiglione:
Prevent unnecessary work from being done while searching for an Info.plist file. For the project I tested with (which specifies `INFOPLIST_FILE` in most cases) this change results in:

* A >50mb reduction in cumulative memory allocated
* A 3-5% time reduction for project generation

This changes the behavior from:

1. Scan to find the Info.plist path for every target, regardless of whether the found path is used
2. Build an array containing every source file, and from that, take the first Info.plist

To:

1. Defer scanning for Info.plist to the time it's needed
2. Stop recursive scanning upon finding the first Info.plist

For number 1, if a config defines `INFOPLIST_FILE`, then there's no need to scan the file system for an Info.plist.

For number 2, since the behavior is to take the first found Info.plist, then there's no need to generate an array containing every descendent file, for all sources, for every target. The search can stop as soon as an Info.plist is found.

### DIFF
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -160,7 +160,7 @@ public class PBXProjGenerator {
             // automatically set INFOPLIST_FILE path
             if !spec.targetHasBuildSetting("INFOPLIST_FILE", basePath: spec.basePath, target: target, config: config) {
                 if searchForPlist {
-                    plistPath = getInfoPlist(spec, target.sources)
+                    plistPath = getInfoPlist(target.sources)
                     searchForPlist = false
                 }
                 if let plistPath = plistPath {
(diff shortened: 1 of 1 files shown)


Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -464,6 +458,20 @@ public class PBXProjGenerator {
...
 
+    func getInfoPlist(_ spec: ProjectSpec, _ sources: [TargetSource]) -> Path? {
+        return sources
+            .lazy
+            .map { spec.basePath + $0.path }
+            .flatMap { (path) -> Path? in
+                if path.isFile {
+                    return path.lastComponent == "Info.plist" ? path : nil
+                } else {
+                    return path.first(where: { $0.lastComponent == "Info.plist" })
```
yonaskolb: Does this look in direct children or all recursive children? If the former it would be a regression as files in sub folders won't be found anymore. It's an edge case, but I have projects that rely on this

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -464,6 +458,20 @@ public class PBXProjGenerator {
         return pbxtarget
     }
 
+    func getInfoPlist(_ spec: ProjectSpec, _ sources: [TargetSource]) -> Path? {
```
yonaskolb: Spec is an instance property so no need to pass it in 👍

Sources/XcodeGenKit/PBXProjGenerator.swift
kastiglione: Yep. PathKit uses [`FileManager.enumerator(at URL:)`](https://developer.apple.com/documentation/foundation/nsfilemanager/1409571-enumeratoraturl?language=objc) which does "deep" aka recursive enumeration.

PR #195, opened 2017-12-20
Added support for testables in `Scheme`

### LINKED ISSUE #193: Can't test target as `spec.schemes` doesn't generate testables
Sample code:

```
schemes:
  STCore:
    build:
      targets:
        - target: STCore
          buildTypes: all
        - target: STCoreTests
          buildTypes: testing
    test:
      config: Staging-Debug
    run:
      config: Staging-Debug
    analyze:
      config: Staging-Debug
```

Resulting `STCore` Xcode scheme doesn't have any `Testable` set so it's impossible to test scheme `STCore`

I've checked on 1.4.0 but after a quick look into the code I think it is still presented on master.

vhbit:
Fixes #193

This addresses the issue that it was impossible to test targets if
custom scheme was used by providing array of testables to `Test` action.

Now it should be possible using the following syntax:

```yaml
STCore:
  build:
    targets:
      - target: STCore
        buildTypes: all
  test:
    testables: [STCoreTests]
    config: Staging-Debug
```

Note that original example in #193 has to be modified for this change
as it causes a confusion in Xcode: if we have 2 targets `STCore` and
`STCoreTests`, if `STCoreTests` is also added as testable Xcode
duplicates test target in build actions (i.e. `STCore`, `STCoreTests`,
`STCoreTests` are shown).

### DIFF
--- a/Sources/ProjectSpec/Scheme.swift
+++ b/Sources/ProjectSpec/Scheme.swift
@@ -24,12 +24,12 @@ public struct Scheme: Equatable {
         self.archive = archive
     }
 
-    public init(name: String, targets: [BuildTarget], debugConfig: String, releaseConfig: String, gatherCoverageData: Bool = false, commandLineArguments: [String: Bool] = [:]) {
+    public init(name: String, targets: [BuildTarget], debugConfig: String, releaseConfig: String, gatherCoverageData: Bool = false, commandLineArguments: [String: Bool] = [:], testables: [String] = []) {
         self.init(
             name: name,
             build: .init(targets: targets),
             run: .init(config: debugConfig, commandLineArguments: commandLineArguments),
-            test: .init(config: debugConfig, gatherCoverageData: gatherCoverageData, commandLineArguments: commandLineArguments),
+            test: .init(config: debugConfig, gatherCoverageData: gatherCoverageData, commandLineArguments: commandLineArguments, targets: testables),
             profile: .init(config: releaseConfig, commandLineArguments: commandLineArguments),
             analyze: .init(config: debugConfig),
             archive: .init(config: releaseConfig)
@@ -65,16 +65,20 @@ public struct Scheme: Equatable {
         public var config: String
         public var gatherCoverageData: Bool
         public var commandLineArguments: [String: Bool]
-        public init(config: String, gatherCoverageData: Bool = false, commandLineArguments: [String: Bool] = [:]) {
+        public var targets: [String]
+
+        public init(config: String, gatherCoverageData: Bool = false, commandLineArguments: [String: Bool] = [:], targets: [String] = []) {
             self.config = config
             self.gatherCoverageData = gatherCoverageData
             self.commandLineArguments = commandLineArguments
+            self.targets = targets
         }
 
         public static func == (lhs: Test, rhs: Test) -> Bool {
             return lhs.config == rhs.config &&
                 lhs.commandLineArguments == rhs.commandLineArguments &&
-                lhs.gatherCoverageData == rhs.gatherCoverageData
+                lhs.gatherCoverageData == rhs.gatherCoverageData &&
+                lhs.targets == rhs.targets
         }
     }
 
@@ -155,6 +159,7 @@ extension Scheme.Test: JSONObjectConvertible {
         config = try jsonDictionary.json(atKeyPath: "config")
         gatherCoverageData = jsonDictionary.json(atKeyPath: "gatherCoverageData") ?? false
         commandLineArguments = jsonDictionary.json(atKeyPath: "commandLineArguments") ?? [:]
+        targets = jsonDictionary.json(atKeyPath: "targets") ?? []
     }
 }
 
--- a/Sources/ProjectSpec/Scheme.swift
+++ b/Sources/ProjectSpec/Scheme.swift
@@ -24,12 +24,12 @@ public struct Scheme: Equatable {
         self.archive = archive
     }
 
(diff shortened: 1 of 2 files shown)


Sources/XcodeGenKit/ProjectGenerator.swift
```diff
@@ -49,13 +49,14 @@ public class ProjectGenerator {
...
-        let testBuildTargets = tests.map {
+        let testableNames = scheme.test?.testables ?? []
+        let testablesBuildTargets = testableNames.map {
             Scheme.BuildTarget(target: $0, buildTypes: BuildType.testOnly)
         }
 
-        let testBuildTargetEntries = testBuildTargets.map(getBuildEntry)
+        let testBuildTargetEntries = testablesBuildTargets.map(getBuildEntry)
 
-        let buildActionEntries: [XCScheme.BuildAction.Entry] = scheme.build.targets.map(getBuildEntry) + testBuildTargetEntries
```
yonaskolb: Does this fix the issue you mention in the description?

Sources/ProjectSpec/Scheme.swift
```diff
@@ -155,6 +159,7 @@ extension Scheme.Test: JSONObjectConvertible {
         config = try jsonDictionary.json(atKeyPath: "config")
         gatherCoverageData = jsonDictionary.json(atKeyPath: "gatherCoverageData") ?? false
         commandLineArguments = jsonDictionary.json(atKeyPath: "commandLineArguments") ?? [:]
+        testables = jsonDictionary.json(atKeyPath: "testables") ?? []
```
yonaskolb: The name `testables` comes from the scheme xml which makes sense. Other name candidates include `testTargets`, or `tests`.
My personal preference would be `tests`, as it's clear and matches the Xcode UI. Thoughts?

Note that in the future

Sources/XcodeGenKit/ProjectGenerator.swift
vhbit: Yep, it does.

Sources/ProjectSpec/Scheme.swift
vhbit: well, I'm fine with both `testTargets` and `tests`, although when I look to the project spec it look a bit repetetive:

```yaml
STCore:
  build:
    targets:
      - target: STCore
        buildTypes: all
  test:
    tests: [STCoreTests]
    config: Staging-Debug
```

that might be a bit less repetitive:
```yaml
STCore:
  build:
    targets:
      - target: STCore
        buildTypes: all
  test:
    targets: [STCoreTests]
    config: Staging-Debug
```

Sources/XcodeGenKit/ProjectGenerator.swift
yonaskolb: Ok. In this case logic needs to be added so that if a TargetScheme defines some tests, they also get added as buildActionEntries.
Down where the Scheme is initialised, the tests need to be appended to the target.

Sources/ProjectSpec/Scheme.swift
yonaskolb: Yeah, I like targets! 👍

Sources/XcodeGenKit/ProjectGenerator.swift
vhbit: they actually are (better seen in split mode):
https://github.com/yonaskolb/XcodeGen/pull/195/files#diff-bcdced0f751833616e896b5b1e6e0fa8R118
and 
https://github.com/yonaskolb/XcodeGen/pull/195/files#diff-bcdced0f751833616e896b5b1e6e0fa8R129

yonaskolb: From what I can see it's passing the test targets in, but they are no longer automatically added as build entries, so you also need to pass them in as targets

vhbit: If you're talking about build entries as shown in "Scheme/Build/Targets" - Xcode automatically adds anything registered as testable there, quote from the PR description:
>  if we have 2 targets `STCore` and `STCoreTests`, if `STCoreTests` is also added as testable Xcode duplicates test target in build actions (i.e. `STCore`, `STCoreTests`,
`STCoreTests` are shown).

yonaskolb: Ah I see. By automatically add, does it create an extra diff in the project, or just display them in the UI?

vhbit: No diff, it just displays them in the UI

Sources/ProjectSpec/Scheme.swift
yonaskolb: So rename `Test.testables` to `targets` (and the yaml property)

Sources/ProjectSpec/Scheme.swift
```diff
@@ -24,12 +24,12 @@ public struct Scheme: Equatable {
         self.archive = archive
     }
 
-    public init(name: String, targets: [BuildTarget], debugConfig: String, releaseConfig: String, gatherCoverageData: Bool = false, commandLineArguments: [String: Bool] = [:]) {
+    public init(name: String, targets: [BuildTarget], debugConfig: String, releaseConfig: String, gatherCoverageData: Bool = false, commandLineArguments: [String: Bool] = [:], testables: [String] = []) {
```
yonaskolb: Looking at this now, I think I will change this init to take a `name: String, targetScheme: TargetScheme, debugConfig: String, releaseConfig: String` and move as an extension in the ProjectGenerator.
But that can happen after this.

PR #200, opened 2017-12-21
Allow missing files in sources

allu22:
Adds `optional` flag to `TargetSource` to allow files that do not exist yet.
This is useful in cases where build script will generate files.

### DIFF
--- a/Tests/XcodeGenKitTests/ProjectSpecTests.swift
+++ b/Tests/XcodeGenKitTests/ProjectSpecTests.swift
@@ -101,6 +101,16 @@ func projectSpecTests() {
                 try expectValidationError(spec, .invalidSchemeConfig(scheme: "scheme1", config: "debugInvalid"))
                 try expectValidationError(spec, .invalidSchemeConfig(scheme: "scheme1", config: "releaseInvalid"))
             }
+
+            $0.it("allows missing optional file") {
+                var spec = baseSpec
+                spec.targets = [Target(name: "target1",
+                                       type: .application,
+                                       platform: .iOS,
+                                       sources: [.init(path: "generated.swift", optional: true)]
+                    )]
+                try spec.validate()
+            }
         }
     }
 }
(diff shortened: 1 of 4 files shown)


yonaskolb:
Yeah I thought someone might bring up the strictness of the file validation eventually :)
In this sources case though, if the source directory doesn't exist nothing will be added, so even if they were generated later it wouldn't change anything. Or would you use this when pointing directly at a source file?
Also is this maybe better as a top level option? `allowMissingSources`?

allu22:
We are pointing directly to file, like so:
``` yaml
sources:
  - path: Assets.generated.swift
    type: file
    optional: true
  - path: GraphQL.generated.swift
    type: file
    optional: true
```

I kind of like to have this missing source check enables for other files as this kind of missing file inclusion is exception.

Tests/XcodeGenKitTests/ProjectSpecTests.swift
```diff
@@ -101,6 +101,16 @@ func projectSpecTests() {
                 try expectValidationError(spec, .invalidSchemeConfig(scheme: "scheme1", config: "debugInvalid"))
                 try expectValidationError(spec, .invalidSchemeConfig(scheme: "scheme1", config: "releaseInvalid"))
             }
+
+            $0.it("allows missing optional file", closure: {
```
yonaskolb: nitpick: we use trailing closure for these 👍 
Thanks for the test! :)

yonaskolb:
Ok great, some good reasons there 👍

Tests/XcodeGenKitTests/ProjectSpecTests.swift
allu22: Uniform style is important. Changed it.

keith:
Here's a small bug related to this change https://github.com/yonaskolb/XcodeGen/issues/210

PR #201, opened 2017-12-21
Schemes

yonaskolb:
As Schemes are in a more fleshed out state, they are now documented


yonaskolb:
@vhbit

yonaskolb:
There are still a lot more options that could be added to Scheme, but this is a good starting point

PR #203, opened 2017-12-21
Change Scheme.Build.targets spec

yonaskolb:
I've just learned that the order of targets in a Scheme's build settings has no affect, and in fact when you reorder them, nothing gets changed in the project.

Within this in mind, this tweaks the Scheme.build.targets to be a bit simpler and mirrors the changes in #54. The change can be seen in the documentation diffs.

As a dictionary wouldn't be in a stable order for diffs, the targets are sorted by name.

Technically this is a breaking change, which is the reason I didn't document Schemes until now, so the format could be locked down.

Any thoughts?

### DIFF
--- a/Sources/ProjectSpec/Scheme.swift
+++ b/Sources/ProjectSpec/Scheme.swift
@@ -189,36 +189,28 @@ extension Scheme: NamedJSONDictionaryConvertible {
 extension Scheme.Build: JSONObjectConvertible {
 
     public init(jsonDictionary: JSONDictionary) throws {
-        targets = try jsonDictionary.json(atKeyPath: "targets")
-    }
-}
-
-extension Scheme.BuildTarget: JSONObjectConvertible {
-
-    public init(jsonDictionary: JSONDictionary) throws {
-        target = try jsonDictionary.json(atKeyPath: "target")
-        if jsonDictionary["buildTypes"] == nil {
-            buildTypes = BuildType.all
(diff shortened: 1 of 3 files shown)


yonaskolb:
@vhbit

Sources/ProjectSpec/Scheme.swift
```diff
@@ -189,36 +189,28 @@ extension Scheme: NamedJSONDictionaryConvertible {
...
+            } else {
+                buildTypes = BuildType.all
             }
-            self.buildTypes = buildTypes
-        } else {
-            let stringBuildTypes: [String] = try jsonDictionary.json(atKeyPath: "buildTypes")
-            self.buildTypes = stringBuildTypes.flatMap(BuildType.from)
+            targets.append(Scheme.BuildTarget(target: target, buildTypes: buildTypes))
         }
+        self.targets = targets.sorted { $0.target < $1.target }
```
rahul-malik: Do we need to sort here? I believe `xcproj` will sort the targets before serializing

yonaskolb: `xcproj` just writes them in the order they are received https://github.com/xcodeswift/xcproj/blob/master/Sources/xcproj/XCScheme.swift#L251

yonaskolb:
Another benefit of this is it's impossible to duplicate targets that will be build. It also play's nice with the `include` feature, as you can override a single target's settings easily

PR #205, opened 2017-12-22
Add deployment target

yonaskolb:
This changes how platform versions are applied.

- removed all platform versions from the setting presets. This lets Xcode determine the latest version
- added `options.platformVersions` to set project level versions
- added `target.platformVersion` to set application specific version

```
options:
  platformVersions:
    watchOS: 3.0
    tvOS: 11.0
targets:
  App_iOS:
    type: application
    platform: iOS
    platformVersion: 10.0
```

These new options just apply default build settings, any any custom settings will still override this.

This is technically a breaking change, as those projects relying on the old default versions set in the setting presets will now have the latest platform version, which for example upgrade iOS 10 to iOS 11.1 (depending on Xcode version)

EDIT: renamed to `deploymentTarget`


yonaskolb:
Is`sdkVersion` a better name than `platformVersion`

allu22:
I think `platformVersion` is ok. `sdkVersion` should be avoided as it may create confusion with `SDKROOT` build setting.

```
SDKROOT = iphoneos11.2;
IPHONEOS_DEPLOYMENT_TARGET = 11.2;
```

yonaskolb:
@allu22 Yeah fair enough. Another possible name would be `deploymentTarget` to match Xcode's UI terminology. 

Also the properties on `options.settingPresets` could potentially just live directory on `options` with names like `iOSVersion` (or `iOSDeploymentTarget)`. That means it's a bit cleaner if a spec just has one platform

Sources/ProjectSpec/ProjectSpec.swift
```diff
@@ -51,7 +52,7 @@ public struct ProjectSpec {
             }
         }
 
-        public init(carthageBuildPath: String? = nil, createIntermediateGroups: Bool = false, bundleIdPrefix: String? = nil, settingPresets: SettingPresets = .all, developmentLanguage: String? = nil, indentWidth: Int? = nil, tabWidth: Int? = nil, usesTabs: Bool? = nil, xcodeVersion: String? = nil) {
+        public init(carthageBuildPath: String? = nil, createIntermediateGroups: Bool = false, bundleIdPrefix: String? = nil, settingPresets: SettingPresets = .all, developmentLanguage: String? = nil, indentWidth: Int? = nil, tabWidth: Int? = nil, usesTabs: Bool? = nil, xcodeVersion: String? = nil, platformVersions: PlatformVersions = .init()) {
```
toshi0383: I think you should consider adding more new-lines.
It's hard for reviewers to scroll horizontally to see what has changed.

Tests/Fixtures/TestProject/Project.xcodeproj/project.pbxproj
```diff
@@ -953,6 +948,7 @@
 				SWIFT_OPTIMIZATION_LEVEL = "-Owholemodule";
 				SWIFT_VERSION = 4.0;
 				VALIDATE_PRODUCT = YES;
+				WATCHOS_DEPLOYMENT_TARGET = 2.0;
```
toshi0383: Is this on purpose?

yonaskolb: Yeah, the TestProject spec adds this

Sources/ProjectSpec/PlatformVersions.swift
```diff
@@ -0,0 +1,74 @@
...
+        case .watchOS: return watchOS
+        case .macOS: return macOS
+        }
+    }
+
+    public static func == (lhs: PlatformVersions, rhs: PlatformVersions) -> Bool {
+        return lhs.iOS == rhs.iOS &&
+            lhs.tvOS == rhs.tvOS &&
+            lhs.watchOS == rhs.watchOS &&
+            lhs.macOS == rhs.macOS
```
rahul-malik: This will probably be fine but will fail for instances where you might compare "11" and "11.0" since we are dealing with string equality vs a structured representation of a semantic version

Sources/ProjectSpec/PlatformVersions.swift
```diff
@@ -0,0 +1,74 @@
...
+        }
+    }
+}
+
+extension PlatformVersions: JSONObjectConvertible {
+
+    public init(jsonDictionary: JSONDictionary) throws {
+
+        func parseVersion(_ platform: String) -> String? {
+            if let string: String = jsonDictionary.json(atKeyPath: .key(platform)) {
```
rahul-malik: Should we just have the version input type as `String` for simplicity?

Sources/ProjectSpec/ProjectSpec.swift
rahul-malik: +1, is there a swift format setting to do this automatically?

Tests/XcodeGenKitTests/SpecLoadingTests.swift
```diff
@@ -229,9 +229,17 @@ func specLoadingTests() {
...
+                                              developmentLanguage: "ja",
+                                              platformVersions: PlatformVersions(iOS: "11.1", tvOS: "10.0", watchOS: "3.0", macOS: "10.12"))
             let expected = ProjectSpec(basePath: "", name: "test", options: options)
-            let parsedSpec = try getProjectSpec(["options": ["carthageBuildPath": "../Carthage/Build", "bundleIdPrefix": "com.test", "createIntermediateGroups": true, "developmentLanguage": "ja"]])
+            let dictionary: [String: Any] = ["options": [
+                "carthageBuildPath": "../Carthage/Build",
+                "bundleIdPrefix": "com.test",
+                "createIntermediateGroups": true,
+                "developmentLanguage": "ja",
+                "platformVersions": ["iOS": 11.1, "tvOS": 10.0, "watchOS": "3.0", "macOS": "10.12" ]
```
rahul-malik: Might be good for the test to have the case where we have 3 components to the version ("10.12.1" for example)

Sources/ProjectSpec/PlatformVersions.swift
yonaskolb: I just wanted to support the easy format of just `iOS:  11.1` without wrapping it in a string

Sources/ProjectSpec/PlatformVersions.swift
yonaskolb: That's true. We could potentially implement a proper Version type (or use a dependency, like Swift PM), but at the moment it's probably fine

yonaskolb:
I think I will change this to `deploymentTarget`. Any objections?

Sources/ProjectSpec/PlatformVersions.swift
pepibumur: I think you could copy this implementation that I made for Sake https://github.com/xcodeswift/sake/blob/master/Sources/SakefileUtils/Version.swift that has the equatable defined. That'd solve this issue easily.

Sources/ProjectSpec/PlatformVersions.swift
```diff
@@ -0,0 +1,74 @@
...
+        return lhs.iOS == rhs.iOS &&
+            lhs.tvOS == rhs.tvOS &&
+            lhs.watchOS == rhs.watchOS &&
+            lhs.macOS == rhs.macOS
+    }
+}
+
+extension Platform {
+
+    public var versionBuildSetting: String {
```
pepibumur: I think it'd be great to unit test that this map is properly done.

Sources/ProjectSpec/ProjectSpec.swift
yonaskolb: Will do 👍

Sources/ProjectSpec/PlatformVersions.swift
yonaskolb: Done. One change I had to make was have a getter on that version type that returns an Apple deployment version, as they use `9.0` not `9.0.0`

PR #206, opened 2017-12-22
Add TargetSource.buildPhase for overriding build phase

yonaskolb:
This lets you override the build phase for a source file or all the files in a directory.
This is useful if XcodeGen makes the wrong guesses based off the file extension. 

I think we have a pretty stable list of files right now, but this gives us more flexibility. Especially as we are treating all unknown file types as resources by default, which may not be the desired behaviour.

### DIFF
--- a/Sources/ProjectSpec/TargetSource.swift
+++ b/Sources/ProjectSpec/TargetSource.swift
@@ -1,6 +1,7 @@
 import Foundation
 import JSONUtilities
 import PathKit
+import xcproj
 
 public struct TargetSource {
 
@@ -10,6 +11,23 @@ public struct TargetSource {
     public var excludes: [String]
     public var type: SourceType?
     public var optional: Bool
+    public var buildPhase: BuildPhase?
+
+    public enum BuildPhase: String {
+        case sources
+        case headers
+        case resources
+        case none
+
+        public var buildPhase: xcproj.BuildPhase? {
+            switch self {
+            case .sources: return .sources
+            case .headers: return .headers
+            case .resources: return .resources
+            case .none: return nil
+            }
+        }
+    }
 
     public enum SourceType: String {
         case group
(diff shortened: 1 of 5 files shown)


Tests/XcodeGenKitTests/ProjectGeneratorTests.swift
```diff
@@ -571,6 +571,89 @@ func projectGeneratorTests() {
...
+                    - file.tpp
+                    - file.hxx
+                    - file.def
+                    - file.xcconfig
+                    - file.entitlements
+                    - file.gpx
+                    - file.apns
+                    - file.123
+                    - file.xcassets
+                    - Info.plist
```
toshi0383: What happens if it's `MyApp_Info.plist` instead of `Info.plist` ?

yonaskolb: It’s treated as a resource. Would be a good test case as well 👍

pepibumur: I was about to comment the same. I wouldn't assume that `Info.plist` is the name of the target info.plist file. Is it feasible to get a list of all the info.plists defined in the yaml and us that to filter these .plist out from the resources?

Sources/ProjectSpec/TargetSource.swift
```diff
@@ -10,20 +11,38 @@ public struct TargetSource {
...
+    }
 
     public enum SourceType: String {
         case group
         case file
         case folder
     }
 
-    public init(path: String, name: String? = nil, compilerFlags: [String] = [], excludes: [String] = [], type: SourceType? = nil, optional: Bool = false) {
+    public init(path: String, name: String? = nil, compilerFlags: [String] = [], excludes: [String] = [], type: SourceType? = nil, optional: Bool = false, buildPhase: BuildPhase? = nil) {
```
pepibumur: I think @toshi0383 mentioned something in another PR but what do you think about splitting this into multiple lines to make it easier to read. Out of curiosity, is there any linter being passed on CI in every PR? That could help keeping method names short.

Tests/XcodeGenKitTests/ProjectGeneratorTests.swift
yonaskolb: Yeah it's a tricky problem. I think it's fair to say that if we find a `Info.plist`, it's being used for that purpose. But yes, some projects may also have other names. It would be possible to check all the build settings for any `INFOPLIST_FILE` settings, and then ignore those. I've created a seperate issue for this #207

Sources/ProjectSpec/TargetSource.swift
yonaskolb: At the moment no linter is run on CI. There's a `format_code` make task though that I occasionally run. This is yet to be automated. Along with the automation, rules for line wrapping can be added 👍

PR #216, opened 2017-12-29
Stop validating missing configurations

keith:
Previously if you had a XcodeGen definition file that has multiple
project.yml consumers where one project has one set of configurations,
and the other project has different configurations, you could not
generate the project unless the configurations overlapped perfectly.

For example if `project1.yml` has configurations Debug and Release, but
`project2.yml` has configurations Debug, Alpha, and Release, you want
your shared dependency to have xcconfig files for Debug, Alpha, and
Release, even though project1 will not consume the Alpha configuration.

The downside to this change is that typos such as `Deubg` would no
longer trigger a validation error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Master
 
+#### Changed
+- stop validating configurations that do not exist the project
+
 ## 1.5.0
 
 #### Added
--- a/Sources/ProjectSpec/SpecValidation.swift
+++ b/Sources/ProjectSpec/SpecValidation.swift
@@ -36,9 +36,6 @@ extension ProjectSpec {
             if !(basePath + configFile).exists {
                 errors.append(.invalidConfigFile(configFile: configFile, config: config))
             }
-            if getConfig(config) == nil {
-                errors.append(.invalidConfigFileConfig(config))
-            }
         }
 
         for settings in settingGroups.values {
@@ -55,9 +52,6 @@ extension ProjectSpec {
             for (config, configFile) in target.configFiles {
                 if !(basePath + configFile).exists {
                     errors.append(.invalidTargetConfigFile(target: target.name, configFile: configFile, config: config))
-                }
-                if getConfig(config) == nil {
-                    errors.append(.invalidConfigFileConfig(config))
                 }
             }
 
--- a/Sources/ProjectSpec/SpecValidationError.swift
+++ b/Sources/ProjectSpec/SpecValidationError.swift
@@ -17,7 +17,6 @@ public struct SpecValidationError: Error, CustomStringConvertible {
         case invalidSettingsGroup(String)
         case invalidBuildScriptPath(target: String, name: String?, path: String)
         case invalidFileGroup(String)
-        case invalidConfigFileConfig(String)
         case missingConfigForTargetScheme(target: String, configType: ConfigType)
 
         public var description: String {
@@ -46,8 +45,6 @@ public struct SpecValidationError: Error, CustomStringConvertible {
                 return "Target \(target.quoted) has a script \(name != nil ? "\(name!.quoted) which has a " : "")path that doesn't exist \(path.quoted)"
             case let .invalidFileGroup(group):
                 return "Invalid file group \(group.quoted)"
-            case let .invalidConfigFileConfig(config):
-                return "Config file has invalid config \(config.quoted)"
             case let .missingConfigForTargetScheme(target, configType):
                 return "Target \(target.quoted) is missing a config of type \(configType.rawValue) to generate its scheme"
             }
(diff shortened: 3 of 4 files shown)


keith:
I'm opening this PR for discussion, this is a problem that affects us, but there might be a better solution that doesn't involve losing this typo validation.

yonaskolb:
@keith I'm not against this change, but maybe there is a different solution. Are you using the `include` feature, or what is your setup? If so, can you just define the configs in the final file?

keith:
We have configurations per target (for some targets) defined in the included files. Which is the problem. We also have top level configs so those don’t have this problem.

yonaskolb:
I see. Would adding those target `configFile` properties only in the final files solve it? Is that too much duplication? `include` merges everything, so you can add or change properties deep down in a hierarchy (there is also a syntax for replacing a whole node)

yonaskolb:
If I'm misunderstanding your setup, maybe you could post some example snippets?

keith:
Here's an example of our configuration:

```yaml
# project1.yml
name: project1
configs:
  Debug: debug
  Alpha: release
  Release: release
configFiles:
  Debug: debug.xcconfig
  Alpha: alpha.xcconfig
  Release: release.xcconfig
include:
  - shared.yml
```

```yaml
# project2.yml
name: project2
configs:
  Debug: debug
  Release: release
configFiles:
  Debug: debug.xcconfig
  Release: release.xcconfig
include:
  - shared.yml
```

```yaml
# shared.yml
targets:
  shared:
    type: framework
    platform: iOS
    sources:
      - Shared/Sources
    configFiles:
      Debug: Shared_debug.xcconfig
      Alpha: Shared_alpha.xcconfig
      Release: Shared_release.xcconfig
```

So here we have 2 main `project.yml` files, and they both share a framework that has its own xcconfig files. In `shared.yml` we have 3 xcconfigs, even though `project2` wouldn't use the `Alpha` configuration. This is what currently fails.

yonaskolb:
Ok yeah, so one way of solving that would be defining the `configFiles` in project1 and project2 as follows:

```yaml
# project1.yml
name: project1
configs:
  Debug: debug
  Alpha: release
  Release: release
configFiles:
  Debug: debug.xcconfig
  Alpha: alpha.xcconfig
  Release: release.xcconfig
targets:
  shared:
    configFiles:
      Debug: Shared_debug.xcconfig
      Alpha: Shared_alpha.xcconfig
      Release: Shared_release.xcconfig
include:
  - shared.yml
```

```yaml
# project2.yml
name: project2
configs:
  Debug: debug
  Release: release
configFiles:
  Debug: debug.xcconfig
  Release: release.xcconfig
targets:
  shared:
    configFiles:
      Debug: Shared_debug.xcconfig
      Release: Shared_release.xcconfig
include:
  - shared.yml
```

```yaml
# shared.yml
targets:
  shared:
    type: framework
    platform: iOS
    sources:
      - Shared/Sources
```

keith:
Yea that would definitely work. We were trying to avoid duplicating any configuration here since we actually have many shared modules and the path to the xcconfigs would be the same (for the shared configurations) for all of them.

yonaskolb:
Understandable. You could also keep any common configs in the shared file.

```yaml
# project1.yml
name: project1
configs:
  Debug: debug
  Alpha: release
  Release: release
configFiles:
  Debug: debug.xcconfig
  Alpha: alpha.xcconfig
  Release: release.xcconfig
targets:
  shared:
    configFiles:
      Alpha: Shared_alpha.xcconfig
include:
  - shared.yml
```

```yaml
# project2.yml
name: project2
configs:
  Debug: debug
  Release: release
configFiles:
  Debug: debug.xcconfig
  Release: release.xcconfig
include:
  - shared.yml
```

```yaml
# shared.yml
targets:
  shared:
    type: framework
    platform: iOS
    sources:
      - Shared/Sources
    configFiles:
      Debug: Shared_debug.xcconfig
      Release: Shared_release.xcconfig
```

I understand though if you don't want to split up the target configuration. If removing some of the checks is a better solution for you I suggest we maybe have some sort of option to disable them, rather than removing them outright.

The common types of validation errors you might want to disable would be `missingConfigs` and `missingFiles`.
- `missingConfigs` would disable any validation that checks for the existence of a config (including your configFiles)
- `missingFiles` would disable any validation checks for the existence of files. It would also serve as a catch all solution to #200
```
options:
  disableValidation:
    - missingFiles
    - missingConfigs
```
What do you think?

keith:
I'm fine with those new error disables if that's the way you want to go!

keith:
Submitted https://github.com/yonaskolb/XcodeGen/pull/220 with an attempt at that approach

yonaskolb:
Closing in favour of #220

PR #217, opened 2018-01-03
Filter out duplicate file references from sources build phase

### LINKED ISSUE #210: Optional file paths with existing files get duplicated
With this project configuration:

```yaml
name: Project
targets:
  Project:
    type: application
    platform: iOS
    sources:
      - Sources
      - path: Sources/Foo.swift
        optional: True
```

And this folder structure:

```
.
├── Sources
│   └── Foo.swift
└── project.yml

1 directory, 2 files
```

`Foo.swift` is included twice if it already exists at the time the project file is generated.

Here's an example project demonstrating the issue: [XcodeGenTest.zip](https://github.com/yonaskolb/XcodeGen/files/1588462/XcodeGenTest.zip)

allu22:
Fixes #210

### DIFF
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -408,6 +408,11 @@ public class PBXProjGenerator {
         func getBuildFilesForPhase(_ buildPhase: BuildPhase) -> [String] {
             let files = sourceFiles
                 .filter { $0.buildPhase == buildPhase }
+                .reduce(into: [SourceFile]()) { (output, sourceFile) in
+                  if !output.contains(where: { $0.fileReference == sourceFile.fileReference }) {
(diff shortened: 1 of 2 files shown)


Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -409,6 +409,11 @@ public class PBXProjGenerator {
             let files = sourceFiles
                 .filter { $0.buildPhase == buildPhase }
                 .sorted { $0.path.lastComponent < $1.path.lastComponent }
+                .reduce(into: [SourceFile]()) { (output, sourceFile) in
```
yonaskolb: Let's move this before sorted

allu22: Done.

PR #218, opened 2018-01-03
Change default LD_RUNPATH_SEARCH_PATHS to string

allu22:
XCode will convert them to space separated string anyway.
This way we avoid generating unnecessary difference in file.


yonaskolb:
Looks like Xcode 9.3 might change this back https://twitter.com/xcodereleases/status/960946647957032960

allu22:
@yonaskolb it seems it will an array for "Xcode 9.3 - compatible" (`objectVersion = 50`, `			compatibilityVersion = "Xcode 9.3";`) projects.
But as Xcode 9.2 is unable to open this new format, i think we should not switch to this new format for time being.

I was thinking of opening ticket in xcproj to track changes for this new format(?), but haven't gotten to it yet.

PR #219, opened 2018-01-04
Add Support for ResourceBundles

anreitersimon:
Swift now allows static-libraries which have one drawback when compared to Frameworks.
They cannot inlcude Resources (Images, xibs, storyboards, ...).
To workaround this issue its possible to use resource-bundles.

This PR adds another type of Dependency: __ResourceBundles__

They can be included in two ways:

Embed ResourceBundle Target (When in same xcodeproj)
`Dependency(type: .target, reference: "MyResourceBundle")`

or add from another project embedded in the same xcworkspace
`Dependency(type: .resourceBundle, reference: "MyResourceBundle.bundle", implicit: true)`

### DIFF
--- a/Sources/ProjectSpec/Dependency.swift
+++ b/Sources/ProjectSpec/Dependency.swift
@@ -30,6 +30,7 @@ public struct Dependency: Equatable {
         case target
         case framework
         case carthage
+        case resourceBundle
     }
 
     public static func == (lhs: Dependency, rhs: Dependency) -> Bool {
@@ -65,6 +66,9 @@ extension Dependency: JSONObjectConvertible {
         } else if let carthage: String = jsonDictionary.json(atKeyPath: "carthage") {
             type = .carthage
             reference = carthage
+        } else if let resourceBundle: String = jsonDictionary.json(atKeyPath: "resourceBundle") {
+            type = .resourceBundle
+            reference = resourceBundle
         } else {
             throw SpecParsingError.invalidDependency(jsonDictionary)
         }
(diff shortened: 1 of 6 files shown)


keith:
We’ve been successfully using just the bundle type that already exists. Is there any issue with that?

anreitersimon:
Oh i wasnt aware this is possible right now.

Could you quickly elaborate how you set this up: 
Are you using a folder with `.bundle` extension or a separate target

My original goal was to be able to depend on resource bundles from other `.xcodeproj` residing in the same `.xcworkspace`.

It was my understanding that the only way to express this is by adding a `Dependency` with `implicit: true`

and none of the types defined in `DependencyType`

```swift
public enum DependencyType {
    case target
    case framework
    case carthage
}
```

seemed to fit.

keith:
Hmm, our use case the resource bundles are managed by XcodeGen as well, so that might be a bit different, but for them we use the `target` dependency version of this:

```
- target: FooResources
   embed: True
```

anreitersimon:
Thanks.

So it seems the changes regarding embedding the targets were not necessary at all.

I still think embedding bundles from other `.xcodeproj` would be useful.

(Also a probaby more appropriate naming for the new `DependencyType` would just be `bundle` instead of `resourceBundle`)

yonaskolb:
Is there a reason that this doesn't work?
```
- target: FooResources
  embed: true
  implicit: true

**Comment by anreitersimon** [comment]
Currently this is not working.
I get a invalid target dependency error. (Because the target cannot be found since its in another .xcodeproj)
Even when disabling the check it does not embed the bundle into an app-target.

**Comment by yonaskolb** [comment]
Does anyone have any thoughts on this? I'm not familiar with resource bundles

**Comment by pepibumur** [comment]
Reading the original description I think it can be solved easily by just creating a static framework rather than a static library. @anreitersimon have you tried using framework as a type and using the following build-setting?

```
MACH_O_TYPE = staticlib;
```

PR #220, opened 2018-01-04
Add new option for disabling validations

keith:
This gives us an entry point for disabling specific project.yml
validations that sometimes shouldn't apply. The first example is
missingConfigs, which, if you use your .yml file in multiple projects,
can be too strict if the projects have different top level
configurations.

### DIFF
--- a/Sources/ProjectSpec/ProjectSpec.swift
+++ b/Sources/ProjectSpec/ProjectSpec.swift
@@ -24,6 +24,7 @@ public struct ProjectSpec {
         public var createIntermediateGroups: Bool
         public var bundleIdPrefix: String?
         public var settingPresets: SettingPresets
+        public var disabledValidations: [ValidationType]
         public var developmentLanguage: String?
         public var usesTabs: Bool?
         public var tabWidth: Int?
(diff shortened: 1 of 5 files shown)


keith:
This replaces https://github.com/yonaskolb/XcodeGen/pull/216

Sources/ProjectSpec/ProjectSpec.swift
```diff
@@ -52,6 +53,10 @@ public struct ProjectSpec {
             }
         }
 
+        public enum DisabledValidations: String {
```
yonaskolb: Let's rename to `ValidationType` and move to `SpecValidation.swift`?

Sources/ProjectSpec/SpecValidation.swift
```diff
@@ -36,7 +36,7 @@ extension ProjectSpec {
             if !(basePath + configFile).exists {
                 errors.append(.invalidConfigFile(configFile: configFile, config: config))
             }
-            if getConfig(config) == nil {
+            if !options.disabledValidations.contains(.missingConfigs) && getConfig(config) == nil {
```
yonaskolb: I guess we can also add this check to https://github.com/yonaskolb/XcodeGen/blob/master/Sources/ProjectSpec/SpecValidation.swift#L20

keith:
Updated!

keith:
Thanks!

PR #221, opened 2018-01-05
Generate UI Test Target Attributes

anreitersimon:
It seems that iOS UITest targets generated with XCodeGen fail to run (tested with Xcode 9.2).

The `TargetApplication` Setting for the UITest target isn't configured correctly.

Ive noticed toggling this setting in Xcode results in these diffs.

```
isa = PBXProject;
	attributes = {
		LastUpgradeCheck = 0920;
+		TargetAttributes = {
+			NT_493099431306 = {
+				TestTargetID = NT_791681633892;
+			};
+		};
	};
``` 

This appears to be some sort of look-up table for determining what target to test.

`NT_493099431306` is the file-reference for the UITestTarget
`NT_791681633892 ` is the file-reference for the target matching the UITestTargets `TEST_TARGET_NAME`

This PR adds generation of these target attributes

### DIFF
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -159,7 +159,7 @@ public class PBXProjGenerator {
 
         let projectAttributes: [String: Any] = ["LastUpgradeCheck": spec.xcodeVersion]
             .merged(spec.attributes)
-            .merged(self.targetAttributes)
+            .merged(self.generateTargetAttributes() ?? [:])
         
         let root = PBXProject(
             name: spec.name,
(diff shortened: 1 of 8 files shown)


Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -173,6 +176,42 @@ public class PBXProjGenerator {
...
+        // look up TEST_TARGET_NAME build setting
+        func testTargetName(_ target: PBXTarget) -> String? {
+            guard let configurationList = target.buildConfigurationList else { return nil }
+            guard let buildConfigurationReferences = self.proj.objects.configurationLists[configurationList]?.buildConfigurations else { return nil }
+            
+            let configs = buildConfigurationReferences
+                .flatMap { ref in self.proj.objects.buildConfigurations[ref] }
+            
+            return configs
+                .flatMap { $0.buildSettings["TEST_TARGET_NAME"] as? String }
```
yonaskolb: It's interesting that this could change across configs. I wonder how Xcode deals with this. This is probably fine though

anreitersimon: I was wondering about that too.
Xcode doesnt let you specify this on a per configuration basis.
Meaning it adds the `TEST_TARGET_NAME ` setting to all configurations.
 and then (i assume) generates the target-attributes.

Tests/XcodeGenKitTests/ProjectGeneratorTests.swift
```diff
@@ -215,10 +243,12 @@ func projectGeneratorTests() {
 
             $0.it("generates dependencies") {
                 let pbxProject = try getPbxProj(spec)
+                
                 let nativeTargets = pbxProject.objects.nativeTargets.referenceValues
                 let dependencies = pbxProject.objects.targetDependencies.referenceValues
-                try expect(dependencies.count) == 1
-                try expect(dependencies.first!.target) == nativeTargets.first { $0.name == framework.name }!.reference
+                try expect(dependencies.count) == 2
+                try expect(dependencies[0].target) == nativeTargets.first { $0.name == framework.name }!.reference
```
anreitersimon: relying on the order in which target target-dependencies are generated  might not be very resilient.
I did not know how to fix this. 
Actually "resolving" the dependencies and then asserting that they reference the correct  target might be a better aproach.

PR #225, opened 2018-01-21
Update PBXProjGenerator to embed frameworks also in test bundles and extensions

pepibumur:
Before this change, XcodeGen generated project didn't embed Carthage frameworks when the target was a testing bundle, or a extension. In those cases, Carthage dynamic frameworks should be embedded as well.

I've updated `XCProjExtensions.swift` adding a couple of getters and used them from `PBXProjGenerator.swift`.

### DIFF
--- a/Sources/ProjectSpec/XCProjExtensions.swift
+++ b/Sources/ProjectSpec/XCProjExtensions.swift
@@ -29,6 +29,14 @@ extension PBXProductType {
     public var isApp: Bool {
         return fileExtension == "app"
     }
+    
+    public var isTestBundle: Bool {
+        return fileExtension == "xctest"
+    }
+    
+    public var shouldEmbedDynamicFrameworks: Bool {
+        return isApp || isExtension || isTestBundle
+    }
 
     public var name: String {
         return rawValue.replacingOccurrences(of: "com.apple.product-type.", with: "")
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -331,7 +331,7 @@ public class PBXProjGenerator {
 
         for dependency in target.dependencies {
 
-            let embed = dependency.embed ?? (target.type.isApp ? true : false)
+            let embed = dependency.embed ?? (target.type.shouldEmbedDynamicFrameworks ? true : false)
             switch dependency.type {
             case .target:
                 let dependencyTargetName = dependency.reference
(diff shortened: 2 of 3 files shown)


Sources/ProjectSpec/XCProjExtensions.swift
```diff
@@ -29,6 +29,14 @@ extension PBXProductType {
     public var isApp: Bool {
         return fileExtension == "app"
     }
+    
+    public var isTestBundle: Bool {
+        return fileExtension == "xctest"
+    }
+    
+    public var shouldEmbedDynamicFrameworks: Bool {
+        return isApp || isExtension || isTestBundle
```
keith: This isn’t entirely correct for the case where you have both an extension (notification / Siri / etc, not watch) you’d actually only want to embed it once, and just have the runtime search paths for the extension to include it

pepibumur: You are right. My case was that I wanted to embed some Carthage frameworks into the watch extension. However, they didn't get copied because XcodeGen detected that the destination product was a extension. I wouldn't make that decision from XcodeGen and leave it up to the developer using the tool. In essence, if they say the dependency is a Carthage one, and they want to embed it, XcodeGen would just embed it, regardless of the destination product.
What do you think? cc @yonaskolb

keith: We should definitely fix the watch case, since you always want to embed it there, but since the watch is a special case, we might not want to do that for all extension types

keith: I guess the problem here more generally that we've also been trying to combat at Lyft is that XcodeGen is a very thin layer over the xcodeproj file format, and it's not particularly smart about all the complexities of this. I suppose in an ideal world it would know:

- Whether or not the framework target was depended on by multiple target types
- What the target types depending on it were

And then use that information in order to affect the build settings, like we said in this case we would want to add an entry to the runtime search paths for this.

yonaskolb: I think for now we should change this from checking file extensions to switching over the product types and returning true for application types and watch apps

yonaskolb: Be aware too that this affects more than just carthage dependencies, but also framework and target dependencies.

yonaskolb:
There is a watchOS app in the Test fixture project now that runs without issues (after carthage has been bootstrapped)

PR #227, opened 2018-01-24
Update to xcproj 4.0

### LINKED ISSUE #223: Update to xcproj 2.0 or later

yonaskolb:
Resolves #223 

This mostly involves dealing with the fact that PBXObject no longer has a `reference`


Package.swift
```diff
@@ -16,7 +16,7 @@ let package = Package(
         .package(url: "https://github.com/yonaskolb/JSONUtilities.git", from: "3.3.0"),
         .package(url: "https://github.com/kylef/Spectre.git", from: "0.8.0"),
         .package(url: "https://github.com/onevcat/Rainbow.git", from: "3.0.0"),
-        .package(url: "https://github.com/xcodeswift/xcproj.git", from: "1.7.0")
+        .package(url: "https://github.com/xcodeswift/xcproj.git", .revision("ee552d3"))
```
rahul-malik: Pin to version?

Sources/ProjectSpec/Target.swift
```diff
@@ -219,7 +218,7 @@ extension Target: NamedJSONDictionaryConvertible {
 
     public init(name: String, jsonDictionary: JSONDictionary) throws {
         self.name = jsonDictionary.json(atKeyPath: "name") ?? name
-        self.productName = jsonDictionary.json(atKeyPath: "productName")
+        productName = jsonDictionary.json(atKeyPath: "productName")
```
rahul-malik: just removing `self.`? should we do this on the line above as well (`self.name`) ?

Sources/XcodeGenKit/ProjectGenerator.swift
```diff
@@ -31,24 +31,21 @@ public class ProjectGenerator {
...
-            let targetReference =
-                pbxProject.objects.nativeTargets.referenceValues.first { predicate($0 as PBXTarget) } ??
-                pbxProject.objects.legacyTargets.referenceValues.first { predicate($0 as PBXTarget) }!
+            let targetRerence = pbxProject.objects.targets(named: buildTarget.target).first!
 
             let buildableReference = XCScheme.BuildableReference(
                 referencedContainer: "container:\(spec.name).xcodeproj",
-                blueprintIdentifier: targetReference.reference,
-                buildableName: buildTarget.target + (targetReference.productType?.fileExtension.map { ".\($0)" } ?? ""),
+                blueprintIdentifier: targetRerence.reference,
```
rahul-malik: Nit: Spelling of `Reference`

Package.swift
yonaskolb: This relies on unreleased coded. It will be updated with xcproj 4.0

Sources/ProjectSpec/Target.swift
yonaskolb: SwiftFormat did this. 
self.name is still required as name is a parameter

PR #230, opened 2018-01-25
Fix handling of scheme target build type

kastiglione:
While looking into #229, I noticed what looks to be a bug in build type handling for the `String` case.

An example project could look like:

```yaml
schemes:
  Foo:
    build:
      targets:
        SomeTarget: test
```

But the `test` value would be ignored, instead it would the build types would be set to `.all`.


yonaskolb:
No real harm in adding this, but it wasn't quite designed like this. I doubt someone would want just a single run type. Also if someone passed `testing` it would actually fall into the `testing` case above and apply multiple types. If someone really wants just testing they could do
```
myTarget: [testing] //or [test]
```

kastiglione:
Thanks for the clarification. I think I misunderstood the documentation.

PR #233, opened 2018-01-31
Make scheme `config` optional, relying on defaults

kastiglione:
Change the `config` field in scheme actions to optional. When not specified, a default will be used. This updates the spec to reflect this change.

### DIFF
--- a/Docs/ProjectSpec.md
+++ b/Docs/ProjectSpec.md
@@ -427,7 +427,7 @@ targets:
 ### Common Build Action options
 The different actions share some properties:
 
-- [ ] **config**: **String** - All build actions can be set to use a certain config. If this is not set the first configuration found of a certain type will be used, depending on the type:
+- [x] **config**: **String** - All build actions can be set to use a certain config. If this is not set the first configuration found of a certain type will be used, depending on the type:
 	- `debug`: run, test, analyze
 	- `release`: profile, archive
 - [ ] **commandLineArguments**: **[String:Bool]** - `run`, `test` and `profile` actions have a map of command line arguments to whether they are enabled
--- a/Docs/ProjectSpec.md
+++ b/Docs/ProjectSpec.md
@@ -427,7 +427,7 @@ targets:
 ### Common Build Action options
 The different actions share some properties:
 
-- [x] **config**: **String** - All build actions can be set to use a certain config. If this is not set the first configuration found of a certain type will be used, depending on the type:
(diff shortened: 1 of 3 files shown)


yonaskolb:
Yeah that's true actually. 👍  
The reason it's documented as optional in the description is that it's a required property to set on a scheme, but if something like the testAction isn't even defined it will look for a default config.
It currently says 
> All build actions can be set to use a certain config. If this is not set the first configuration found of a certain type will be used, depending on the type:


We could either make the config property optional, or change the description to something like:

> All build actions can be set to use a certain config. If a build action is not defined, the first configuration found of a certain type will be used, depending on the type:

Thoughts?

kastiglione:
I agree that it should not be mandatory to define each build action.

I think both this change should happen, and the change of wording you suggest.

I see that it is strange to have a required field within an optional field, but there is sort of precedent for this. The top level `schemes` field is optional, but the `build` field within it is marked required. My recent addition of pre- and post-actions mark the `script` field as required even though it's within optional fields.

The reason I opened this pull request is because on a couple occasions recently I have had to debug why my build actions weren't showing up in the generated Xcode project, and in those cases it was because I had left out the `config` fields.

Perhaps the solution is making this a runtime warning or error, like:

> Ignoring (build|run|test|...) action in <schemeName> scheme because it's missing a config field.

yonaskolb:
Yeah I think both those changes should be made 👍 

- make `config` optional
- make description `All build actions can be set to use a certain config. If a config, or the build action itself, is not defined the first configuration found of a certain type will be used, depending on the type:`

I agree it's strange to have a required field in a optional property. `JSONUtilities` doesn't help there either because optional properties ignore their internal errors (which is why you weren't seeing any errors in your case). The move to `Decodable` which has an explicit `decodeIfPresent` will solve that issue, as it will still throw errors if the property is present but malformed.

The other cases you mentioned are slightly different as they are either lists of things, or don't have defaults. (you need assign a script to an action). In this case config has a default (the first found config of the required type) which is a good default for people who have just debug and release (or don't even define any configs and get the defaults)

kastiglione:
> make `config` optional

so change the code so that it's optional and finds the default as you've described?

yonaskolb:
Yep. All that’s needed in the generator side should be an extra `?`

kastiglione:
I updated the code and documentation. Let me know if this looks like what you expected.

PR #238, opened 2018-02-08
sources.excludes wasn't applied for localisedDirectory such as Base.lproj

toshi0383:
I happened to have `App/Base.lproj/.DS_Store` and XcodeGen treated as a BuildFile.
Fixed to ignore `.DS_Store` and also files specified by `sources.excludes`.

ref:
- https://github.com/yonaskolb/XcodeGen/blob/master/Docs/ProjectSpec.md#sources


.gitignore
```diff
@@ -1,4 +1,7 @@
-.DS_Store
+# NOTE: Do not ignore .DS_Store for test fixtures.
+# .DS_Store
+./.DS_Store
```
yonaskolb: Does this mean `.DS_Store` won't be ignored in other parts of the repo

toshi0383: Yea it does for now.

PR #241, opened 2018-02-13
Add parallelizeBuildables and buildImplicitDependencies flags to the …

rahul-malik:
…build

scheme settings

We have a few scenarios where we'd like to control if these are enabled or disabled. I've set the default values to "true" for both properties to ensure we don't change existing behavior.

### DIFF
--- a/Sources/ProjectSpec/Scheme.swift
+++ b/Sources/ProjectSpec/Scheme.swift
@@ -51,19 +51,19 @@ public struct Scheme: Equatable {
 
     public struct Build: Equatable {
         public var targets: [BuildTarget]
-        public var parallelizeBuildables: Bool
+        public var parallelizeBuild: Bool
         public var buildImplicitDependencies: Bool
         public var preActions: [ExecutionAction]
         public var postActions: [ExecutionAction]
         public init(
             targets: [BuildTarget],
-            parallelizeBuildables: Bool = true,
+            parallelizeBuild: Bool = true,
             buildImplicitDependencies: Bool = true,
             preActions: [ExecutionAction] = [],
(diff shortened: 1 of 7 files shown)


keith:
Are you seeing big differences by changing these?

Sources/ProjectSpec/Scheme.swift
```diff
@@ -51,14 +51,20 @@ public struct Scheme: Equatable {
 
     public struct Build: Equatable {
         public var targets: [BuildTarget]
+        public var parallelizeBuildables: Bool
```
yonaskolb: Can we rename this to `parallelizeBuild` to match both Xcode UI and internal name

rahul-malik:
@keith - Not performance differences. We've been playing around with generating multiple projects in a workspace and allowing Xcode's implicit dependency management to bring everything together.

The only issue right now is integration tests which currently try to build our app and test bundles in parallel which creates a race condition at link time (.app is sometimes not there yet). To solve we're just making those targets not build in parallel.

rahul-malik:
@yonaskolb - added a few lines to the project spec docs

yonaskolb:
@rahul-malik Thanks! One more thing.
Can you add some test expectations [here](https://github.com/yonaskolb/XcodeGen/blob/master/Tests/XcodeGenKitTests/ProjectGeneratorTests.swift#L324), and also turn the settings to false in the test project fixture [here](https://github.com/yonaskolb/XcodeGen/blob/master/Tests/Fixtures/TestProject/spec.yml#L103)

rahul-malik:
@yonaskolb - Thanks for pointing out those test updates. These should be addressed now

PR #244, opened 2018-02-24
 Add carthageExecutablePath option

akkyie:
I added an option called `carthageExecutablePath` to where Carthage binary is installed. You can change it from `/usr/local/bin/carthage` to use non-global installation of Carthage when, for example, you want to install it locally with [Mint](https://github.com/yonaskolb/Mint).

Let me know if I missed something to add or change. Thanks!


akkyie:
Wow, thanks for quick response 🎉 

I just followed the status quo, but defaulting to just `carthage` sounds natural 👍

PR #245, opened 2018-02-24
Allow setting buildPhase on folders

### LINKED ISSUE #242: Framework public headers folder structure
Hello,

first of all thanks for this great tool, digging it a lot so far! 

I need to have a defined folder structure in the Public Headers section of my framework but so far I am struggling to obtain this through XcodeGen.

Basically if I declare a folder like:
      - path: MyProject/ProtectedInterfaces/Foo
        buildPhase: headers
        type: folder

I would expect to have the folder Foo copied in the Headers folder of my framework. But what I get actually copied is each single files contained in Foo so and the structure is not respected (and my client cannot properly include the framework).

Do you have any idea on how I can accomplish this?

Thanks!

yonaskolb:
Before this PR the `buildPhase` for a `folder` was assumed to be `resources`. This allows it to be overridden with the `buildPhase` property on `TargetSource`, allowing for folders with headers.

Possibly resolves #242

### DIFF
--- a/Sources/XcodeGenKit/SourceGenerator.swift
+++ b/Sources/XcodeGenKit/SourceGenerator.swift
@@ -382,7 +382,7 @@ class SourceGenerator {
                 rootGroups.insert(fileReference)
             }
 
-            let sourceFile = generateSourceFile(targetSource: targetSource, path: folderPath, buildPhase: .resources)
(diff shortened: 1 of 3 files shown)


libe:
Thanks, that was just what I was looking for!

Issue #234, opened 2018-02-01
Include standalone XcodeGen binaries in releases


kastiglione:
Has there been any interest in making a binary available with each release? The way we manage versioning and updates of XcodeGen is by creating a release in our fork that includes a binary. If others want this too, maybe it's worth adding binaries to the official releases.

yonaskolb:
No one has asked about yet. One thing is that the `settingPresets` also need to live in the same directory, which mint and homebrew setup correctly.

It's possible to bundle everything up into a zip. I'd want to do this automatically though via the makefile. If you'd like to write a script to talk to github that's great!

rahul-malik:
We're using XCodeGen as a framework so this doesn't benefit us but just wanted to support this since we're thinking of having precompiled binaries for all our dependencies so it's easy to upgrade team members by avoiding compilation on everyones machine.

toshi0383:
I guess this would help resolving the issue?
https://github.com/yonaskolb/Mint/pull/52

We can put this in Makefile as part of release process,
```
mint archive xcodegen
```
then in your project you can install like this.
```
PREFIX=mint/XcodeGen mint install yonaskolb/XcodeGen --binary
```
Giving the `$PREFIX` results the binary installed at `$PWD/mint/XcodeGen/bin/xcodegen`.

toshi0383:
> I'd want to do this automatically though via the makefile. If you'd like to write a script to talk to github that's great!

@yonaskolb 
In my experience, [GitHub release API is not straight forward to use](https://github.com/fastlane/fastlane/issues/575#issuecomment-139307075), so I would just upload the generated zip file manually to the release page.
It was a while ago, so you can of-course try.😉

keith:
I think ideally we would include the settings presents files as strings in the code, so a standalone binary would be feasible. We’re actually using a standalone binary we build currently and just exclude those files since we don’t use them.

toshi0383:
@keith 
> we would include the settings presents files as strings in the code

In which code you talking about?👀

If I understood you correctly, would that be `.mint.yml` or something like that?
https://github.com/yonaskolb/Mint/issues/2

keith:
No I just mean that ideally XcodeGen could be entirely standalone without any supporting yams files. Those could be strings in the code

toshi0383:
Ah I see. So ideally we don’t need project.yml, xcconfigs, and all that kind of configuration files.🤔

yonaskolb:
[...]

PR #246, opened 2018-02-25
Binary archive install

toshi0383:
Resolves https://github.com/yonaskolb/XcodeGen/issues/234 .

Installing is done in **3.81 sec** on CI. ✅

# How to install binary?
```
# Manually download the zip
# e.g. like this
curl -L https://github.com/toshi0383/XcodeGen/releases/download/1.6.1/xcodegen.zip -o a.zip

# optionally compare hash value

unzip a.zip
./XcodeGen/install.sh
```

If you want to change the `$PREFIX`,
```
PREFIX=/usr ./XcodeGen/install.sh
```

# archive.sh

Use this to archive a zip file.
`install.sh` is embedded inside the zip.

# install-binary.sh

This script performs binary install locally from `xcodegen.zip` at current directory.
This is modified by `archive.sh` and embedded into the zipfile.

# contents of zipfile
```
XcodeGen/
├── bin
│   └── xcodegen
├── install.sh
└── share
    └── SettingPresets
        ├── Configs
        │   ├── debug.yml
        │   └── release.yml
        ├── Platforms
        │   ├── iOS.yml
        │   ├── macOS.yml
        │   ├── tvOS.yml
        │   └── watchOS.yml
        ├── Product_Platform
        │   ├── application_iOS.yml
        │   ├── application_macOS.yml
        │   ├── application_tvOS.yml
        │   └── application_watchOS.yml
        ├── Products
        │   ├── application.watchapp2.yml
        │   ├── application.yml
        │   ├── bundle.ui-testing.yml
        │   ├── bundle.unit-test.yml
        │   ├── framework.yml
        │   ├── tv-app-extension.yml
        │   └── watchkit2-extension.yml
        └── base.yml

7 directories, 20 files
```

# Release Steps
Maintainers need to..
- `make archive` to create the zip
- upload the zip to GitHub release page

# Try out
I uploaded the zip to my fork repo.
https://github.com/toshi0383/XcodeGen/releases/tag/1.6.1


Makefile
```diff
@@ -41,3 +41,9 @@ release: format_code
 	git add .
 	git commit -m "Update to $(VERSION)"
 	git tag $(VERSION)
+
+install-binary:
+	./scripts/install.sh
```
yonaskolb: Would this need to be called from the makefile?

scripts/archive.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+TOOL_NAME=XcodeGen
```
yonaskolb: Can we move this script into the make file. Many of these variables are repeated

scripts/install-binary.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+PREFIX=${PREFIX:-/usr/local}
+
+### NOTE: Stripped or replaced by archive.sh from this line
```
yonaskolb: We shouldn't need these markers

scripts/install-binary.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+PREFIX=${PREFIX:-/usr/local}
```
yonaskolb: Instead of an ENV can we just make this an optional argument
```
sh install-binary.sh /usr/local
```

Makefile
toshi0383: Yea it was just for debug purpose, but the name is outdated. Will remove it👌

scripts/install-binary.sh
toshi0383: Well, it's okay to me, but I thought it's better to mention about `archive.sh` hack. New contributors wouldn't know how this lines are treated.
I thought about separating `install.sh` for release and `install-binary.sh` for debug purpose, but I guess this is easier to maintain.
I can separate them if you prefer that way.😉

scripts/install-binary.sh
toshi0383: Shell arguments are always mystery for new user. `PREFIX=/usr sh install.sh` is more explicit than `sh install.sh /usr`. I prefer ENV style. Also in future we might add new arguments for  something else.
So for now, how about this?
```
PREFIX=${PREFIX:-${1:-/usr/local}}
```

yonaskolb: That's fine 👍 
I suspect most people won't specify the path anyway.
Can we also rename the script simply `install.sh`? Shorter to type and it's installing more than just a binary

scripts/install-binary.sh
yonaskolb: One script is fine. Can we at least strip those lines from the script that ends up in the zip file then?

scripts/archive.sh
```diff
@@ -0,0 +1,25 @@
...
+TOOL_NAME_LOWER=xcodegen
+TMP=$(mktemp -d)/$TOOL_NAME
+BINDIR=$TMP/bin
+SHAREDIR=$TMP/share
+ZIPFILE=$TMP/$TOOL_NAME_LOWER.zip
+INSTALLSH=scripts/install-binary.sh
+
+mkdir -p $BINDIR
```
yonaskolb: Can we also copy the LICENSE.md file

scripts/install-binary.sh
toshi0383: Yep, I already rename it as `install.sh` in `archive.sh`, but will simply rename it, removing unzipping part. (as we discussed in slack) 👍

scripts/archive.sh
toshi0383: Added `export` in Makefile to share variables. d07f160
I have to fix code format for Makefile, so I keep the logic in separated script for now.

Makefile
toshi0383: done in a808975

scripts/install-binary.sh
toshi0383: Fixed in c0aabd0

scripts/archive.sh
toshi0383: Done in f540285

toshi0383:
Improved the directory name shown to user when unzipped. be0e537
So the contents would be:
```
XcodeGen-1.6.0/
├── LICENSE
├── bin
│   └── xcodegen
├── install.sh
└── share
    └── SettingPresets
        ├── Configs
        │   ├── debug.yml
        │   └── release.yml
        ├── Platforms
        │   ├── iOS.yml
        │   ├── macOS.yml
        │   ├── tvOS.yml
        │   └── watchOS.yml
        ├── Product_Platform
        │   ├── application_iOS.yml
        │   ├── application_macOS.yml
        │   ├── application_tvOS.yml
        │   └── application_watchOS.yml
        ├── Products
        │   ├── application.watchapp2.yml
        │   ├── application.yml
        │   ├── bundle.ui-testing.yml
        │   ├── bundle.unit-test.yml
        │   ├── framework.yml
        │   ├── tv-app-extension.yml
        │   └── watchkit2-extension.yml
        └── base.yml

7 directories, 21 files
```

scripts/archive.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+PACKAGE_NAME=${TOOL_NAME:?}-${VERSION:?}
```
yonaskolb: I'm not sure I've ever seen release artefacts with versions on them on github, as they are already uploaded to a specific version. Do you have any examples?
It would require anyone downloading them from a script to dynamically handle the version in the filename (I know they would already be doing it for the url)

toshi0383: You're right, but once you have downloaded from release page, you can't tell which one is which version. It's better for me to have version number in the extracted directory name.

screen shot 2018-02-27 at 8 26 13

By the way I didn't change the zipfile name. `xcodegen.zip`

It's easy to handle the extracted directory name to execute `install.sh`. Like this.
```
curl https://latest.release.zip -o tmp/a.zip
(cd tmp; unzip a.zip; ./**/install.sh)
```

> (I know they would already be doing it for the url)

They wouldn't have to, you know? If they use GitHub API to get the latest release.

yonaskolb: I see, it’s not the zip filename, it’s the unzipped directory name. I would say that’s even worse in my opinion, as CI script authors would assume the directory would have the same name as the zip file. Forcing them to use a glob pattern or specify a version path isn’t great. (Especially if they use a GitHub api as you say from a high level and then don’t think about the version afterwards)
I think your use case of wanting to know which file you unzipped is an edge case and comes about just because you have so many duplicates from testing.
Thoughts?

toshi0383: Indeed, you're right. I'll change the unzipped dir name to `xcodegen` as well 👌

toshi0383: Done in 6cd4367

PR #250, opened 2018-02-26
Write project file atomically

### LINKED ISSUE #235: Atomic writing?
Is it possible to make XcodeGen support atomic writes? I noticed that if the program is killed, it can end up in a weird state ( i.e. schemes may not be written but the project was ).

I'd like to write the project to a temporary file path, and then copy it over once writes are fully complete. I noticed this is non trivial due to validation in XcodeGen, which seems to check paths relative to the output directory.

Here is how I'm working around this:
- Give XcodeGen a path adjacent to the project I want to write
- Let that project be written
- On success, move that to the actual path

yonaskolb:
Resolves #235 

This:
- copies the project file to a temporary directory (if it exists)
- writes into that temporary project
- deletes the original
- copies that project back to the original

### DIFF
--- a/Sources/XcodeGen/main.swift
+++ b/Sources/XcodeGen/main.swift
@@ -44,8 +44,18 @@ func generate(spec: String, project: String, isQuiet: Bool, justVersion: Bool) {
         let project = try projectGenerator.generateProject()
 
         logger.info("⚙️  Writing project...")
+        
         let projectFile = projectPath + "\(spec.name).xcodeproj"
-        try project.write(path: projectFile, override: true)
+        let tempPath = Path.temporary + "XcodeGen_\(Int(NSTimeIntervalSince1970))"
+        try? tempPath.delete()
+        if projectFile.exists {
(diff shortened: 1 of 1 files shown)


Sources/XcodeGen/main.swift
```diff
@@ -44,8 +44,18 @@ func generate(spec: String, project: String, isQuiet: Bool, justVersion: Bool) {
...
-        try project.write(path: projectFile, override: true)
+        let tempPath = Path.temporary + "XcodeGen_\(Int(NSTimeIntervalSince1970))"
+        try? tempPath.delete()
+        if projectFile.exists {
+            try projectFile.copy(tempPath)
+        }
+        try project.write(path: tempPath, override: true)
+        try? projectFile.delete()
+        try tempPath.copy(projectFile)
+        try? tempPath.delete()
```
toshi0383: We should trap signals to ensure deletion, but it's under tmp dir anyways, so I guess we can leave it as future TODO.

PR #252, opened 2018-02-26
Add TargetSource headerVisibility

yonaskolb:
Alternative to #187

This adds an optional `headerVisibility` property to `TargetSource`.

It also changes the default visibility of headers from Public to Project. This matches the default that Xcode has for header files. (Interestingly Xcode also doesn't explicitly write the visibility if it set to  Project) 
This change isn't required for this PR, and is potentially breaking. Does anyone have any thoughts on changing the default?

### DIFF
--- a/Docs/ProjectSpec.md
+++ b/Docs/ProjectSpec.md
@@ -238,6 +238,10 @@ A source can be provided via a string (the path) or an object of the form:
 	- `file`: a file reference with a parent group will be created (Default for files or directories with extensions)
 	- `group`: a group with all it's containing files. (Default for directories without extensions)
 	- `folder`: a folder reference.
+- [ ] **headerVisibility**: **String** The visibility of any headers. This defaults to `project`, but can be either:
+	- `public`
+	- `private`
+	- `project`
 
 ```yaml
 targets:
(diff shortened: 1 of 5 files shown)


Docs/ProjectSpec.md
```diff
@@ -238,6 +238,10 @@ A source can be provided via a string (the path) or an object of the form:
 	- `file`: a file reference with a parent group will be created (Default for files or directories with extensions)
 	- `group`: a group with all it's containing files. (Default for directories without extensions)
 	- `folder`: a folder reference.
+- [ ] **headerVisibility**: **String** The visibility of any headers. This defaults to `project`, but can be either:
+	- `public`
+	- `private`
+	- `project`
```
toshi0383: Maybe we can allow upper-case at first letter. `Public` `Private` `Project`

yonaskolb: Yeah I guess the decoding could lowercase it first. I'd want the documented way to be lowercase though to match the rest of the spec

toshi0383:
> it worked perfectly fine for my app

Oh maybe I just had dirty workspace. For embedded frameworks, maybe it should be `public` by default? 🤔

screen shot 2018-02-27 at 17 04 39

yonaskolb:
Yeah keeping public is the easiest thing to do, that won't break existing builds, and makes it so you probably don't have to do anything. You would be leaking some internal headers though

toshi0383:
Xcode makes the umbrella header public when you add "Cocoa Touch Framework" target, so I guess we can go either way.
I would keep current behavior, but I'd like to hear what others think.

yonaskolb:
Changed default back to `public`

iamtomcat:
The default isn't reflected in the docs.

yonaskolb:
Good pickup @iamtomcat. Fixed in master 👍

iamtomcat:
@yonaskolb No problem.

PR #254, opened 2018-02-27
Fix shouldUseLaunchSchemeArgsEnv disabling logic

toshi0383:
Adding `environmentVariables` was not disabling the `shouldUseLaunchSchemeArgsEnv` flag.
Therefore the process could not read the variables.

### DIFF
--- a/Sources/ProjectSpec/Scheme.swift
+++ b/Sources/ProjectSpec/Scheme.swift
@@ -140,6 +140,10 @@ public struct Scheme: Equatable {
                 lhs.postActions == rhs.postActions &&
                 lhs.environmentVariables == rhs.environmentVariables
         }
+
+        public var shouldUseLaunchSchemeArgsEnv: Bool {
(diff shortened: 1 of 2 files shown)


turekj:
@yonaskolb the environment variables won't ever be empty in that particular context, because they are re-mapped to nil when empty:

```
let testVariables = scheme.test.flatMap { $0.environmentVariables.isEmpty ? nil : $0.environmentVariables }
```

The command line arguments behave the other way I believe, it's a dict that is initialized to empty one by default and `profileCommandLineArgs` can only be nil if there is no profile scheme. 

I think the cleanest solution would be to add an extension on a `Profile` struct and `Test` struct that computes that:

```
extension Profile {
    var shouldUseLaunchSchemeArgsEnv: Bool {
        return commandLineArguments.isEmpty && environmentVariables.isEmpty
    }
}
```

What do you think @toshi0383?

toshi0383:
@yonaskolb  @turekj Good point. Fixed in 0a37fd5 👌

PR #255, opened 2018-03-01
Fix iMessage extension

yonaskolb:
Setting Presets
- sets the correct app icon
- sets runtime search paths to `$(inherited) @executable_path/Frameworks @executable_path/../../Frameworks`, allow it to lookup frameworks in the application

Fixes https://github.com/pepibumur/xcode-modular-apps-workshop


SettingPresets/Products/app-extension.messages.yml
```diff
@@ -0,0 +1,2 @@
+ASSETCATALOG_COMPILER_APPICON_NAME: iMessage App Icon
+LD_RUNPATH_SEARCH_PATHS: "$(inherited) @executable_path/Frameworks @executable_path/../../Frameworks"
```
toshi0383: > @executable_path/../../Frameworks

Wow, this is new👀

yonaskolb: What’s new?

yonaskolb:
@keith

SettingPresets/Products/app-extension.messages.yml
toshi0383: I mean the location path was new to me.  NVM👋

PR #256, opened 2018-03-01
Update carthage embedding

yonaskolb:
- Embed carthage dependencies into tests
- Listen to explicit `embed` property for Carthage as well

### DIFF
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -364,7 +364,7 @@ public class PBXProjGenerator {
 
         for dependency in target.dependencies {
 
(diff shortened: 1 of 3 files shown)


Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -364,7 +364,7 @@ public class PBXProjGenerator {
 
         for dependency in target.dependencies {
 
-            let embed = dependency.embed ?? (target.type.isApp ? true : false)
+            let embed = dependency.embed ?? target.shouldEmbedDependencies
```
toshi0383: Oh maybe I embedded the bug here. Thanks👍
I was about to add Quick.framework via Carthage to my app's test target.

PR #257, opened 2018-03-01
Only parse true and false as bools

### LINKED ISSUE #253: Spec decoder's type inference issue
Currently this `environmentVariables ` part fails to parse quietly and XcodeGen is not writing it to xcscheme. I debugged a bit, and the value `YES` is mistakenly treated as bool.
```yml
  BitriseClientTests:
    platform: iOS
    type: bundle.unit-test
    configFiles:
      Debug: configs/BitriseClientTests-Debug.xcconfig
    sources:
      - Tests/BitriseClientTests
    dependencies:
      - target: BitriseClient
    scheme:
      testTargets:
        - BitriseClientTests
      gatherCoverageData: true
      environmentVariables:
        - variable: TEST
          value: YES
          isEnabled: true
```

The document demands `environmentVariables` to be `[String: String]`, so the decoder should accept `YES` or `1` as string, too.

yonaskolb:
Fixes #253 

This uses the new customizable `Resolver` in Yams, to disable inferring strings like `YES` to bools from yaml files.

```diff
- ^(?:yes|Yes|YES|no|No|NO\|true|True|TRUE|false|False|FALSE\|on|On|ON|off|Off|OFF)$
+ ^(?:true|false)$
```

This is technically a breaking change if someone was using these strings to define bools. I can't see any other way to fix #253 though.
Thoughts?

### DIFF
--- a/Sources/ProjectSpec/Yaml.swift
+++ b/Sources/ProjectSpec/Yaml.swift
@@ -7,26 +7,17 @@ public func loadYamlDictionary(path: Path) throws -> [String: Any] {
     if string == "" {
         return [:]
     }
-    guard let yaml = try Yams.load(yaml: string) else {
+
+    // just treat true and false as Bools
+    let boolRule = try! Resolver.Rule(.bool, "^(?:true|false)$")
+
+    let resolver = Resolver.default
+        .removing(.null) // remove rule so that empty quotes are treated as empty strings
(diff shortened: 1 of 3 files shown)


yonaskolb:
Would maybe only take out `YES` and `NO`, as they are in a lot of Xcode variables. But if it’s handled so specifically, we can just convert Bools to `YES` and `NO` in the environment variable decoder.

PR #259, opened 2018-03-01
Fix env parsing

### LINKED ISSUE #253: Spec decoder's type inference issue

yonaskolb:
Fixes #253 
Alternative to #257 

Turns `YES` and `NO` that get parsed to bools in yaml, back into `YES` and `NO` strings.
Also turns any other type into a string (eg Int)

### DIFF
--- a/Sources/ProjectSpec/Scheme.swift
+++ b/Sources/ProjectSpec/Scheme.swift
@@ -366,15 +366,31 @@ extension BuildType: JSONPrimitiveConvertible {
 
 extension XCScheme.EnvironmentVariable: JSONObjectConvertible, Equatable {
 
+    static private func parseValue(_ value: Any ) -> String {
+        if let bool = value as? Bool {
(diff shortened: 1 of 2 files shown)


yonaskolb:
@turekj

PR #263, opened 2018-03-08
Rename ProjectSpec -> ProjectSpec.Project

jerrymarino:
I've started using the XcodeGen data structures in a large project, and
its been awesome so far!

The problem I'm running into is I can't import types from the
`ProjectSpec` module. This caused by a naming collision between the struct
`ProjectSpec` and the module `ProjectSpec`.

To remedy the issue, this patch renames the struct to `Project`. This is
an overall improvement, since all of the types are now nicely namespaced
and can be imported as expected.

i.e.
```
import ProjectSpec

let projectSpec: ProjectSpec.Project
let targetSpec: ProjectSpec.Target
```

This makes it easier to differentiate in the code when we are dealing
with types of similar names, like `PBXProject`.

Any feedback is welcome 👍

### DIFF
--- a/Sources/XcodeGenKit/SettingsBuilder.swift
+++ b/Sources/XcodeGenKit/SettingsBuilder.swift
@@ -112,7 +112,7 @@ extension Project {
 
         if let configSettings = loadConfigFileBuildSettings(path: configPath) {
             for key in configSettings.keys {
-                // FIXME: Catch platform projectifier. e.g. LD_RUNPATH_SEARCH_PATHS[sdk=iphone*]
+                // FIXME: Catch platform specifier. e.g. LD_RUNPATH_SEARCH_PATHS[sdk=iphone*]
                 buildSettings.removeValue(forKey: key)
                 buildSettings.removeValue(forKey: key.quoted)
             }
--- a/Sources/XcodeGenKit/SettingsBuilder.swift
+++ b/Sources/XcodeGenKit/SettingsBuilder.swift
@@ -5,7 +5,7 @@ import ProjectSpec
 import xcproj
 import Yams
 
-extension ProjectSpec {
+extension Project {
 
     public func getProjectBuildSettings(config: Config) -> BuildSettings {
         var buildSettings: BuildSettings = [:]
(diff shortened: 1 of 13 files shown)


Sources/XcodeGenKit/SettingsBuilder.swift
```diff
@@ -112,7 +112,7 @@ extension ProjectSpec {
 
         if let configSettings = loadConfigFileBuildSettings(path: configPath) {
             for key in configSettings.keys {
-                // FIXME: Catch platform specifier. e.g. LD_RUNPATH_SEARCH_PATHS[sdk=iphone*]
+                // FIXME: Catch platform projectifier. e.g. LD_RUNPATH_SEARCH_PATHS[sdk=iphone*]
```
yonaskolb: Search and replace typo

Tests/XcodeGenKitTests/ProjectGeneratorTests.swift
```diff
@@ -110,7 +110,7 @@ func projectGeneratorTests() {
             }
 
             $0.it("generates configs") {
-                let spec = ProjectSpec(
+                let spec = ProjectSpec.Project(
                     basePath: "",
```
yonaskolb: It seems everything in Tests has a different renaming applied. Does leaving off the `ProjectSpec` module cause naming conflicts?

jerrymarino: Hmm - we shouldn't have a naming conflict, will update to match the rest of the renames!

Sources/XcodeGenKit/SettingsBuilder.swift
jerrymarino: Good catch!

jerrymarino:
@yonaskolb thanks again for the review, and sorry for the latency on this. I've rebased this on master and updated the test renaming to match the rest 👍

yonaskolb:
Thanks @jerrymarino! I've added your changes into here https://github.com/yonaskolb/XcodeGen/pull/281

PR #272, opened 2018-03-26
Add LD_RUNPATH_SEARCH_PATHS to app-extension preset

LinusU:
This was added to the iMessage extensions, but it seems that it's required for an action extension to work as well.


LinusU:
Wow, quick pull 😍

PR #274, opened 2018-03-26
Fix some typos & presentation in ProjectSpec doc

AliSoftware:
Especially:

* `targets` root key is supposed to be a dictionary but is documented as root level to just be a `Target`
* Some mis-indentation in one of the examples
* Some markdown renderers don't like if you don't have an empty line after a header, so I added them to be sure to render it properly in all markdown softwares.


yonaskolb:
Thanks @AliSoftware 👍

PR #280, opened 2018-03-29
Target scheme pre and post actions

### LINKED ISSUE #279: preActions for Target Schemes (suggestion)
Currently it seems it is not possible to define `preActions` for target schemes.
Consider the following target scheme:

```
    scheme:  
      configVariants:
        - Dev
        - AdHoc
        - AppStore
```
it would be nice to have the ability to specify an (optional) common `preActions` element that is embedded for each generated scheme.

yonaskolb:
Resolve #279

### DIFF
--- a/Sources/ProjectSpec/Scheme.swift
+++ b/Sources/ProjectSpec/Scheme.swift
@@ -36,7 +36,7 @@ public struct Scheme: Equatable {
         public var script: String
         public var name: String
         public var settingsTarget: String?
-        public init(name: String, script: String, settingsTarget: String?) {
+        public init(name: String, script: String, settingsTarget: String? = nil) {
             self.script = script
             self.name = name
             self.settingsTarget = settingsTarget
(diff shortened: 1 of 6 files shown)


Sources/ProjectSpec/Scheme.swift
```diff
@@ -253,8 +253,8 @@ extension Scheme.Run: JSONObjectConvertible {
     public init(jsonDictionary: JSONDictionary) throws {
         config = jsonDictionary.json(atKeyPath: "config")
         commandLineArguments = jsonDictionary.json(atKeyPath: "commandLineArguments") ?? [:]
-        preActions = try jsonDictionary.json(atKeyPath: "preActions")?.map(Scheme.ExecutionAction.init) ?? []
-        postActions = try jsonDictionary.json(atKeyPath: "postActions")?.map(Scheme.ExecutionAction.init) ?? []
+        preActions = jsonDictionary.json(atKeyPath: "preActions") ?? []
+        postActions = jsonDictionary.json(atKeyPath: "postActions") ?? []
```
kastiglione: 👍 Did something change? I'm trying to recall why I would have added the `map`s if they're not necessary.

yonaskolb: Nothing changed. I just noticed these, and removed the unnecessary map. If you can think of a reason why they were there, let me know! 😀

PR #288, opened 2018-04-05
Fixed build phase of `none` not being applied to folders

### LINKED ISSUE #287: BuildPhase "none" not possible for folders
Adding a folder without copying it (so having no Target Membership) does not work:

```
- path: Carthage/Checkouts
   buildPhase: none
   type: folder
```

setting buildPhase to headers seems to have the effect, but it seems wrong 😉

yonaskolb:
Fixes #287

### DIFF
--- a/Sources/XcodeGenKit/SourceGenerator.swift
+++ b/Sources/XcodeGenKit/SourceGenerator.swift
@@ -422,8 +422,15 @@ class SourceGenerator {
             if !createIntermediateGroups {
                 rootGroups.insert(fileReference)
             }
+            
+            let buildPhase: BuildPhase?
+            if let targetBuildPhase = targetSource.buildPhase {
+                buildPhase = targetBuildPhase.buildPhase
+            } else {
+                buildPhase = .resources
+            }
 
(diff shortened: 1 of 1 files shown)


Lutzifer:
works for me when running

mint run git@github.com:yonaskolb/XcodeGen.git@no_build_phase_folders

yonaskolb:
Cool

And great example of another good use for Mint :smile: 
(Curious why you don't go with the shorter `mint run yonaskolb/XcodeGen@no_build_phase_folders`?)

Lutzifer:
> why you don't go with the shorter  `mint run yonaskolb/XcodeGen@no_build_phase_folders`

because I typed the name myself at the first try, wrote xcodegen instead of XcodeGen and got a git password prompt. then just copied the ssh address and went with that.

Long story short:
`mint run yonaskolb/XcodeGen@no_build_phase_folders` works
`mint run yonaskolb/xcodegen@no_build_phase_folders` doesn't

yonaskolb:
That's weird, works for me (though not that exact command right now as the branch doesn't exist anymore)
So this doesn't work for you?
```
mint run yonaskolb/xcodegen@master
```
How is your git case sensitive? You're not on Linux are you?

Lutzifer:
ok, ignore what I said, there was an additional k in your name 😉
should have run the above commands once again...

PR #291, opened 2018-04-10
Fix folder in root with createIntermediateGroups

### LINKED ISSUE #290: Adding Folder Reference with only subfolders fails
Structure:

```
BuildScripts/
   .../Assets/
      ... Files
   .../Common/
      ... Files
   .../External/
      ... Files
   .../Local/
      ... Files
```

Cannot be added using

```
- path: BuildScripts
        buildPhase: none
        type: folder
```

but needs to be added using

```
     - path: BuildScripts/Assets
        buildPhase: none
        type: folder
      - path: BuildScripts/Common
        buildPhase: none
        type: folder
      - path: BuildScripts/External
        buildPhase: none
        type: folder
      - path: BuildScripts/Local
        buildPhase: none
        type: folder
      - path: BuildScripts/Phases
        buildPhase: none
        type: folder
```

yonaskolb:
Fixes #290

### DIFF
--- a/Sources/XcodeGenKit/SourceGenerator.swift
+++ b/Sources/XcodeGenKit/SourceGenerator.swift
@@ -419,7 +419,7 @@ class SourceGenerator {
                 lastKnownFileType: "folder"
             )
 
-            if !createIntermediateGroups {
+            if !createIntermediateGroups || path.parent() == spec.basePath  {
                 rootGroups.insert(fileReference)
             }
             


Lutzifer:
thank you

Lutzifer:
works for me.

Another (possibly) related issue:

```
      - path: Mintfile
        buildPhase: none
        type: file
```

results in

screen shot 2018-04-10 at 18 26 15

instead of being in the Root (which I would somehow have expected.

yonaskolb:
Do you have `createIntermediateGroups` on? That's the behaviour that adds

Lutzifer:
It is on, but I don't think it should then start at the project root, not at $HOME :)

yonaskolb:
Are you saying the directory where your spec is in `mintTemp`? Hmm.
Are you running XcodeGen with a `--spec` path from somewhere else, or just in that `mintTemp` directory?

yonaskolb:
I've fixed that here #293. 
Also, it's strange to add files to a target with a build phase of none, you should add them to the project. In that PR I've also added support for adding files to `fileGroups`, not just directories 👍

Lutzifer:
> In that PR I've also added support for adding files to fileGroups, not just directories 👍

Great! Almost finished migrating our main project to Xcodegen, so I guess the number of strange situations I discover will go down a bit 😉

yonaskolb:
No worries, keep logging them 😄

PR #293, opened 2018-04-11
Support adding root files

yonaskolb:
This adds 2 things:
- fixes files that are added as root sources from having invalid parent groups outside the project directory
- add support for files in `fileGroups`. The name is a bit strange now though, perhaps we should rename it to `files`


yonaskolb:
@Lutzifer can you confirm that this fixes your issue?

Lutzifer:
> fixes files that are added as root sources from having invalid parent groups outside the project directory

this works for intermediateGroups

> add support for files in fileGroups. The name is a bit strange now though, perhaps we should rename it to files

This has the same issue with the whole parent path as subfolders

Lutzifer:
> add support for files in fileGroups. The name is a bit strange now though, perhaps we should rename it to files

Also works, I just initially forgot to restart Xcode

yonaskolb:
Ah great! 😄

PR #296, opened 2018-04-13
Remove file/product reference for legacy targets

sascha:
Apparently legacy targets created by Xcode do not have a file or product reference. Xcodeproj is unable to parse legacy targets generated with XcodeGen with the following error message:

`Xcodeproj doesn't know about the following attributes {"productReference"=>"FR_184465694726"} for the 'PBXLegacyTarget' isa.`

This commit removes file and product references for legacy targets.

### DIFF
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -129,22 +129,24 @@ public class PBXProjGenerator {
                 lastKnownFileType = fileType
             }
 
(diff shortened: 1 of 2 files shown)


yonaskolb:
@bkase you added the LegacyTarget support. What do you think of this?

yonaskolb:
I can't say I have any experience with legacy targets, and it seems you can't create them in Xcode anymore. But I trust you that they don't have build files or file references, so I'm merging this.
@bkase do you have any comments on this?

PR #298, opened 2018-04-19
Improve Carthage dependency lookup performance

keith:
Previously for large projects, project generation time could be up to 40 seconds in our testing because of lookup up Carthage dependencies (regardless of if you have any). This change improves the performance of this with 2 changes.

1. Store a dictionary of target name -> target. This allows much faster lookups than the previous O(n) array lookup (this should also improve performance of other callers of `getTarget`
2. Remove the recursively lookups of carthage dependencies. Previously this traversed your entire dependency tree for each target. Now we iteratively discover the dependencies, while still only visiting them once.

### DIFF
--- a/Sources/ProjectSpec/Project.swift
+++ b/Sources/ProjectSpec/Project.swift
@@ -8,7 +8,11 @@ public struct Project {
 
     public var basePath: Path
     public var name: String
-    public var targets: [Target]
+    public var targets: [Target] {
+        didSet {
+            self.targetsMap = Dictionary(uniqueKeysWithValues: self.targets.map { ($0.name, $0) })
+        }
+    }
     public var settings: Settings
     public var settingGroups: [String: Settings]
     public var configs: [Config]
--- a/Sources/ProjectSpec/Project.swift
+++ b/Sources/ProjectSpec/Project.swift
@@ -18,6 +18,7 @@ public struct Project {
     public var fileGroups: [String]
     public var configFiles: [String: String]
     public var include: [String] = []
+    private var targetsMap: [String: Target]
 
     public init(
         basePath: Path,
@@ -35,6 +36,7 @@ public struct Project {
         self.basePath = basePath
         self.name = name
         self.targets = targets
+        self.targetsMap = Dictionary(uniqueKeysWithValues: self.targets.map { ($0.name, $0) })
         self.configs = configs
         self.settings = settings
         self.settingGroups = settingGroups
@@ -46,7 +48,7 @@ public struct Project {
     }
 
     public func getTarget(_ targetName: String) -> Target? {
-        return targets.first { $0.name == targetName }
+        return targetsMap[targetName]
     }
 
     public func getConfig(_ configName: String) -> Config? {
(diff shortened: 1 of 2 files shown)


keith:
This produces the same project for us and passes the tests, but since we don't have any carthage dependencies it would be useful if someone else could verify it still produces a valid project for them.

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -681,29 +681,34 @@ public class PBXProjGenerator {
...
+        while !queue.isEmpty {
+            let target = queue.removeFirst()
+            if visitedTargets.contains(target.name) {
+                continue
+            }
+
+            for dependency in target.dependencies {
+                switch dependency.type {
+                case .carthage:
+                    frameworks.append(dependency)
```
kastiglione: Can this result in `frameworks` having duplicate dependencies in this list, and if so should this function return a unique list of dependencies?

yonaskolb: The carthage dependencies are de-duplicated later on when they are added to a target. This could be done here, but maybe it might be useful at some point to know how many times a dependency is being used

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -681,29 +681,34 @@ public class PBXProjGenerator {
...
-                frameworks.append(dependency)
-            case .target:
-                let targetName = dependency.reference
-                if visitedTargets[targetName] == true {
-                    return []
-                }
-                if let target = project.getTarget(targetName) {
-                    frameworks += getAllCarthageDependencies(target: target, visitedTargets: visitedTargets)
+        var queue: [Target] = [target]
+        while !queue.isEmpty {
```
yonaskolb: Is this traversing the same graph, but just doing it in one function instead of recursively?

I think to get this really performant the carthage dependencies for a target could be cached by just running this once for all targets at the beginning, instead of each target going through the whole graph again

Sources/ProjectSpec/Project.swift
```diff
@@ -18,6 +18,7 @@ public struct Project {
     public var fileGroups: [String]
     public var configFiles: [String: String]
     public var include: [String] = []
+    private var targetsMap: [String: Target]
```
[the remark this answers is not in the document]
keith: done, looks like this is only used for tests, seems like we should remove that mutability.

Sources/XcodeGenKit/PBXProjGenerator.swift
keith: It is traversing the same dependencies yes. It should have the exact same result.

keith: We could do that, I figured that would be a larger enough change that if we could do this fast enough it's not really worth it.

kastiglione: @yonaskolb The recursive version is passing `visitedTargets` by value, which I suspect is why the the iterative version is much faster.

Sources/ProjectSpec/Project.swift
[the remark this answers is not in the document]
yonaskolb: The `ProjectSpec` module can also be used as a framework in other projects.

Personally I've never seen the point of making properties on a Struct immutable, as the property referencing that struct can just be a `let` to control the immutability from a single place, while still allowing easy mutability for other cases without reaching into a full blown initializer

keith:
Thanks!

PR #300, opened 2018-04-23
Apply targeted device family to only application target

allu22:
Default targeted device family setting should be applied to only
application target. Otherwise it gets also applied to frameworks.


yonaskolb:
When creating a new framework in Xcode it applies the `TARGETED_DEVICE_FAMILY`. Is there a reason Apple does it? Is there a specific reason to remove it here?

PR #308, opened 2018-05-17
Support frameworks in source paths

keith:
This adds support for users passing pre-compiled frameworks in as a
source file of a target, and correctly adding it to the copy frameworks
build phase. Previously it would be added to the copy resources build
phase.

### DIFF
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -587,6 +587,7 @@ public class PBXProjGenerator {
             buildPhases.append(copyFilesPhase.reference)
         }
 
+        copyFrameworksReferences += getBuildFilesForPhase(.frameworks)
         if !copyFrameworksReferences.isEmpty {
 
             let copyFilesPhase = createObject(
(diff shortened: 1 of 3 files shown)


yonaskolb:
This would always embed those frameworks. Is that ok?

keith:
Yea I think that's safe since you still have to explicitly add them to your sources paths.


--
Keith Smiley

yonaskolb:
Ok. I guess for some this may be a breaking change, if they were manually embedding the framework via `dependencies` before. Now it might be embedded twice.

I want to make some framework embedding changes anyway though

keith:
I think anyone who's currently adding a framework to a source path is actually accidentally bloating their binary size for no reason. Just because they have it as a dependency doesn't mean it'll be in the sources right?

keith:
@yonaskolb so do you think we're good to merge this?

yonaskolb:
Yeah, it's only people who accidentally had frameworks in their sources (which was bloating their binary as you say) and also added that explicit framework path as a dependency. I don't think anyone would do that however. 
I think this is good to merge

keith:
Thanks!


--
Keith Smiley

PR #317, opened 2018-06-09
Add support for cc source files

maicki:
Hey! We are currently adding a new dependency to our app that contains cc files. Currently they are added as copying resource files phase instead of in the compiling source phase. After this diff it should be added to the right build phase. Cheers!


yonaskolb:
Great, thanks @maicki!

PR #324, opened 2018-06-27
Code sign on copy

### LINKED ISSUE #315: App Extensions failing when building with new build system
## tl;dr

Copying app extensions in a post-build phase includes the `CodeSignOnCopy` attribute, which creates a cycle when using the new build system. Xcode doesn't add the attribute, so it probably shouldn't be included.

## Problem

Copying the app extension into the main target creates a cycle when using the new build system: 

```

Showing Recent Messages
:-1: Cycle inside MyApp; building could produce unreliable results.
Cycle details:
→ Target 'MyApp' : CodeSign /Users/caesar/Library/Developer/Xcode/DerivedData/MyApp-akrqtdfekistduerxppbnpbmbzhr/Build/Products/Debug-iphonesimulator/MyApp.app/PlugIns/iMessageApp.appex
○ Target 'MyApp' : ValidateEmbeddedBinary /Users/caesar/Library/Developer/Xcode/DerivedData/MyApp-akrqtdfekistduerxppbnpbmbzhr/Build/Products/Debug-iphonesimulator/MyApp.app/PlugIns/iMessageApp.appex
○ Target 'MyApp' : CodeSign /Users/caesar/Library/Developer/Xcode/DerivedData/MyApp-akrqtdfekistduerxppbnpbmbzhr/Build/Products/Debug-iphonesimulator/MyApp.app/PlugIns/iMessageApp.appex
```

I have a [sample project](https://github.com/cjwirth/XcodeGenXcode10) that illustrates this.

I'm embedding the app extension using this build phase: 

```yaml
postBuildScripts:
  - name: Embed App Extensions
    script: |
      cp -r ${BUILT_PRODUCTS_DIR}/iMessage.appex ${TARGET_BUILD_DIR}/${PLUGINS_FOLDER_PATH}/iMessage.appex
```

Maybe I'm just doing this bit wrong, and this problem is just user error 🤷‍♂️ 

## Workaround

I am able to get the project to build by going into the Copy Files build phase that is created, removing the `iMessage.appex` from it, and then manually re-adding it to the build phase (which is what the [this commit in the sample project](https://github.com/cjwirth/XcodeGenXcode10/commit/dc1445a02d5840a64d553fee482d5b67af5c7d7d) is doing).

We use XcodeGen pretty heavily, so having to remember to manually remove/re-add the file each time isn't great.

## Explanation

The relevant part of the diff is basically:

```diff
- BF_887147700664 /* iMessageApp.appex in CopyFiles */ = {isa = PBXBuildFile; fileRef = FR_935153865209 /* iMessageApp.appex */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
+ 397D8E8520C85D0E00FFACDE /* iMessageApp.appex in CopyFiles */ = {isa = PBXBuildFile; fileRef = FR_935153865209 /* iMessageApp.appex */; settings = {ATTRIBUTES = (RemoveHeadersOnCopy, ); }; };
```

Or to zero in on it, the `CodeSignOnCopy` attribute isn't included when manually including the file through Xcode.

This kind of makes sense looking at the error message, then --> it looks like it's trying to codesign the same file twice, which isn't something that the new build system wants to do (?).

## Metadata

- XcodeGen Version: 1.8.0
- Xcode Version: 10.0 beta 1 (10L176w) (although also fails in Xcode 9.4)
- macOS Version: 10.13.4 (17E202)

yonaskolb:
Fixes #315

This only sets `CodeSignOnCopy` on non target embeds by default. It still respects an explicit `Dependency.codeSign`


eskerber:
LOL I was about to bug/fix this myself 😆

yonaskolb:
@eskerber Would you of done it this way? I can’t be certain that the default of whether to code sign is correct

eskerber:
@yonaskolb Kind of. I was going to use the same logic that buckets the `extensions` array to determine which should not be codesigned. 

As a coworker of mine pointed out Xcode goes so far as to _hide_ the code sign option in  Xcode if the object being embedded is an app extension:

2018-06-27 at 7 30 pm

yonaskolb:
Perhaps the decision could be based on the destination. So anything in `Plugins` wouldn't have code sign on copy

PR #327, opened 2018-06-29
Transitively embed dependencies

brentleyjones:
If a target embeds dependencies (i.e. App and Test bundles) and a dependency has dependencies that can be embedded, then we should embed those dependencies as long as they weren't embedded along the way.

This already happens for Carthage dependencies, the change makes it happen for non-Carthage frameworks as well.

Also, closes #219, since it correctly picks up on the transitive dependency and embeds it properly.

### DIFF
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -391,9 +391,10 @@ public class PBXProjGenerator {
         var copyWatchReferences: [String] = []
         var extensions: [String] = []
         
-        let dependenciesAndTransient = getAllDependenciesPlusTransientNeedingEmbedding(target: target)
+        let targetDependencies = (target.transientlyLinkDependencies ?? project.options.transientlyLinkDependencies) ?
+            getAllDependenciesPlusTransientNeedingEmbedding(target: target) : target.dependencies
 
-        for dependency in dependenciesAndTransient {
+        for dependency in targetDependencies {
 
             var embedAttributes: [String] = []
 
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -390,8 +390,10 @@ public class PBXProjGenerator {
         var copyResourcesReferences: [String] = []
         var copyWatchReferences: [String] = []
         var extensions: [String] = []
+        
+        let dependenciesAndTransient = getAllDependenciesPlusTransientNeedingEmbedding(target: target)
 
-        for dependency in target.dependencies {
+        for dependency in dependenciesAndTransient {
 
             var embedAttributes: [String] = []
 
@@ -441,10 +443,10 @@ public class PBXProjGenerator {
                     )
                     targetFrameworkBuildFiles.append(buildFile.reference)
                 }
-
-                if (dependency.embed ?? target.type.isApp) && !dependencyTarget.type.isLibrary {
-
-
+                
+                let shouldEmbed = target.type.isApp
+                    || (target.type.isTest && (dependencyTarget.type.isFramework || dependencyTarget.type == .bundle))
+                if (dependency.embed ?? shouldEmbed) && !dependencyTarget.type.isLibrary {
                     let embedFile = createObject(
                         id: dependencyFileReference + target.name,
                         PBXBuildFile(
@@ -497,6 +499,7 @@ public class PBXProjGenerator {
                     )
                     copyFrameworksReferences.append(embedFile.reference)
                 }
+                
             case .carthage:
                 var platformPath = Path(getCarthageBuildPath(platform: target.platform))
                 var frameworkPath = platformPath + dependency.reference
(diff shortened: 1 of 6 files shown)


keith:
This is definitely a nice addition. @yonaskolb what do you think about having a way to disable this behavior for those of us who'd prefer to manually manage this?

brentleyjones:
@keith If the code is correct it won’t hurt if you manually list them as well. Do you see it doing something that you would want to turn off, or is it more of a percausion? 

Related, this should (I need to confirm, or slightly adjust in a follow up PR) address proper use of resource bundles (#219) as well.

brentleyjones:
I just updated it to be explicit/cover the static library cases (we don't want to double-link those).

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -442,9 +444,7 @@ public class PBXProjGenerator {
                     targetFrameworkBuildFiles.append(buildFile.reference)
                 }
 
-                if (dependency.embed ?? target.type.isApp) && !dependencyTarget.type.isLibrary {
-
-
+                if embed && !dependencyTarget.type.isLibrary {
```
brentleyjones: This makes it so you don't have to explicitly list `embed: true` for test targets.

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -441,10 +443,10 @@ public class PBXProjGenerator {
...
                     targetFrameworkBuildFiles.append(buildFile.reference)
                 }
-
-                if (dependency.embed ?? target.type.isApp) && !dependencyTarget.type.isLibrary {
-
-
+                
+                let shouldEmbed = target.type.isApp
+                    || (target.type.isTest && (dependencyTarget.type.isFramework || dependencyTarget.type == .bundle))
+                if (dependency.embed ?? shouldEmbed) && !dependencyTarget.type.isLibrary {
```
brentleyjones: This makes it so you don't have to explicitly list embed: true for test targets when it's a framework or bundle. It already does this for Carthage.

keith:
The problem is there are 2 cases that are very hard for XcodeGen to have knowledge about, and therefore get right if you're trying to do this:

1. Static frameworks. The only attribute we have for this here is `type.isFramework` which is purely based on the product type. There's no insight to what `MACH_O_TYPE` is set here, or in an xcconfig file. In this case you would choose to embed something, when really that framework should be (mostly) treated as a static library
2. Test targets that set `BUNDLE_LOADER`. These test targets end up inheriting any symbols linked to the binary set as the `BUNDLE_LOADER`, without having to link them themselves. This means you want to _build_ the dependent targets, but not link them. In this case all you have to check is `type.isTest`, which will be true for these, but you don't want to actually link any static libraries it depends on.

keith:
Also on a separate note for us I'm worried about the performance implications of this in our project with ~300 targets. Right now we resolve this info ourselves in a different pass, so this would just be overhead. But that's more specific to how we use XcodeGen

brentleyjones:
@Keith Perfect examples. I will make a config for this.

Related to your points though:

1. This is a great example. With buck the linking type is clearly spelled out, and I wonder if we should do the same for when the default isn't correct (a framework is assumed to be dynamic, but we can tag it as static, so we don't have to mess with the `embed` flag). This would be something I would tackle after though, since it's a much more intrusive fix.
2. I hope on fixing this soon in a followup PR. Right now Carthage is being run for all of the `BUNDLE_LOADER` binary dependencies when it's not needed for the sam. I would fix it for all frameworks at once.

brentleyjones:
@Keith I added some flags for this. Let me know what you think.

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -391,9 +391,11 @@ public class PBXProjGenerator {
         var copyWatchReferences: [String] = []
         var extensions: [String] = []
         
-        let dependenciesAndTransient = getAllDependenciesPlusTransientNeedingEmbedding(target: target)
+        let transientlyLinkDependencies = target.transientlyLinkDependencies ?? project.options.transientlyLinkDependencies
+        let targetDependencies = (target.transientlyLinkDependencies ?? project.options.transientlyLinkDependencies) ?
```
keith: looks like this is the same as `transientlyLinkDependencies` above

keith:
Thanks! Seems reasonable to me!

Sources/XcodeGenKit/PBXProjGenerator.swift
brentleyjones: Opps, missed on refactor. Will remove.

yonaskolb:
Also as a note @keith, it is possible to lookup a specific build setting for at target using `Project.getCombinedBuildSettings` (for example for `MACH_O_TYPE`). That will look at the resolved setting and works with any references to nested xcconfigs as well

yonaskolb:
I might just merge this now and we can look at reuse and caching in a followup PR

keith:
Just realized I think the wording here was intended to be "transitively" not "transiently"

brentleyjones:
Good catch, we should fix that 😬.

keith:
If we're going to, it should be quick since the setting is also mispelled

brentleyjones:
I'll make a PR in a couple mins.

keith:
To follow up on the speed part here, with this feature enabled, benchmarking with [hyperfine](https://github.com/sharkdp/hyperfine):

with this feature enabled:

```
Time (mean ± σ):     11.310 s ±  0.413 s    [User: 10.446 s, System: 0.813 s]
Range (min … max):   10.927 s … 12.275 s
```

with this feature disabled:

```
Time (mean ± σ):     10.919 s ±  0.176 s    [User: 10.116 s, System: 0.768 s]
Range (min … max):   10.764 s … 11.369 s
```

yonaskolb:
Is that on the lyft codebase?

keith:
Yes

keith:
Definitely happy to save the half a second by disabling this

PR #341, opened 2018-07-15
Set higher version .xcdatamodeld file as current

rohitpal440:
Sort .xcdatamodeld files in decreasing order to set higher version core data  model as current.

Current sorting of file takes place as:

```
ABC 10.xcdatamodel
ABC 11.xcdatamodel
ABC 12.xcdatamodel
ABC 13.xcdatamodel
ABC 14.xcdatamodel
ABC 15.xcdatamodel
ABC 16.xcdatamodel
ABC 17.xcdatamodel
ABC 18.xcdatamodel
ABC 19.xcdatamodel
ABC 2.xcdatamodel
ABC 20.xcdatamodel
ABC 3.xcdatamodel
ABC 4.xcdatamodel
ABC 5.xcdatamodel
ABC 6.xcdatamodel
ABC 7.xcdatamodel
ABC 8.xcdatamodel
ABC 9.xcdatamodel
```

after this change sorting will be like

```
ABC 20.xcdatamodel
ABC 19.xcdatamodel
ABC 18.xcdatamodel
ABC 17.xcdatamodel
ABC 16.xcdatamodel
ABC 15.xcdatamodel
ABC 14.xcdatamodel
ABC 13.xcdatamodel
ABC 12.xcdatamodel
ABC 11.xcdatamodel
ABC 10.xcdatamodel
ABC 9.xcdatamodel
ABC 8.xcdatamodel
ABC 7.xcdatamodel
ABC 6.xcdatamodel
ABC 5.xcdatamodel
ABC 4.xcdatamodel
ABC 3.xcdatamodel
ABC 2.xcdatamodel
```

and First element will be set as current core data model

### DIFF
--- a/Sources/XcodeGenKit/SourceGenerator.swift
+++ b/Sources/XcodeGenKit/SourceGenerator.swift
@@ -119,7 +119,7 @@ class SourceGenerator {
                 let models = (try? path.children()) ?? []
                 let modelFileReference = models
                     .filter { $0.extension == "xcdatamodel" }
-                    .sorted()
+                    .sorted { $0.string.localizedStandardCompare($1.string) == .orderedDescending }
                     .map { path in
                         createObject(
                             id: path.byRemovingBase(path: project.basePath).string,
--- a/Sources/XcodeGenKit/SourceGenerator.swift
+++ b/Sources/XcodeGenKit/SourceGenerator.swift
@@ -119,7 +119,7 @@ class SourceGenerator {
                 let models = (try? path.children()) ?? []
                 let modelFileReference = models
                     .filter { $0.extension == "xcdatamodel" }
-                    .sorted { $0.string.localizedStandardCompare($1.string) == .orderedDescending }
+                    .sorted { $0.string.localizedStandardCompare($1.string) == .orderedAscending }
                     .map { path in
                         createObject(
                             id: path.byRemovingBase(path: project.basePath).string,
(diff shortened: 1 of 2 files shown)


yonaskolb:
Yes, specifically this line https://github.com/rohitpal440/XcodeGen/blob/b0fa7a2845213aeeaa4caf07ad30eb060bfdf00c/Sources/XcodeGenKit/PBXProjGenerator.swift#L291
Sorry, I guess the line number didn't copy across on my phone properly

Sources/XcodeGenKit/SourceGenerator.swift
```diff
@@ -119,7 +119,7 @@ class SourceGenerator {
                 let models = (try? path.children()) ?? []
                 let modelFileReference = models
                     .filter { $0.extension == "xcdatamodel" }
-                    .sorted()
+                    .sorted { $0.string.localizedStandardCompare($1.string) == .orderedDescending }
```
yonaskolb: Even though this will be rewritten in `sortGroups` shouldn't this be `.orderedAscending`?

rohitpal440: Then I've to do something like ` currentVersion: modelFileReference.last?.reference` over here https://github.com/rohitpal440/XcodeGen/blob/4d964c07861cd524801b85d4edf611c77e1d2c3b/Sources/XcodeGenKit/SourceGenerator.swift#L134
 to make higher version of `.xcdatamodel` file as current

rohitpal440: So what do you suggest me to do? 
1. Keep the code as in this PR
OR
2. Compare using `.orderedAscending` and update the code for getting currentVersion to `currentVersion: modelFileReference.last?.reference`

In both ways I'm going to get the same result, chose approach 1 since the change stays in one place instead of 2.

yonaskolb: I'd go with option 2. As you want to sort in ascending order

Issue #344, opened 2018-07-17
module maps should be copied to Products Directory


brentleyjones:
Currently if a `.modulemap` is in a source folder it will be copied as a resource. It should instead be copied to the Products Directory at the path `include/$(PRODUCT_NAME)`.

There currently is no way to manually specify this either, since `resources` is the only real copy phase supported. Ideally this could be manually specified as well with more `buildPhase`s.

PR #345, opened 2018-07-17
Add support for Copy Files build phase

brentleyjones:
Related to #344, this allows manually setting up Copy Files build phases.

### DIFF
--- a/Docs/ProjectSpec.md
+++ b/Docs/ProjectSpec.md
@@ -260,21 +260,33 @@ A source can be provided via a string (the path) or an object of the form:
 	- `sources` - Compile Sources phase
 	- `resources` - Copy Bundle Resources phase
 	- `headers` - Headers Phase
-	- `none` - Will not be added to any build phases
+	- `copyFiles` - Copy Files Phase. Must be specified as an object with the following fields:
+		- [x] **destination**: **String** - Destination of the Copy Files phase. This can be one of the following values:
+			- `absolutePath`
+			- `productsDirectory`
+			- `wrapper`
+			- `executables`
+			- `resources`
+			- `javaResources`
+			- `frameworks`
+			- `sharedFrameworks`
+			- `sharedSupport`
+			- `plugins`
+		- [ ] **subpath**: **String** - The path inside of the destination to copy the files.
 - [ ] **type**: **String**: This can be one of the following values
 	- `file`: a file reference with a parent group will be created (Default for files or directories with extensions)
 	- `group`: a group with all it's containing files. (Default for directories without extensions)
 	- `folder`: a folder reference.
-- [ ] **headerVisibility**: **String** The visibility of any headers. This defaults to `public`, but can be either:
+- [ ] **headerVisibility**: **String** - The visibility of any headers. This defaults to `public`, but can be either:
 	- `public`
 	- `private`
 	- `project`
 
 ```yaml
 targets:
-  MyTarget
+  MyTarget:
     sources: MyTargetSource
-  MyOtherTarget
+  MyOtherTarget:
     sources:
       - MyOtherTargetSource1
       - path: MyOtherTargetSource2
@@ -288,6 +300,11 @@ targets:
           - "-Wextra"
       - path: MyOtherTargetSource3
         compilerFlags: "-Werror -Wextra"
+      - path: ModuleMaps
+        buildPhase:
+          copyFiles:
+            destination: productsDirectory
+            subpath: include/$(PRODUCT_NAME)
       - path: Resources
         type: folder
 ```
--- a/Docs/ProjectSpec.md
+++ b/Docs/ProjectSpec.md
@@ -273,6 +273,7 @@ A source can be provided via a string (the path) or an object of the form:
 			- `sharedSupport`
 			- `plugins`
 		- [ ] **subpath**: **String** - The path inside of the destination to copy the files.
+	- `none` - Will not be added to any build phases
 - [ ] **type**: **String**: This can be one of the following values
 	- `file`: a file reference with a parent group will be created (Default for files or directories with extensions)
 	- `group`: a group with all it's containing files. (Default for directories without extensions)
--- a/Sources/ProjectSpec/SpecParsingError.swift
+++ b/Sources/ProjectSpec/SpecParsingError.swift
@@ -5,6 +5,7 @@ public enum SpecParsingError: Error, CustomStringConvertible {
     case unknownTargetPlatform(String)
     case invalidDependency([String: Any])
     case unknownSourceBuildPhase(String)
+    case invalidSourceCopyFilesPhase
     case invalidVersion(String)
 
     public var description: String {
@@ -17,6 +18,8 @@ public enum SpecParsingError: Error, CustomStringConvertible {
             return "Unknown Target dependency: \(dependency)"
         case let .unknownSourceBuildPhase(buildPhase):
             return "Unknown Source Build Phase: \(buildPhase)"
+        case .invalidSourceCopyFilesPhase:
+            return "copyFiles Build Phase without a copyFiles section"
         case let .invalidVersion(version):
             return "Invalid version: \(version)"
         }
--- a/Sources/ProjectSpec/SpecParsingError.swift
+++ b/Sources/ProjectSpec/SpecParsingError.swift
@@ -4,8 +4,7 @@ public enum SpecParsingError: Error, CustomStringConvertible {
     case unknownTargetType(String)
     case unknownTargetPlatform(String)
     case invalidDependency([String: Any])
-    case unknownSourceBuildPhase(String)
-    case invalidSourceCopyFilesPhase
+    case invalidSourceBuildPhase(String)
     case invalidVersion(String)
 
     public var description: String {
@@ -16,10 +15,8 @@ public enum SpecParsingError: Error, CustomStringConvertible {
             return "Unknown Target platform: \(platform)"
         case let .invalidDependency(dependency):
             return "Unknown Target dependency: \(dependency)"
-        case let .unknownSourceBuildPhase(buildPhase):
-            return "Unknown Source Build Phase: \(buildPhase)"
-        case .invalidSourceCopyFilesPhase:
-            return "copyFiles Build Phase without a copyFiles section"
+        case let .invalidSourceBuildPhase(error):
+            return "Invalid Source Build Phase: \(error)"
         case let .invalidVersion(version):
             return "Invalid version: \(version)"
         }
--- a/Sources/ProjectSpec/TargetSource.swift
+++ b/Sources/ProjectSpec/TargetSource.swift
@@ -159,10 +159,10 @@ extension TargetSource.BuildPhase {
         case "headers": self = .headers
         case "resources": self = .resources
         case "copyFiles":
-            throw SpecParsingError.invalidSourceCopyFilesPhase
+            throw SpecParsingError.invalidSourceBuildPhase("copyFiles must specify directory")
         case "none": self = .none
         default:
-            throw SpecParsingError.unknownSourceBuildPhase(string)
+            throw SpecParsingError.invalidSourceBuildPhase("Unknown build phase \(string.quoted)")
         }
     }
 }
--- a/Sources/ProjectSpec/TargetSource.swift
+++ b/Sources/ProjectSpec/TargetSource.swift
@@ -159,10 +159,10 @@ extension TargetSource.BuildPhase {
         case "headers": self = .headers
         case "resources": self = .resources
         case "copyFiles":
-            throw SpecParsingError.invalidSourceBuildPhase("copyFiles must specify directory")
+            throw SpecParsingError.invalidSourceBuildPhase("copyFiles must specify a \"destination\" and optional \"subpath\"")
         case "none": self = .none
         default:
-            throw SpecParsingError.invalidSourceBuildPhase("Unknown build phase \(string.quoted)")
+            throw SpecParsingError.invalidSourceBuildPhase(string.quoted)
         }
     }
 }
--- a/Sources/ProjectSpec/TargetSource.swift
+++ b/Sources/ProjectSpec/TargetSource.swift
@@ -28,17 +28,59 @@ public struct TargetSource: Equatable {
         }
     }
 
-    public enum BuildPhase: String {
+    public enum BuildPhase: Equatable {
         case sources
         case headers
         case resources
+        case copyFiles(CopyFilesSettings)
         case none
+        // Not currently exposed as selectable options, but used internally
+        case frameworks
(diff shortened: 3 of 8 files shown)


Sources/ProjectSpec/TargetSource.swift
```diff
@@ -32,17 +33,52 @@ public struct TargetSource: Equatable {
...
             case .sources: return .sources
             case .headers: return .headers
             case .resources: return .resources
+            case .copyFiles: return .copyFiles
             case .none: return nil
             }
         }
     }
+    
+    public struct CopyFiles: Equatable, Hashable {
```
brentleyjones: Thinking about this, maybe this should be called `CopyFilesBuildPhaseSettings`. I know it's longer, but I think it makes it clear what it's for.

Docs/ProjectSpec.md
```diff
@@ -251,12 +251,14 @@ A source can be provided via a string (the path) or an object of the form:
 	- `sources` - Compile Sources phase
 	- `resources` - Copy Bundle Resources phase
 	- `headers` - Headers Phase
+	- `copyFiles` - Copy Files Phase
 	- `none` - Will not be added to any build phases
+- [ ] **copyFiles**: **[Copy Files](#copy-files)** - Options for the Copy Files build phase. If specified the `buildPhase` needs to be `copyFiles` or unspecified (in which case it will default to `copyFiles`).
```
brentleyjones: Instead of this I could make it so when you specify `copyFiles` for `buildPhase` you have to do it as an object:

```yml
- buildPhase:
    type: copyFiles
    destination: productsDirectory
    subpath: include/$(PRODUCT_NAME)
```

This might complicate it too much though, which is why I went the route I did.

Sources/ProjectSpec/TargetSource.swift
```diff
@@ -12,6 +12,7 @@ public struct TargetSource: Equatable {
     public var type: SourceType?
     public var optional: Bool
     public var buildPhase: BuildPhase?
+    public var copyFiles: CopyFiles?
```
brentleyjones: Instead of adding another field, I’m going to make `BuildPhase` have an associated value. It makes my follow up PR easier as well as cleaning up the paths that don’t care about copyFiles.

Docs/ProjectSpec.md
brentleyjones: Also, I believe the way people will specify this is just by using the `copyFiles` object and not specifying a `buildPhase` manually.

brentleyjones:
I'll update `CHANGELOG.md` soon. I keep forgetting about that step.

brentleyjones:
@yonaskolb Not sure on the protocol here. Is this fine for me to merge?

yonaskolb:
There were just a couple of comments I wanted to make. Just on the road at the moment, but will get to them all tomorrow 😀👍

Sources/ProjectSpec/TargetSource.swift
```diff
@@ -100,12 +142,45 @@ extension TargetSource: JSONObjectConvertible {
...
+}
+
+extension TargetSource.BuildPhase {
+    
+    public init?(string: String?, copyFilesSettings: TargetSource.BuildPhase.CopyFilesSettings? = nil) throws {
+        let chosenString: String
+        if let string = string {
+            chosenString = string
+        } else if copyFilesSettings != nil {
+            chosenString = "copyFiles"
```
yonaskolb: The downside of this is that if they have a `TargetSource` with mixed files, there's no way to just specify the settings for the ones that will be in `copyFiles` by default, because if you do that everything will be moved to the `copyFiles` build phase.
I would be ok with keeping the `copyFiles` struct out of the enum like you had it before. The validation error could then be thrown in the generator when searching through the files and finding any copyFiles build phase `SourceFiles` (like modulemap) and it not having a `copyFiles` settings.

Are there other sorts of files you would want to add to `copyFiles`, or is it just `modulemap`?

Docs/ProjectSpec.md
```diff
@@ -279,10 +281,31 @@ targets:
           - "-Wextra"
       - path: MyOtherTargetSource3
         compilerFlags: "-Werror -Wextra"
+    - path: ModuleMaps
+        copyFiles:
```
yonaskolb: indentation

Sources/ProjectSpec/TargetSource.swift
```diff
@@ -28,17 +28,59 @@ public struct TargetSource: Equatable {
...
+            public var destination: Destination
+            public var subpath: String
+        }
 
         public var buildPhase: xcproj.BuildPhase? {
             switch self {
             case .sources: return .sources
             case .headers: return .headers
             case .resources: return .resources
+            case .copyFiles(_): return .copyFiles
```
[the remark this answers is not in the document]
brentleyjones: I did it without and got an error. I’ll try again.

Sources/ProjectSpec/TargetSource.swift
brentleyjones: Header files in the case of static libraries.

Let me think about your downside comment. When I had it different before it would default the whole sources to the copy files build phase, and the way I see you recommending it is that _if_ it were determined to need to be copied it should use these settings. Right now I’m treating it exactly like all the other build phases and if specified they override any default logic. This phase just happens to require additional configuration options.

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -576,6 +592,19 @@ public class PBXProjGenerator {
...
+        
+        let copyFilesBuildPhasesFiles = getBuildFilesForCopyFilesPhases()
+        if !copyFilesBuildPhasesFiles.isEmpty {
+            for (copyFiles, buildPhaseFiles) in copyFilesBuildPhasesFiles {
+                let buildPhase = PBXCopyFilesBuildPhase(
+                    dstPath: copyFiles.subpath,
+                    dstSubfolderSpec: copyFiles.destination.destination,
+                    files: buildPhaseFiles
+                )
+                let copyFilesBuildPhase = createObject(id:  target.name, buildPhase)
```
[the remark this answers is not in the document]
brentleyjones: Yes, I’ll adjust 🙂.

Sources/ProjectSpec/TargetSource.swift
```diff
@@ -28,17 +28,59 @@ public struct TargetSource: Equatable {
...
 
-    public enum BuildPhase: String {
+    public enum BuildPhase: Equatable {
         case sources
         case headers
         case resources
+        case copyFiles(CopyFilesSettings)
         case none
+        // Not currently exposed as selectable options, but here to translate to PBXCopyFilesBuildPhase completely
+        case frameworks
```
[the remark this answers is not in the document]
brentleyjones: Yes, since this type is now used where xcproj ones were before, to allow defaulting to work correctly (see the `modulemap` PR for an example).

brentleyjones:
@yonaskolb I adressed all your feedback, expect for 
https://github.com/yonaskolb/XcodeGen/pull/345#discussion_r204228072. With that I changed the structure so `copyFiles` is simply an object specified for `buildPhase`. It's only special in the sense that it has additional required fields to be specified. It behaves the same as all the other build phases, if you specify it you mean it for all of the files (so use excludes or be specific). Defaults will be handled by #346 and a PR I plan on making soon for static library headers.

yonaskolb:
Cool, I'll take a look later 😄👍 
In the future if a PR has been reviewed could you add new commits instead of rebasing? It makes it easier to see the changes since reviewing. We always have the option of squashing on merge if there are lots of commits

brentleyjones:
Will do! I'll add commits next time (I rebased a couple other PRs, sorry, I won't in the future).

yonaskolb:
Also I should clarify. It's ok to rebase off the target branch to get new changes, but I meant not amending or squashing new commits 👍

brentleyjones:
@yonaskolb Rebased and ready to go.

Sources/ProjectSpec/TargetSource.swift
```diff
@@ -28,17 +28,59 @@ public struct TargetSource: Equatable {
...
 
-    public enum BuildPhase: String {
+    public enum BuildPhase: Equatable {
         case sources
         case headers
         case resources
+        case copyFiles(CopyFilesSettings)
         case none
+        // Not currently exposed as selectable options, but used internally
+        case frameworks
```
yonaskolb: indenting

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -591,6 +605,22 @@ public class PBXProjGenerator {
...
             buildPhases.append(resourcesBuildPhase.reference)
         }
+        
+        let copyFilesBuildPhasesFiles = getBuildFilesForCopyFilesPhases()
+        if !copyFilesBuildPhasesFiles.isEmpty {
+            for (copyFiles, buildPhaseFiles) in copyFilesBuildPhasesFiles {
+                let copyFilesBuildPhase = createObject(
+                    id: "copy files" + copyFiles.destination.rawValue + copyFiles.subpath + target.name,
+                    PBXCopyFilesBuildPhase(
+                        dstPath: copyFiles.subpath,
```
yonaskolb: I guess you've tested if a the default `subpath` of `""` is valid, as opposed to nil? As the value is optional in `PBXCopyFilesBuildPhase`. Just trying to minimise diffs when projects are opened in Xcode

Sources/ProjectSpec/TargetSource.swift
```diff
@@ -100,12 +142,42 @@ extension TargetSource: JSONObjectConvertible {
...
+    public init(jsonDictionary: JSONDictionary) throws {
+        self = .copyFiles(try jsonDictionary.json(atKeyPath: "copyFiles"))
+    }
+}
+
+extension TargetSource.BuildPhase.CopyFilesSettings: JSONObjectConvertible {
+    
+    public init(jsonDictionary: JSONDictionary) throws {
+        destination = try jsonDictionary.json(atKeyPath: "destination")
+        subpath = jsonDictionary.json(atKeyPath: "subpath") ?? ""
```
yonaskolb: Would the default be nil?

Sources/ProjectSpec/SpecParsingError.swift
```diff
@@ -5,6 +5,7 @@ public enum SpecParsingError: Error, CustomStringConvertible {
     case unknownTargetPlatform(String)
     case invalidDependency([String: Any])
     case unknownSourceBuildPhase(String)
+    case invalidSourceCopyFilesPhase
```
yonaskolb: This might get unwieldy if we add more build phase enum options. Can we just used the existing `unknownSourceBuildPhase` but renamed to `invalidSourceBuildPhase` along with the error. Then the string can be `copyFiles must be a dictionary`

brentleyjones: Yep. I’ll adjust.

Sources/XcodeGenKit/PBXProjGenerator.swift
brentleyjones: I’ll reaffirm.

Sources/ProjectSpec/TargetSource.swift
brentleyjones: I'll adjust the docs to say it's optional. We should still default it to `""` though to match Xcode behavior (there isn't a way to make it not `""` by using just Xcode).

Sources/XcodeGenKit/PBXProjGenerator.swift
brentleyjones: The default is `""`. If you don't specify it, it won't exist, which is still valid, but to match what Xcode does naturally it's `""` specified. Xcode won't change when opening either way.

Sources/ProjectSpec/TargetSource.swift
brentleyjones: @yonaskolb I’m not sure what you are referring to here. Do you want a newline before the comment, the comment to be indented differently, or something else?

brentleyjones:
@yonaskolb All feedback addressed besides for the indentation one (I left a question). This is G2G 😄.

Sources/XcodeGenKit/PBXProjGenerator.swift
yonaskolb: 👍 
Sometimes Xcode won't change the project on open, but if something else changes (like reordering some files is an easy way to test), then it will rewrite things in it's preferred way

Sources/ProjectSpec/TargetSource.swift
yonaskolb: Hmm, it's not showing up now. I swear `case frameworks` looked indented differently before

Docs/ProjectSpec.md
```diff
@@ -260,21 +260,33 @@ A source can be provided via a string (the path) or an object of the form:
 	- `sources` - Compile Sources phase
 	- `resources` - Copy Bundle Resources phase
 	- `headers` - Headers Phase
-	- `none` - Will not be added to any build phases
```
yonaskolb: Has none been removed?

brentleyjones: No. I’m not sure how that happened. I’ll fix it.

Sources/ProjectSpec/TargetSource.swift
```diff
@@ -100,12 +142,42 @@ extension TargetSource: JSONObjectConvertible {
...
+
+extension TargetSource.BuildPhase {
+    
+    public init(string: String) throws {
+        switch string {
+        case "sources": self = .sources
+        case "headers": self = .headers
+        case "resources": self = .resources
+        case "copyFiles":
+            throw SpecParsingError.invalidSourceBuildPhase("copyFiles must specify directory")
```
yonaskolb: Should this be `copyFiles must specify a \"destination\" and optional \"subPath\"`

Sources/ProjectSpec/TargetSource.swift
```diff
@@ -100,12 +142,42 @@ extension TargetSource: JSONObjectConvertible {
...
+        case "none": self = .none
+        default:
+            throw SpecParsingError.invalidSourceBuildPhase("Unknown build phase \(string.quoted)")
```
yonaskolb: I guess this can just be the `string`

Sources/ProjectSpec/TargetSource.swift
brentleyjones: Yes, that's more descriptive.

Sources/ProjectSpec/TargetSource.swift
brentleyjones: :+1:

yonaskolb:
Feel free to merge

yonaskolb:
Nice work @brentleyjones! 🎉

PR #346, opened 2018-07-17
Automatically copy modulemap files into products directory

### LINKED ISSUE #344: module maps should be copied to Products Directory

brentleyjones:
Resolves #344.

Depends on #345 (this is actually a small change on top of it).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 - Improved linking for `static.library` targets [352](https://github.com/yonaskolb/XcodeGen/pull/352) @brentleyjones
 - Changed default group sorting to be after files [356](https://github.com/yonaskolb/XcodeGen/pull/356) @yonaskolb
 - Moved `Frameworks` and `Products` top level groups to bottom [356](https://github.com/yonaskolb/XcodeGen/pull/356) @yonaskolb
+- `modulemap` files are automatically copied to the products directory for static library targets [346](https://github.com/yonaskolb/XcodeGen/pull/346) @brentleyjones
 
 #### Internal
 - Moved brew formula to homebrew core


brentleyjones:
I’ll add something to the test fixture for this as well.

yonaskolb:
Hi @brentleyjones. Could you please retarget this PR onto the branch it is based off. That makes it easier to actually see the diff 👍

brentleyjones:
@yonaskolb Since both are in my fork, it won’t let me. You can just select the last commit though in Files Changed and review from there.

yonaskolb:
Sure will do. In the future you can create branches in this repo as you’re a collaborator now 😉

brentleyjones:
After digging into this more (#350) we should only copy this over if the target is a static library, since other types don’t need this behavior. Otherwise the file should be ignored (by default). I’ll adjust for this.

Sources/XcodeGenKit/SourceGenerator.swift
```diff
@@ -156,14 +156,20 @@ class SourceGenerator {
...
-    private func getDefaultBuildPhase(for path: Path) -> TargetSource.BuildPhase? {
+    private func getDefaultBuildPhase(for path: Path, targetType: PBXProductType) -> TargetSource.BuildPhase? {
         if path.lastComponent == "Info.plist" {
             return nil
         }
         if let fileExtension = path.extension {
             switch fileExtension {
             case "swift", "m", "mm", "cpp", "c", "cc", "S", "xcdatamodeld": return .sources
             case "h", "hh", "hpp", "ipp", "tpp", "hxx", "def": return .headers
+            case "modulemap":
```
yonaskolb: Are there other sorts of files that would have different `CopyFilesSettings`?
If not, then we maybe don't need to pass `PBXProductType` all the way down, and these settings can be provided in `PBXProjGenerator`

brentleyjones: Headers for `.staticLibrary` will need to use copy files as well, so we will need to pass the `PBXProductType` down.

PR #349, opened 2018-07-19
Add minimumXcodeGenVersion option

brentleyjones:
If XcodeGen's version is less than this version validation will fail.

### DIFF
--- a/Sources/XcodeGen/main.swift
+++ b/Sources/XcodeGen/main.swift
@@ -40,8 +40,11 @@ func generate(spec: String, project: String, isQuiet: Bool, justVersion: Bool) {
 
     do {
         logger.info("⚙️  Generating project...")
+        
+        try project.validateMinimumXcodeGenVersion(version)
+        
         let projectGenerator = ProjectGenerator(project: project)
-        let xcodeProject = try projectGenerator.generateXcodeProject(xcodeGenVersion: version)
+        let xcodeProject = try projectGenerator.generateXcodeProject()
 
         logger.info("⚙️  Writing project...")
 
--- a/Sources/XcodeGen/main.swift
+++ b/Sources/XcodeGen/main.swift
@@ -6,7 +6,7 @@ import ProjectSpec
 import XcodeGenKit
 import xcproj
 
-let version = "1.10.3"
+let version = try Version("1.10.3")
 
 func generate(spec: String, project: String, isQuiet: Bool, justVersion: Bool) {
     if justVersion {
@@ -90,4 +90,4 @@ command(
         description: "Show XcodeGen version"
     ),
     generate
-).run(version)
+).run(version.description)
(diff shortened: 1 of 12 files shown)


Sources/XcodeGen/main.swift
```diff
@@ -41,7 +41,7 @@ func generate(spec: String, project: String, isQuiet: Bool, justVersion: Bool) {
     do {
         logger.info("⚙️  Generating project...")
         let projectGenerator = ProjectGenerator(project: project)
-        let xcodeProject = try projectGenerator.generateXcodeProject()
+        let xcodeProject = try projectGenerator.generateXcodeProject(xcodeGenVersion: version)
```
yonaskolb: I agree this can be a useful feature, but at the moment I feel its affects are a bit too pervasive in the codebase. I'd be happy to have an extension `Project.meetsMinimumVersion(_ version: Version) -> Bool`  that gets checked for in this file and throws an error.

XcodeGenKit is also a library that gets used by other packages as well.

brentleyjones: Sure. I’ll adjust.

brentleyjones:
@yonaskolb I made it less pervasive. Let me know if there is anything else needed 😄.

Tests/XcodeGenKitTests/ProjectSpecTests.swift
```diff
@@ -74,6 +74,17 @@ class ProjectSpecTests: XCTestCase {
                 configSettings: ["invalidConfig": [:]],
                 groups: ["invalidSettingGroup"]
             )
+            
+            $0.it("fails with invalid XcodeGen version") {
+                let minimumVersion = try! Version("1.11.1")
```
yonaskolb: No biggie, but these tests are allowed to throw so no need for the exclamation marks

Tests/XcodeGenKitTests/ProjectSpecTests.swift
```diff
@@ -205,3 +216,9 @@ fileprivate func expectValidationError(_ project: Project, _ expectedError: Spec
     }
     throw failure("Supposed to fail with \"\(expectedError)\"", file: file, line: line)
 }
+
+fileprivate func expectMinimumXcodeGenVersionError(_ project: Project, minimumVersion: Version, xcodeGenVersion: Version, file: String = #file, line: Int = #line) throws {
```
yonaskolb: Can this just go inside the `it` closure, as it's only used there

Tests/XcodeGenKitTests/ProjectSpecTests.swift
brentleyjones: Ahh, true. I'll adjust it.

Tests/XcodeGenKitTests/ProjectSpecTests.swift
brentleyjones: Sounds good.

Issue #350, opened 2018-07-19
Better static library support


brentleyjones:
Static libraries that are used by other targets in the same project/workspace have some conventions that need be followed that Xcode doesn't easily provide by default:

- Need to have a `Copy Files` phase for all public headers
    - Needs to include a `PRODUCT_NAME-Swift.h` if it's a Swift library and it's going to be consumed by Objective-C.
    - Needs to include a `modulemap` if it's going to be `@import`ed in Objective-C, or if it's an Objective-C library going to be `import`ed into Swift.
- Need to only link to the final executable to prevent duplicate symbols.
    - If the code includes Objective-C categories (or Swift extensions on Objective-C classes) it needs to be linked with `-ObjC` or the linker will drop those symbols.
- Need to not link against other libraries or frameworks (else https://github.com/CocoaPods/CocoaPods/issues/7251 like warnings happen)
- Need to have `SKIP_INSTALL` set to `YES` to prevent an invalid archive.

Here are my plans for how we can improve the use of static libraries with XcodeGen to address the above requirements:

1.  ✅ `static.library` targets will copy (#345) headers in [Sources](../blob/master/Docs/ProjectSpec.md#sources) with `public` `headerVisibility` to `include/$(PRODUCT_NAME)` in the Products Directory, as long as the `buildPhase` is `headers` or not specified. – #365 
    i. If any Swift files are included in Sources, and `SWIFT_OBJC_INTERFACE_HEADER_NAME` is not an empty string, the Swift-ObjC Interface Header will be copied to the same location with a script. – #366
    ii. ✅ To support modules a `module.modulemap` will need to be manually supplied. If it's included in Sources it will be copied to the same location. – #346 (_I need to see if there is a way to manually generate this similar to how Xcode does for frameworks. I also need to see if we need to set `MODULEMAP_FILE`, since I haven't needed to yet._)

1. If a `target` dependency of type `static.library` is declared for a target:
    1. ✅ `link` will default to `true` only for `isExecutable` targets. If a dynamic framework needs to link against a static library (maybe to create a wrapper) the default can be changed. – #352
    1. ✅ A new `requiresObjCLinking` property on 
[Target](../blob/master/Docs/ProjectSpec.md#target), if `true`, will add `-ObjC` to `OTHER_LDFLAGS` for targets that link against it. It will default to `true` if `type` is `library.static`. – #353
    1. ✅ If `transitivelyLinkDependencies` is `true` static libraries will now propagate. – #352
[...]

Issue #343, opened 2018-07-17
Carthage Static Dependencies Support


peymankh:
Hi @yonaskolb 

Quick question about Carthage's new static framework support. To my understanding, Carthage will create a separate directory under `build/{Platform}` called `Static`. Do we need to do anything extra to get support for this in Xcodegen.

From the top of my mind, the dependency resolver and framework search paths probably need to be updated?

Thanks.

yonaskolb:
Hi @peymankh 
From looking at Carthage's documentation it seems what we need to do is to add a `static` property to `Dependency` for carthage dependencies that would control: 
- which directory we look for the framework in. Static would look in `Carthage/Build/$(PLATFORM_NAME)/Static`
- whether to add the framework to the Carthage `copy-frameworks` input and output files

brentleyjones:
@yonaskolb That property (or something similar) would be useful for the other forms of `Dependency` as well. Only mentioning it so we name it properly from the start (I’m hoping to tackle the non-Carthage versions of it soon). 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

yonaskolb:
@brentleyjones yep, my thinking too 👍

brentleyjones:
@yonaskolb Think I can take a stab at this first? I want to make sure it fits the grander static library/framework vision I'm trying to roll out. If not, that's cool, but one change I was thinking was instead of `static` we call the property `preferredLinkage` (similar to the [buck property](https://buckbuild.com/rule/prebuilt_apple_framework.html#preferred_linkage)) and it defaults to `dynamic` and can be also be set to `static` or `none`. I plan on using the same property and values for other dependency types as well as for `framework` targets.

yonaskolb:
Sounds good. Go for it 👍

peymankh:
@yonaskolb @brentleyjones 

Hey guys,

[...]

PR #352, opened 2018-07-19
Improve linking for static library targets

brentleyjones:
Child of #350. Addresses 2.i, 2.iii, 3.

With these changes there are fewer warnings in the new build system (no longer links dynamic frameworks to static libraries), and more things have the correct default (only links static libraries to  executables and `transitivelyLinkDependencies` now works with them).

### DIFF
--- a/Sources/ProjectSpec/Linkage.swift
+++ b/Sources/ProjectSpec/Linkage.swift
@@ -1,0 +1,41 @@
+import Foundation
+
+public enum Linkage {
+    case dynamic
+    case `static`
+    case none
+}
+
+extension Target {
+    
+    public var defaultLinkage: Linkage {
+        switch type {
+        case .none,
+             .appExtension,
+             .application,
+             .bundle,
+             .commandLineTool,
+             .messagesApplication,
+             .messagesExtension,
+             .ocUnitTestBundle,
+             .stickerPack,
+             .tvExtension,
+             .uiTestBundle,
+             .unitTestBundle,
+             .watchApp,
+             .watchExtension,
+             .watch2App,
+             .watch2Extension,
+             .xcodeExtension,
+             .xpcService:
+            return .none
+        case .framework:
+            // TODO: This should check `MACH_O_TYPE` in case this is a "Static Framework"
+            return .dynamic
+        case .dynamicLibrary:
+            return .dynamic
+        case .staticLibrary:
+            return .static
+        }
+    }
+}
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -779,7 +779,7 @@ public class PBXProjGenerator {
                     }
                 case .target:
                     if let dependencyTarget = project.getTarget(dependency.reference) {
-                        if isTopLevel || (dependency.embed == nil && dependencyTarget.type != .staticLibrary) {
+                        if isTopLevel || dependency.embed == nil {
                             dependencies[dependency.reference] = dependency
                             if !dependencyTarget.shouldEmbedDependencies {
                                 // traverse target's dependencies if it doesn't embed them itself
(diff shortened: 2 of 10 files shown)


brentleyjones:
@Keith Can you make sure this version doesn't break anything on your end? I assume it doesn't since you have `transitivelyLinkDependencies` turned off and you are linking manually.

keith:
Looks like this branch compared to master have no diff for us on our pbxproj. Also it doesn't seem to have any performance impact.

Sources/ProjectSpec/Linkage.swift
```diff
@@ -0,0 +1,41 @@
...
+             .unitTestBundle,
+             .watchApp,
+             .watchExtension,
+             .watch2App,
+             .watch2Extension,
+             .xcodeExtension,
+             .xpcService:
+            return .none
+        case .framework:
+            // TODO: This should check `MACH_O_TYPE` in case this is a "Static Framework"
```
yonaskolb: I'd like to get this in. This logic should then probably be moved into XcodeGenKit

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -435,7 +435,10 @@ public class PBXProjGenerator {
 
                 dependencies.append(targetDependency.reference)
 
-                if (dependencyTarget.type.isLibrary || dependencyTarget.type.isFramework) && dependency.link {
+                let dependecyLinkage = dependencyTarget.defaultLinkage
+                let shouldLink = (dependecyLinkage == .dynamic && target.type != .staticLibrary)
+                    || (dependecyLinkage == .static && target.type.isExecutable)
+                if dependency.link ?? shouldLink {
```
yonaskolb: Once you're checking for build settings, then that code shouldn't run if they have manually specified link setting, to aid in performance

Tests/XcodeGenKitTests/ProjectGeneratorTests.swift
```diff
@@ -274,10 +274,15 @@ class ProjectGeneratorTests: XCTestCase {
                 //     - framework: FrameworkA.framework
                 //     - framework: FrameworkB.framework
                 //       embed: false
+                // iOSFrameworkZ:
+                //   dependencies: []
```
yonaskolb: What's missing for this to be uncommented?

brentleyjones: This whole comment is explaining the targets below. I could remove the whole block, it just made it easier to reason about all of the dependencies.

Sources/XcodeGenKit/PBXProjGenerator.swift
brentleyjones: Good point. I’ll adjust.

Sources/ProjectSpec/Linkage.swift
brentleyjones: The only reason I didn’t is because just like #350 is all about static libraries there is another class of changes needed for static frameworks. I didn’t want to put part of the static frameworks change in with this, especially since it wasn’t the purpose. The reason for a TODO is because I plan on doing that very soon (as soon as #350 is done, so this week).

brentleyjones: I think the type should stay in ProjectSpec since you will be able to specify it as a `preferredLinkage` (as described in #343). I can move the extension if desired though?

yonaskolb: Yeah the type can stay here, though the logic in the switch could go into XcodeGen. It would have to move later anyway if you check for `MACH_O_TYPE` within the switch.

Tests/XcodeGenKitTests/ProjectGeneratorTests.swift
yonaskolb: Yeah ok. You can also generate a spec from a yaml string if that's easier to declare 👍

Tests/XcodeGenKitTests/ProjectGeneratorTests.swift
```diff
@@ -274,10 +274,15 @@ class ProjectGeneratorTests: XCTestCase {
                 //     - framework: FrameworkA.framework
                 //     - framework: FrameworkB.framework
                 //       embed: false
+                // iOSFrameworkZ:
```
yonaskolb: This test can probably me moved into a seperate `testDependencies` test function later

PR #353, opened 2018-07-20
Add `requiresObjCLinking` to `Target`

brentleyjones:
Child of #350. Addresses 2.ii.

### DIFF
--- a/Docs/ProjectSpec.md
+++ b/Docs/ProjectSpec.md
@@ -164,9 +164,11 @@ Settings are merged in the following order: groups, base, configs.
 - [ ] **settings**: **[Settings](#settings)** - Target specific build settings. Default platform and product type settings will be applied first before any custom settings defined here. Other context dependant settings will be set automatically as well:
 	- `INFOPLIST_FILE`: If it doesn't exist your sources will be searched for `Info.plist` files and the first one found will be used for this setting
 	- `FRAMEWORK_SEARCH_PATHS`: If carthage dependencies are used, the platform build path will be added to this setting
+	- `OTHER_LDFLAGS`:  See `requiresObjCLinking` below
 - [ ] **dependencies**: **[[Dependency](#dependency)]** - Dependencies for the target
 - [ ] **templates**: **[String]** - A list of target templates that will be merged in order
 - [ ] **transitivelyLinkDependencies**: **Bool** - If this is not specified the value from the project set in [Options](#options)`.transitivelyLinkDependencies` will be used.
+- [ ] **requiresObjCLinking**: **Bool** - If this is `true` any targets that link to this target will have `-ObjC` added to their `OTHER_LDFLAGS`. This is required if a static library has any catagories or extensions on Objective-C code. See [this guide](https://pewpewthespells.com/blog/objc_linker_flags.html#objc) for more details. Defaults to `true` if `type` is `library.static`. If you are 100% sure you don't have catagories or extensions on Objective-C code (pure Swift with no use of Foundation/UIKit) you can set this to `false`, otherwise it's best to leave it alone.
 - [ ] **prebuildScripts**: **[[Build Script](#build-script)]** - Build scripts that run *before* any other build phases
 - [ ] **postbuildScripts**: **[[Build Script](#build-script)]** - Build scripts that run *after* any other build phases
 - [ ] **buildRules**: **[[Build Rule](#build-rule)]** - Custom build rules
--- a/Sources/ProjectSpec/Target.swift
+++ b/Sources/ProjectSpec/Target.swift
@@ -29,6 +29,7 @@ public struct Target: ProjectTarget {
     public var sources: [TargetSource]
     public var dependencies: [Dependency]
     public var transitivelyLinkDependencies: Bool?
+    public var requiresObjCLinking: Bool?
     public var prebuildScripts: [BuildScript]
     public var postbuildScripts: [BuildScript]
     public var buildRules: [BuildRule]
@@ -61,6 +62,7 @@ public struct Target: ProjectTarget {
         sources: [TargetSource] = [],
         dependencies: [Dependency] = [],
         transitivelyLinkDependencies: Bool? = nil,
+        requiresObjCLinking: Bool? = nil,
         prebuildScripts: [BuildScript] = [],
         postbuildScripts: [BuildScript] = [],
         buildRules: [BuildRule] = [],
@@ -77,6 +79,7 @@ public struct Target: ProjectTarget {
         self.sources = sources
         self.dependencies = dependencies
         self.transitivelyLinkDependencies = transitivelyLinkDependencies
+        self.requiresObjCLinking = requiresObjCLinking
         self.prebuildScripts = prebuildScripts
         self.postbuildScripts = postbuildScripts
         self.buildRules = buildRules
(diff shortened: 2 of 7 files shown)


Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -680,6 +611,98 @@ public class PBXProjGenerator {
...
+                
+                return dependencyTarget.requiresObjCLinking
+                    ?? dependencySourceFiles().contains { ["m", "mm", "swift"].contains($0.path.extension) }
+            }
+            if anyDependencyRequiresObjCLinking {
+                let otherLinkingFlags = "OTHER_LDFLAGS"
+                if let string = buildSettings[otherLinkingFlags] as? String {
+                    buildSettings["OTHER_LDFLAGS"] = "\(string) -ObjC"
+                } else {
+                    buildSettings["OTHER_LDFLAGS"] = "$(inherited) -ObjC"
```
brentleyjones: This is the new code in here. Lots of red and green since this section had to be moved below `targetDependencies` generation

Sources/ProjectSpec/Target.swift
```diff
@@ -159,6 +162,63 @@ extension Target {
         merged["targets"] = crossPlatformTargets
         return merged
     }
+    
+    public var shouldEmbedDependencies: Bool {
+        return type.isApp || type.isTest
```
yonaskolb: This could be an extension on `PBXProjectType` like the getters it's using

Sources/ProjectSpec/Target.swift
```diff
@@ -159,6 +162,63 @@ extension Target {
...
+    
+    public func generateEffectiveDependencies(project: Project) -> [Dependency] {
+        if transitivelyLinkDependencies ?? project.options.transitivelyLinkDependencies {
+            return getAllDependenciesPlusTransitiveNeedingEmbedding(project: project)
+        } else {
+            return dependencies
+        }
+    }
+    
+    private func getAllDependenciesPlusTransitiveNeedingEmbedding(project: Project) -> [Dependency] {
```
yonaskolb: I'd rather keep all this code out of the `ProjectSpec` module. Some people use `ProjectSpec` as a seperate library without XcodeGen's logic.
But mostly this also removes the ability to add caching across the project as there is no controller for this logic

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -680,6 +611,98 @@ public class PBXProjGenerator {
...
+                        let dependencyTarget = project.getTarget(dependency.reference),
+                        dependencyTarget.type == .application {
+                        buildSettings["TEST_TARGET_NAME"] = dependencyTarget.name
+                        break
+                    }
+                }
+            }
+            
+            // objc linkage
+            let anyDependencyRequiresObjCLinking = targetDependencies.contains { dependency in
```
yonaskolb: This could be quite performance intensive

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -680,6 +611,98 @@ public class PBXProjGenerator {
...
+                    // This is duplicative and should be cached
+                    return (try? sourceGenerator.getAllSourceFiles(sources: dependencyTarget.sources)) ?? []
+                }
+                
+                return dependencyTarget.requiresObjCLinking
+                    ?? dependencySourceFiles().contains { ["m", "mm", "swift"].contains($0.path.extension) }
+            }
+            if anyDependencyRequiresObjCLinking {
+                let otherLinkingFlags = "OTHER_LDFLAGS"
+                if let string = buildSettings[otherLinkingFlags] as? String {
```
yonaskolb: The setting may also be a `[String]`. Perhaps a convenience somewhere that appends an item to a list style setting using `$(inherited)` would be good

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -680,6 +611,98 @@ public class PBXProjGenerator {
...
+            
+            // objc linkage
+            let anyDependencyRequiresObjCLinking = targetDependencies.contains { dependency in
+                guard dependency.type == .target else { return false }
+                guard let dependencyTarget = project.getTarget(dependency.reference) else { return false }
+                guard targetsLinked.contains(dependencyTarget.name) else { return false }
+                
+                func dependencySourceFiles() -> [SourceFile] {
+                    // This is duplicative and should be cached
+                    return (try? sourceGenerator.getAllSourceFiles(sources: dependencyTarget.sources)) ?? []
```
yonaskolb: This would generate duplicates wouldn't it?

yonaskolb:
I'm not convinced all this complexity and performance hit is worth it. Is there a big downside to either just adding `OTHER_LDFLAGS: "$(inherited) -ObjC"` to every target by default, or at least to targets that have `library.static` target dependencies?

Docs/ProjectSpec.md
```diff
@@ -162,8 +162,10 @@ Settings are merged in the following order: groups, base, configs.
 - [ ] **settings**: **[Settings](#settings)** - Target specific build settings. Default platform and product type settings will be applied first before any custom settings defined here. Other context dependant settings will be set automatically as well:
 	- `INFOPLIST_FILE`: If it doesn't exist your sources will be searched for `Info.plist` files and the first one found will be used for this setting
 	- `FRAMEWORK_SEARCH_PATHS`: If carthage dependencies are used, the platform build path will be added to this setting
+	- `OTHER_LDFLAGS`:  See `requiresObjCLinking` below
 - [ ] **dependencies**: **[[Dependency](#dependency)]** - Dependencies for the target
 - [ ] **transitivelyLinkDependencies**: **Bool** - If this is not specified the value from the project set in [Options](#options)`.transitivelyLinkDependencies` will be used.
+- [ ] **requiresObjCLinking**: **Bool** - If this is `true` any targets that link to this target will have `-ObjC` added to their `OTHER_LDFLAGS`. Defaults to `true` if `type` is `library.static` and [Sources](#sources) contain any `.swift`, `.m`, or `.mm` files.
```
yonaskolb: People would really have to have a deep understanding of what this means

brentleyjones:
Great feedback. I’ll adjust it to not look at the sources and just default the value to `true` for static libraries. I’ll clean up the documentation so it’s bettter understandable by the average person (and why you would want to adjust the default). I’ll also move this back into their proper places.

I want to keep the property though, since I see it being added to the `carthage` and `framework` dependency types for static frameworks.

Ohh, and to answer your question about just putting this on targets: there is a very real binary size hit you take if you use this option without he need for it. I’ll make sure the documentation reflects this.

Sources/XcodeGenKit/PBXProjGenerator.swift
brentleyjones: I changed it to follow the pattern. I didn't add a convenience though.

Sources/XcodeGenKit/PBXProjGenerator.swift
brentleyjones: All searching of sources was removed, and it now just sets a flag once as soon as `anyDependencyRequiresObjCLinking` would be true. Performance shouldn't be an issue anymore.

Sources/ProjectSpec/Target.swift
brentleyjones: Removed that commit altogether, it's back in `XcodeGenKit`.

brentleyjones:
@yonaskolb Addressed your feedback. Per the other PR's comments, I'll add new commits instead of amending next time.

PR #354, opened 2018-07-22
Add aggregate targets

### LINKED ISSUE #47: Support PBXAggregateTarget
It seems that https://github.com/swift-xcode/xcodeproj supports PBXAggregateTarget (?), but XcodeGen doesn't.
Priority would be low though, I often use the kind of target in framework projects to invoke archive script.

yonaskolb:
Resolves #47

### DIFF
--- a/Sources/ProjectSpec/AggregateTarget.swift
+++ b/Sources/ProjectSpec/AggregateTarget.swift
@@ -32,7 +32,7 @@ public struct AggregateTarget: ProjectTarget {
 extension AggregateTarget: CustomStringConvertible {
 
     public var description: String {
-        return "\(name): \(targets.joined(separator: ", "))"
+        return "\(name)\(targets.isEmpty ? "" : ": \(targets.joined(separator: ", "))")"
     }
 }
 
--- a/Sources/ProjectSpec/AggregateTarget.swift
+++ b/Sources/ProjectSpec/AggregateTarget.swift
@@ -1,13 +1,14 @@
 import Foundation
 import JSONUtilities
 
-public struct AggregateTarget: ProjectTarget, Equatable {
+public struct AggregateTarget: ProjectTarget {
     public var name: String
     public var targets: [String]
     public var settings: Settings
     public var buildScripts: [BuildScript]
     public var configFiles: [String: String]
     public var scheme: TargetScheme?
+    public var attributes: [String: Any]
 
     public init(
         name: String,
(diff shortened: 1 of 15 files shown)


yonaskolb:
There is still something unresolved here as I can't work out why the aggregate target is being removed when the test fixture project is opened in Xcode

brentleyjones:
@yonaskolb This doesn't seem to work. I added the following to my `project.yml`:

```yaml
aggregateTargets:
  Carthage Verify:
    buildScripts:
      - name: Carthage Verify
        script: cd $SRCROOT && $SRCROOT/carthage-verify.sh
```

And it says in the output:

```console
  Aggregate Targets:
    Carthage Verify: 
```

But there isn't any target when I check in Xcode.

Sources/ProjectSpec/AggregateTarget.swift
```diff
@@ -0,0 +1,63 @@
...
+        self.buildScripts = buildScripts
+        self.scheme = scheme
+        self.attributes = attributes
+    }
+}
+
+extension AggregateTarget: CustomStringConvertible {
+
+    public var description: String {
+        return "\(name): \(targets.joined(separator: ", "))"
```
brentleyjones: When `targets` is empty this shouldn't include `:`. It's valid to not have any targets if it's only used to run a script, like `carthage-verify.sh`.

yonaskolb: Yeah at the moment `targets` is a required field. Are there cases where you wouldn't actually aggregate targets? Isn't that the point of them? I haven't really used them before so I'm not sure

yonaskolb:
Yeah that's the issue I mentioned above. If you look at the project diff the aggregate target does get generated, but as soon as you open the project in Xcode it removes it, and I have no idea why...

Sources/ProjectSpec/AggregateTarget.swift
brentleyjones: Yes. They are named weird in that way. It's also a common way to just run scripts. We use it to `carthage-verify.sh` before all our other targets so we get a fast failure and don't have to run the same script for every target in our project.

yonaskolb: Ok cool. A normal empty target would achieve the same thing right, this is just better suited to it?

brentleyjones:
Here is what Xcode gives me when I create the same thing by hand:

```
/* Begin PBXAggregateTarget section */
		3BF4E56421061F4000C16C69 /* Carthage Verify */ = {
			isa = PBXAggregateTarget;
			buildConfigurationList = 3BF4E56521061F4100C16C69 /* Build configuration list for PBXAggregateTarget "Carthage Verify" */;
			buildPhases = (
			);
			dependencies = (
			);
			name = "Carthage Verify";
			productName = "Carthage Verify";
		};
/* End PBXAggregateTarget section */
```

And in the `targets` section of the `project` object:

```
3BF4E56421061F4000C16C69 /* Carthage Verify */,
```

Seems you are missing that part in the `project` object.

Sources/ProjectSpec/AggregateTarget.swift
brentleyjones: An empty target fails to build I believe, though I haven't tried that hard since I've only really done it this way (which I believe is the recommended way).

PR #356, opened 2018-07-23
Change group sorting

### LINKED ISSUE #292: Specify the structure not only alphabetically but also using priorities for folders / files
If the project is named starting letter `P` the project is always generated as below. The Product reference folder is placed on top of the file structure. Is there any option to mark the folders with priority of order ? Also If I do have README.md file in the project it is sometimes placed to the top sometimes not. Is it possible to place it always on the top?  
```
Test.xcodeproj
Product
Test
---- Sources
-------- Extension
-------- Scenes
------------ Feature.swift
---- Resources
---- SupportingFiles
```

`project.yml`
```
name: TemplateTest
options:
  bundleIdPrefix: com.test
targets:
  TemplateTest:
    type: application
    platform: iOS
    sources: [TemplateTest]
```

It is probably for another issue but if I specify `excludes` this way:
```
name: TemplateTest
options:
  bundleIdPrefix: com.test
targets:
  TemplateTest:
    type: application
    platform: iOS
    sources: [TemplateTest]
    excludes: 
       - "README.md" 
```
it basically never gets excluded.

yonaskolb:
This tweaks the sorting of groups and files in 2 ways
- Groups like `Frameworks` and `Products` are added to the end of the top level groups
- Groups are sorted after other files by default
- Added new `options.groupSortPosition`

Resolves #292


yonaskolb:
@brentleyjones I'm in agreement with you, but after looking at other projects, and seeing what Xcode does by default (also after 'sort by types`), it seems this is more common. If it's an issue we can add an `option`

yonaskolb:
I've added `options.groupSortPosition` of either `top`, `bottom` or `none`

PR #360, opened 2018-07-25
Add names to non-generic Copy Files phases

brentleyjones:
They've always confused me nameless, and now with #345 there can be even more generically named phases.


keith:
Nice, been meaning to do this 👍

Issue #362, opened 2018-07-29
CoreData model versioning configuration


rpassis:
Hi all,

Just wondering if it is possible to select a versioned CoreData model "current version" option using XcodeGen? I had a look at the documentation but haven't been able to find it.

Thanks for developing such a great tool!

Rog

yonaskolb:
Hi @rpassis, there is currently no way to manually select a core data model version, however some work was done recently to automatically select the newest one by sorting them https://github.com/yonaskolb/XcodeGen/pull/341 
It will be in the next release. Would that solve your issue?

rpassis:
Hey @yonaskolb thanks for the quick response! I was just having a look at that PR. I'll build from master to test it and update here but it looks like that's likely to do the trick!

FWIW the filename/path fix seems a bit fragile and a possible alternative would be to check if there is a `.xccurrentversion` file inside a versioned model, and set the version based on that. Here's a related issue that was tackled by Xcodeproj https://github.com/CocoaPods/Xcodeproj/issues/81

rpassis:
Yup that file sorting change has fixed the issue, thanks again!

yonaskolb:
Great! I'll reopen this for tracking the proper fix, if anyone wants to work on it

rpassis:
I'll see if I can tackle this one myself, watch this space!

yonaskolb:
Fantastic! 😄👍

brentleyjones:
Fixed by #364.

PR #364, opened 2018-07-30
Improvements to determining the current version of a versioned CoreData model

rpassis:
The current implementation for generating versioned CoreData models and selecting the current active version uses alphabetical sorting. This can work for most simple projects but is likely to break for less conventional setups.

A better solution is to parse the `.xccurrentversion` file which is included inside the CoreData versioned model. This file is in plist format and is created automatically by Xcode.

This is a fix for issue #362

### DIFF
--- a/Sources/XcodeGenKit/SourceGenerator.swift
+++ b/Sources/XcodeGenKit/SourceGenerator.swift
@@ -115,7 +115,43 @@ class SourceGenerator {
             }
             let lastKnownFileType = lastKnownFileType ?? PBXFileReference.fileType(path: path)
 
-            guard path.extension == "xcdatamodeld" else {
+            if path.extension == "xcdatamodeld" {
+                let versionedModels = (try? path.children()) ?? []
+
+                // Sort the versions alphabetically
+                let sortedPaths = versionedModels
+                    .filter { $0.extension == "xcdatamodel" }
+                    .sorted { $0.string.localizedStandardCompare($1.string) == .orderedAscending }
+
+                let modelFileReference =
+                    sortedPaths.map { path in
+                        createObject(
+                            id: path.byRemovingBase(path: project.basePath).string,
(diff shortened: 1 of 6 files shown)


rpassis:
One thing to note is that the alphabetical order implementation here will break even for very basic use cases. A very simple versioned model could look like this

```
Model.xcdatamodel
Model 1.xcdatamodel
Model 2.xcdatamodel
```

When applying the existing sort logic and taking the last item as the current CoreData model you will end up with 

```
Model 1.xcdatamodel
Model 2.xcdatamodel
Model.xcdatamodel
```

And Model.xcdatamodel incorrectly set as the current version.

I'd suggest we should just remove the alphabetical fallback completely but I can't be certain that all CoreData models do include a `.xccurrentversion` file indicating what the current version is (based on my own tests it does but this doesn't seem to be documented anywhere).

Sources/XcodeGenKit/SourceGenerator.swift
```diff
@@ -115,31 +115,7 @@ class SourceGenerator {
...
-                    currentVersion: modelFileReference.last?.reference,
-                    path: fileReferencePath.string,
-                    sourceTree: sourceTree,
-                    versionGroupType: "wrapper.xcdatamodel",
-                    children: modelFileReference.map { $0.reference }
-                ))
-                fileReferencesByPath[fileReferenceKey] = versionGroup
-                return versionGroup
-            } else {
+            guard path.extension == "xcdatamodeld" else {
```
yonaskolb: Can we keep the `if` as opposed to the `guard`? Not being a `xcdatamodeld` is not a failing case, and we may handle other filetypes differently in the future as well. 
Additionally it will make these diffs easier to read.

rpassis:
@yonaskolb thanks for the feedback! I've addressed your comment and pushed a new commit.

PR #365, opened 2018-07-30
 Automatically copy public headers for static libraries

brentleyjones:
Child of #350. Addresses 1 (but not 1.i or 1.ii).

Depends on #346. Can review just the one "
Automatically copy public headers for static libraries" commit. Per previous feedback, dependent PRs will be handled better in the future 😸.


brentleyjones:
Re-reading my original post in #350, I need to make sure this works for when the build phase is original specified as `headers` as well. I'll amend the original commit soon.

brentleyjones:
@yonaskolb Not sure what you are referring to. If it's "Per previous feedback, dependent PRs will be handled better in the future" then PR's like #345 will be pushed to the main repo and dependent PRs will target them (per your advice). If it's referring to anything else, I don't follow 😺.

yonaskolb:
Sorry, was referring to what `amend original commit` meant. Thought you might of meant you had to change the original commits this PR is based off. Or are the required change just in this PR?

brentleyjones:
Ohh. I was just saying that I was going to do a small amend versus a new commit here since there wasn’t a review yet. All is good now!

PR #366, opened 2018-08-01
Automatically copy Swift.h header for static libraries

### LINKED ISSUE #350: Better static library support

brentleyjones:
Resolves #350, addressing 2.i.

`SWIFT_OBJC_INTERFACE_HEADER_NAME` is set by Xcode correctly depending on if you have Swift sources or not. We can't see that though and always get `nil` for it unless it's overridden. Because of this we have to check our Compile Source files for any Swift files. We only do so after ensuring that we are looking at a static library that hasn't set `SWIFT_OBJC_INTERFACE_HEADER_NAME` to `""`, for the best performance.

I initially went the route of having these files be copied with a Copy Files phase, but it had some ugliness:
- The files used "${CONFIGURATION_TEMP_DIR}" in their path, which means you couldn't open them and they were always red.
- The path wasn't 100% correct, because some variables could be changed by Xcode and we wouldn't know where the file would be located.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 - Moved `Frameworks` and `Products` top level groups to bottom [356](https://github.com/yonaskolb/XcodeGen/pull/356) @yonaskolb
 - `modulemap` files are automatically copied to the products directory for static library targets [346](https://github.com/yonaskolb/XcodeGen/pull/346) @brentleyjones
 - Public header files are automatically copied to the products directory for static library targets [365](https://github.com/yonaskolb/XcodeGen/pull/365) @brentleyjones
+- Swift Objective-C Interface Header files are automatically copied to the products directory for static library targets [366](https://github.com/yonaskolb/XcodeGen/pull/366) @brentleyjones
 
 #### Internal
 - Moved brew formula to homebrew core
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -629,6 +629,30 @@ public class PBXProjGenerator {
             buildPhases.append(resourcesBuildPhase.reference)
         }
         
+        
+        let buildSettings = project.getCombinedBuildSettings(basePath: project.basePath, target: target, config: project.configs[0])
+        let swiftObjCInterfaceHeader = buildSettings["SWIFT_OBJC_INTERFACE_HEADER_NAME"] as? String
+        
+        if target.type == .staticLibrary
+            && swiftObjCInterfaceHeader != ""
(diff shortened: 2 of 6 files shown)


keith:
No diff, thanks!

brentleyjones:
@yonaskolb This is the last thing needed to complete #350. Let me know if there is anything I can do to get this in 😺

PR #367, opened 2018-08-02
Add additional params to archive action

sxua:
This change adds an ability to customize archive name and whether it should be revealed in organizer or not. Basically, one can set options displayed on an image below via `project.yml`

screen shot 2018-08-02 at 3 24 53 pm

### DIFF
--- a/Sources/ProjectSpec/Scheme.swift
+++ b/Sources/ProjectSpec/Scheme.swift
@@ -150,14 +150,20 @@ public struct Scheme: Equatable {
 
     public struct Archive: BuildAction {
         public var config: String?
+        public var customArchiveName: String?
+        public var revealArchiveInOrganizer: Bool
         public var preActions: [ExecutionAction]
         public var postActions: [ExecutionAction]
         public init(
             config: String,
+            customArchiveName: String? = nil,
+            revealArchiveInOrganizer: Bool = true,
             preActions: [ExecutionAction] = [],
             postActions: [ExecutionAction] = []
         ) {
             self.config = config
+            self.customArchiveName = customArchiveName
+            self.revealArchiveInOrganizer = revealArchiveInOrganizer
             self.preActions = preActions
             self.postActions = postActions
         }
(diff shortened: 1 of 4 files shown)


Sources/ProjectSpec/Scheme.swift
```diff
@@ -150,14 +150,20 @@ public struct Scheme: Equatable {
 
     public struct Archive: BuildAction {
         public var config: String?
+        public var revealArchiveInOrganizer: Bool
+        public var customArchiveName: String?
```
brentleyjones: The archive name is shown before `revealArchiveInOrganizer` in the dialog. Can you swap these (here and below) to match that?

Sources/XcodeGenKit/ProjectGenerator.swift
```diff
@@ -133,7 +133,8 @@ public class ProjectGenerator {
 
         let archiveAction = XCScheme.ArchiveAction(
             buildConfiguration: scheme.archive?.config ?? defaultReleaseConfig.name,
-            revealArchiveInOrganizer: true,
+            revealArchiveInOrganizer: scheme.archive?.revealArchiveInOrganizer ?? true,
+            customArchiveName: scheme.archive?.customArchiveName ?? nil,
```
yonaskolb: Nitpick - the `?? nil` here is redundant

yonaskolb:
Nice, thanks @sxua!

yonaskolb:
Could you add the new options in https://github.com/yonaskolb/XcodeGen/blob/master/Docs/ProjectSpec.md#scheme under a new `Archive` section, and also add a changelog entry?

Sources/ProjectSpec/Scheme.swift
sxua: @brentleyjones This is the order of an xcproj's `ArchiveAction` initializer, but I can change it here for sure.

sxua:
@yonaskolb Good catch, thanks!

sxua:
@yonaskolb Should be fine now

Issue #372, opened 2018-08-06
Framework dependency always embedded


danicoello:
I'm having this dependency settings on my target section of project.yml file:

```
dependencies:
      - framework: myFramwork.framework
        embed: false
```

This produces that the "Embedded Binaries Section" contains "My framework" added when this should not happens. If I set "embed" to "true" the same framework is added twice on the same section. In both scenarios the "Linked Frameworks and Libraries" contains the link to framework just one time.

I only want that the framework is linked but not embedded

It appears that framework is always being added in addtion to my settings. What could be happening?

brentleyjones:
Can you post a more complete project.yml (made up names are fine). I’ll take a look at this.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

danicoello:
This is the project.yml that I am using.

```
name: MyApp
options:
  xcodeVersion: "9.4.1"
  bundleIdPrefix: "com.myapp"
  deploymentTarget: 
    iOS: "9.3"
configs:
  QA Debug: debug
  QA Release: release
  PROD Debug: debug
  PROD Release: release
settings:
  configs:
    QA Debug:
      DISPLAY_NAME_CUSTOM: "My App QA"
      PRODUCT_BUNDLE_IDENTIFIER: "com.myapp.MyApp-qa"
      ENVIROMENT: "QA"
    QA Release:
      DISPLAY_NAME_CUSTOM: "My App QA"
      PRODUCT_BUNDLE_IDENTIFIER: "com.myapp.MyApp-qa"
      ENVIROMENT: "QA"
    PROD Debug:
      DISPLAY_NAME_CUSTOM: "My App"
      PRODUCT_BUNDLE_IDENTIFIER: "com.myapp.MyApp"
      ENVIROMENT: "PROD"
    PROD Release:
      DISPLAY_NAME_CUSTOM: "My App"
      PRODUCT_BUNDLE_IDENTIFIER: "com.myapp.MyApp"
      ENVIROMENT: "PROD"
targets:
  MyApp:
    type: application
    platform: iOS
    sources: MyApp
    settings:
      INFOPLIST_FILE: MyApp/Resources/MyApp.plist
      SWIFT_VERSION: "3.3"
      SWIFT_OBJC_BRIDGING_HEADER: MyApp/Resources/MyApp-Bridging-Header.h
[...]

PR #373, opened 2018-08-08
Add to `FRAMEWORK_SEARCH_PATHS` for `framework` dependencies

brentleyjones:
Addresses an issue discussed in #372.


Tests/Fixtures/TestProject/Project.xcodeproj/project.pbxproj
```diff
@@ -26,6 +26,8 @@
 		BF_130062884703 /* Alamofire.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = FR_257516580010 /* Alamofire.framework */; };
 		BF_138356261076 /* InterfaceController.swift in Sources */ = {isa = PBXBuildFile; fileRef = FR_363921640403 /* InterfaceController.swift */; };
 		BF_144945303317 /* Alamofire.framework in Embed Frameworks */ = {isa = PBXBuildFile; fileRef = FR_410645050443 /* Alamofire.framework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
+		BF_151769972761 /* SomeFramework.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = FR_654302359039 /* SomeFramework.framework */; };
+		"BF_151769972761-1" /* SomeFramework.framework in Embed Frameworks */ = {isa = PBXBuildFile; fileRef = FR_654302359039 /* SomeFramework.framework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
```
brentleyjones: @yonaskolb Do you know what causes this?

Tests/Fixtures/TestProject/project.yml
```diff
@@ -61,6 +61,7 @@ targets:
       - carthage: Alamofire
       - target: App_watchOS
       - target: iMessageApp
+      - framework: Vendor/SomeFramework.framework
```
brentleyjones: This doesn't exist. It was just to make sure the framework path worked. I could pass `link: false` and `embed: false` to make sure it doesn't affect the build. Not sure if the project has to compile.

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -798,16 +806,24 @@ public class PBXProjGenerator {
             }
             
             // set Carthage search paths
+            let configFrameworkBuildPaths: [String]
             if !carthageDependencies.isEmpty {
-                let frameworkSearchPaths = "FRAMEWORK_SEARCH_PATHS"
                 let carthagePlatformBuildPath = "$(PROJECT_DIR)/" + getCarthageBuildPath(platform: target.platform)
+                configFrameworkBuildPaths = [carthagePlatformBuildPath] + Array(frameworkBuildPaths).sorted()
```
brentleyjones: I thought it was best that Carthage path was first then the rest alphabetically. If it should be strictly alphabetic let me know.

Tests/Fixtures/TestProject/project.yml
yonaskolb: Yeah, I've been trying to keep the project buildable. I've actually been thinking of adding some `xcodebuild` commands to the tests to validate that everything still builds

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -538,6 +539,13 @@ public class PBXProjGenerator {
                     )
                     copyFrameworksReferences.append(embedFile.reference)
                 }
+                
+                let buildPath = Path(dependency.reference)
+                    .byRemovingBase(path: project.basePath)
+                    .components
```
yonaskolb: You should be able to use `parent.string` instead of `components.dropLast()`.joined(separator: "/")`

Sources/XcodeGenKit/PBXProjGenerator.swift
yonaskolb: All good 👍

Tests/Fixtures/TestProject/Project.xcodeproj/project.pbxproj
yonaskolb: Hmm I don't, interesting. Looks like an issue with `.framework` dependency embedding. We should get it fixed.
As an explanation for the id, if a duplicate is generated it will have a `-number` appended. The addition of a `-` also surrounds it quotes which as a plus, makes it standout.

brentleyjones:
@yonaskolb I addressed the comments and made sure it built for the simulator.

PR #383, opened 2018-08-20
Fix aggregate target dependencies with `transitivelyLinkDependencies`

brentleyjones:
Fixes #381 to work with the `transitivelyLinkDependencies` flag.

### DIFF
--- a/Sources/ProjectSpec/Project.swift
+++ b/Sources/ProjectSpec/Project.swift
@@ -14,7 +14,11 @@ public struct Project: BuildSettingsContainer {
         }
     }
 
-    public var aggregateTargets: [AggregateTarget]
+    public var aggregateTargets: [AggregateTarget] {
+        didSet {
+            aggregateTargetsMap = Dictionary(uniqueKeysWithValues: aggregateTargets.map { ($0.name, $0) })
+        }
+    }
 
     public var settings: Settings
     public var settingGroups: [String: Settings]
@@ -25,7 +29,9 @@ public struct Project: BuildSettingsContainer {
     public var fileGroups: [String]
     public var configFiles: [String: String]
     public var include: [String] = []
+    
     private var targetsMap: [String: Target]
+    private var aggregateTargetsMap: [String: AggregateTarget]
 
     public init(
         basePath: Path,
@@ -46,6 +52,7 @@ public struct Project: BuildSettingsContainer {
         self.targets = targets
         targetsMap = Dictionary(uniqueKeysWithValues: self.targets.map { ($0.name, $0) })
         self.aggregateTargets = aggregateTargets
+        aggregateTargetsMap = Dictionary(uniqueKeysWithValues: self.aggregateTargets.map { ($0.name, $0) })
         self.configs = configs
         self.settings = settings
         self.settingGroups = settingGroups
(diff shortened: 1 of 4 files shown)


yonaskolb:
Could you add a changelog entry

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -981,6 +981,9 @@ public class PBXProjGenerator {
                                 queue.append(dependencyTarget)
                             }
                         }
+                    } else {
```
yonaskolb: Would be good if the code showed what was happening as well as the comment. For completion sake could we add a `Project.getAggregateTarget(_ name: String) -> AggregateTarget` function and use it here
```swift
} else if project.getAggregateTarget(dependency.reference) != nil {
```

yonaskolb:
Good catch 👍

brentleyjones:
Done and done :smile:.

Sources/ProjectSpec/Project.swift
```diff
@@ -25,7 +25,9 @@ public struct Project: BuildSettingsContainer {
     public var fileGroups: [String]
     public var configFiles: [String: String]
     public var include: [String] = []
+    
     private var targetsMap: [String: Target]
+    private var aggregateTargetsMap: [String: AggregateTarget]
```
yonaskolb: One missing map creation in the `aggregateTargets` setter

brentleyjones: Oops. Fixed.

yonaskolb:
Actually, looks like `getAllCarthageDependencies` will need updating too...

brentleyjones:
I hadn't considered that. I'll fix that quick. Related, in my Static Framework work, I'll be consolidating those and doing the caching work we talked about.

yonaskolb:
Awesome. Yeah those 2 could be combined and optimised by only running over the graph once 👍

brentleyjones:
Updated for Carthage dependencies as well.

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -910,37 +910,45 @@ public class PBXProjGenerator {
...
-
-                switch dependency.type {
-                case .carthage:
-                    frameworks[dependency.reference] = dependency
-                case .target:
-                    if let target = project.getTarget(dependency.reference) {
-                        queue.append(target)
+            } else if let aggregateTarget = projectTarget as? AggregateTarget {
+                for dependencyName in aggregateTarget.targets {
+                    if let projectTarget = project.getProjectTarget(dependencyName) {
```
yonaskolb: This highlights a question. Should `getAllDependenciesPlusTransitiveNeedingEmbedding` be checking `aggregateTarget.targets` as well?

brentleyjones: I went down that route at first. It wouldn't since it's the same as `shouldEmbedDependencies` in my mind. An aggregate target will be merging it's targets, they shouldn't propagate up, at least automatically. If that's the desire, it can still be done manually.

In the new algorithm this will be expressible explicitly.

yonaskolb: 👍

yonaskolb:
Released 👍

Issue #389, opened 2018-08-31
Does not generate knownRegions if there are no resources in the project


Saik0s:
When I am trying to run R.swift in the generated project without resources, it gives me an error 
```
error: [R.swift] Project file at 'file:///Users/saik0s/Documents/Projects/modular/Projects/Profile/Profile.xcodeproj/' could not be parsed, is this a valid Xcode project file ending in *.xcodeproj?
The operation couldn’t be completed. (XcodeEdit.AllObjectsError error 0.)
```
After some investigation, I found that the problem is in `knownRegions` parameter. It is not generated when there are no localized resources. But project generated by Xcode sets this parameter to 
```
knownRegions = (
	en,
);
```
by default even without resources.

yonaskolb:
Thanks for the report @Saik0s.
It’s not a required property from Xcode’s point of view, but the library that R.swift uses to parse the project expects it. We can add a default en value though 👍

PR #390, opened 2018-08-31
Set knownRegions to en by default. Fix for #389

Saik0s:
Fix for #389


Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -228,7 +228,9 @@ public class PBXProjGenerator {
             .merged(project.attributes)
             .merged(generateTargetAttributes() ?? [:])
 
-        pbxProject.object.knownRegions = sourceGenerator.knownRegions.sorted()
+        let knownRegions: [String] = sourceGenerator.knownRegions.sorted()
```
yonaskolb: We can drop the explicit type on the knownRegions variable 👍

Issue #392, opened 2018-09-08
Spec validation error when generating Schema for Aggregated Target


vgorloff:
XcodeGen Version: 1.11.1

**Description**:

1. In `project.yml` we can define `aggregateTargets`.
2. Xcode will create automatically schema for that Aggregated Target.
3. We can successfully run that schema.
4. If Xcode can generate schema for that Aggregated Target, this means that XcodeGen can also do so.

![s180908_0914591](https://user-images.githubusercontent.com/3200080/45251527-f9374900-b347-11e8-8720-5ef7c43e6434.png)

![s180908_0916301](https://user-images.githubusercontent.com/3200080/45251529-03f1de00-b348-11e8-9793-af174f8bf8a3.png)

---

**Problem**:

Say I want to `explicitly setup schema` for that Aggregated Target. Doing so will cause following error:

```
$ xcodegen 
📋  Loaded project:
  Name: AggregateTargetsDemo
  Aggregate Targets:
    Verify
⚙️  Generating project...
Spec validation error: Scheme "Verify-XcodeGen" has invalid build target "Verify"
```

Here is a sample `project.yml`:

```
options:
  deploymentTarget:
    macOS: "10.12"
    iOS: "10.0"
    tvOS: "10.0"
  settingPresets: none
  createIntermediateGroups: true
  usesTabs: false
  indentWidth: 3
  tabWidth: 3

attributes:
   ORGANIZATIONNAME: Demo
   LastUpgradeCheck: "1000"

name: AggregateTargetsDemo

aggregateTargets:

  Verify:
    buildScripts:
       - script: |
                 [[ -r ~/.bashrc ]] && . ~/.bashrc
                 echo "**************************** Hello! ****************************"
         name: "Run Script: Verify Sources"
         shell: /bin/bash

# Uncomment below to see an error.

# schemes:

   # Verify-XcodeGen:
   #   build:
   #     targets:
   #       Verify: all
```

Thank you!

---

btw: Can you hint me where to look. Then I probably fix that by myself .) Thanks!

vgorloff:
Thanks. I will try to fix that in upcoming days.

vgorloff:
Seems fixed .)
Please review PR: https://github.com/yonaskolb/XcodeGen/pull/394
Thanks!

yonaskolb:
Released in 1.11.2

PR #394, opened 2018-09-11
Fix for issue https://github.com/yonaskolb/XcodeGen/issues/392.

vgorloff:


Sources/ProjectSpec/SpecValidation.swift
```diff
@@ -135,7 +135,7 @@ extension Project {
 
         for scheme in schemes {
             for buildTarget in scheme.build.targets {
-                if getTarget(buildTarget.target) == nil {
+                if getProjectTarget(buildTarget.target) == nil {
```
vgorloff: Fixes Validation error.

Sources/XcodeGenKit/ProjectGenerator.swift
```diff
@@ -41,15 +41,25 @@ public class ProjectGenerator {
...
-            let target = project.getTarget(buildTarget.target)!
-            let buildableReference = XCScheme.BuildableReference(
-                referencedContainer: "container:\(project.name).xcodeproj",
-                blueprintIdentifier: targetReference.reference,
-                buildableName: target.filename,
-                blueprintName: buildTarget.target
-            )
-
-            return XCScheme.BuildAction.Entry(buildableReference: buildableReference, buildFor: buildTarget.buildTypes)
+            if let target = project.getTarget(buildTarget.target) {
```
vgorloff: Fixes crash on explicitly unwrapped optional.

yonaskolb: Nice. Could we just get the `buildableName` in these `ifs`, and then keep the common code below those?

yonaskolb:
Thank you @vgorloff! One small change, and then we can get this merged 👍

Sources/XcodeGenKit/ProjectGenerator.swift
vgorloff: Ah. I see. Copy paste code. Will fix it today later.

Issue #396, opened 2018-09-13
Directory name containing dots is unexpectedly added to copy bundle resources build phase


toshi0383:
<img width="89" alt="screen shot 2018-09-13 at 11 45 08" src="https://user-images.githubusercontent.com/6007952/45464239-a6e09880-b74a-11e8-812c-37260dc06bb4.png">

brentleyjones:
It's seen as a "file like directory". Wonder if we should instead whitelist that instead of assuming all directories with dots are that.

toshi0383:
I think we should whitelist them. For sure Xcode does whitelist (as we can see icons are different for each resource type), so I guess it makes sense.
It can introduce bugs by forgetting to add any "file like directory" extensions, so I want to make sure we cover all of them in #397 .

toshi0383:
Hi @Lutzifer .
I guess we should create a whitelist for any “file like directories” (which we believe they are), instead of #397 approach.

sashatv126:
Hello  @toshi0383 , ran into this today, i have project with dots in naming. When i add name of project "a.b.c" and same name to source path to my project, folder  a.b.c  is became a reference 
![IMAGE 2023-10-19 21:40:15](https://github.com/yonaskolb/XcodeGen/assets/99916289/c2760e70-458d-4263-875e-466e37dff012)
![IMAGE 2023-10-19 21:40:24](https://github.com/yonaskolb/XcodeGen/assets/99916289/12383871-e85f-4dd5-ac8a-c81206a3d286)

PR #397, opened 2018-09-13
Ignore normal directory with dots

toshi0383:
Fixes https://github.com/yonaskolb/XcodeGen/issues/396


brentleyjones:
“.bundle”

toshi0383:
Yea I worry about it too, but if we thought about it at first time, this should be the correct approach, isn't it?

Sources/XcodeGenKit/PathKitExtension.swift
```diff
@@ -0,0 +1,17 @@
...
+    /// Treat this as a resource instead of a normal directory.
+    ///
+    /// - NOTE: "lproj" is handled differently.
+    var isFileLikeDirectory: Bool {
+        guard let ex = self.extension else { return false }
+        return ["imageset",
+                "xcassets",
+                "complicationset",
+                "appiconset",
+                "xcdatamodeld"].contains(ex)
```
brentleyjones: We should add "bundle" and "framework" to this as well.

toshi0383: Will do👌

brentleyjones: I wonder if there is an API to detect this? It seems to be OS defined.

toshi0383: Found an API to get UTI in string. 1e21b1d
It's `"public.folder"` if it's a normal folder.
`lproj` appears to be `"public.folder"` too. It looks problematic, but we're whitelisting `lproj`, so it wouldn't be a problem.

toshi0383:
```
extension: appiconset, uti: public.folder
extension: bundle, uti: com.apple.generic-bundle
extension: complicationset, uti: public.folder
extension: framework, uti: com.apple.framework
extension: imageset, uti: public.folder
extension: lproj, uti: public.folder
extension: xcassets, uti: com.apple.dt.assetcatalog
extension: xcdatamodeld, uti: com.apple.xcode.model.data-version
```

Sources/XcodeGenKit/PathKitExtension.swift
```diff
@@ -0,0 +1,23 @@
+import Foundation
+import PathKit
+
+extension Path {
+
+    /// Treat this as a resource instead of a normal directory.
+    var isNonFolderDirectory: Bool {
```
pepibumur: `isNonFolderDirectory` sounds a bit ambiguous. What about `isGroupDirectory`? That's how Xcode names directories in an Xcode project.

pepibumur: Then you could just say `!myPath.isGroupDirectory`

Sources/XcodeGenKit/PathKitExtension.swift
```diff
@@ -0,0 +1,23 @@
...
+
+        if isFile {
+            return false
+        }
+
+        if let uti = try! URL(fileURLWithPath: self.string)
+            .resourceValues(forKeys: [URLResourceKey.typeIdentifierKey])
+            .typeIdentifier {
+            // NOTE: lproj is public.folder
+            return uti != "public.folder"
```
pepibumur: If someone looks at this code in three months, it's hard to say what it does. Which types of paths am I ignoring with this expression. Maybe it's worth adding a larger comment where you explain the rationale behind this line.

pepibumur:
> Yea I worry about it too, but if we thought about it at first time, this should be the correct approach, isn't it?

Not making the most appropriate decision in the past doesn't justify that the "correct" approach should be introduced regardless of it breaking the contract with the projects that are already using the tool. What I would do is:

1. Add tests that cover the most common file types.
2. Check whether the behaviour is the same after your change.

If it's not, there's a decision to be made: either introducing a breaking change and going with a major bump, or finding a way to roll this out in a non-breaking manner. @yonaskolb is the person to help you with that.

Sources/XcodeGenKit/PathKitExtension.swift
toshi0383: I understand, but I think it's depends on how you look at the problem.

From Xcode perspective, `isGroupDirectory` is maybe straight forward, but this specific functionality is seen outside of Xcode too. That's why I've put this as `PathKitExtension`. 

Also, I actually tried to rename it, and ended up with rather complicated code. (Needed additional `is a directory, but` condition to retrieve **non "public.folder" directory**.)

```swift
        let directories = children
            .filter { $0.isDirectory && $0.isGroupDirectory && $0.extension != "lproj" }
            .sorted { $0.lastComponent < $1.lastComponent }

        let filePaths = children
            .filter { $0.isFile || ($0.isDirectory && !$0.isGroupDirectory) }
            .sorted { $0.lastComponent < $1.lastComponent }
```

Therefore, I think it's better to define this way (`isNonFolderDirectory`) in whitelist manner.

Sources/XcodeGenKit/PathKitExtension.swift
toshi0383: Thanks. Will do. Maybe as a doc-comment.

toshi0383:
> 1. Add tests that cover the most common file types.
> 2. Check whether the behaviour is the same after your change.

Thanks, will do👌

toshi0383:
Just found out how to list all recognized filetypes on mac. It lists all icons for each filetypes too.
https://discussions.apple.com/thread/3117503

I'll collect filetypes from these things.
```
            CFBundleTypeExtensions =             (
                xcodeproj,
                xcode,
                pbproj
            );
            CFBundleTypeIconFile = "xcode-project_Icon";
            CFBundleTypeName = "Xcode Project";
            CFBundleTypeRole = Editor;
            LSIsAppleDefaultForType = 1;
            LSItemContentTypes =             (
                "com.apple.xcode.project"
            );
            LSTypeIsPackage = 1;
            NSDocumentClass = Xcode3ProjectDocument;
            "_LSIconPath" = "Contents/Resources/xcode-project_Icon.icns";
        },
```

```
                conforms to:   com.apple.package, com.apple.dt.interfacebuilder.documen
t.storyboard
                tags:          .storyboardc
        --------------------------------------------------------
        type    id:            54184
                uti:           com.apple.dt.assetcatalog
                description:   Xcode Asset Catalog
                flags:         exported  active  apple-internal  trusted
                icon:
                conforms to:   public.directory
                tags:          .xcassets
        --------------------------------------------------------
        type    id:            54188
                uti:           com.apple.dt.stickers
                description:   Xcode Stickers Catalog
                flags:         exported  active  apple-internal  trusted
                icon:
                conforms to:   public.directory
                tags:          .xcstickers
        --------------------------------------------------------
        type    id:            54192
                uti:           com.apple.interfacebuilder.plugin
```

As a side note:
I'm not sure if we should depends on this database (which means dynamically executing and parsing this command output to get all these special filetypes). It looks like end-user can update these states using the `lsregister` command or something else.

Sources/XcodeGenKit/PathKitExtension.swift
toshi0383: done in 6933cde

Tests/XcodeGenKitTests/SourceGeneratorTests.swift
```diff
@@ -70,12 +39,14 @@ class SourceGeneratorTests: XCTestCase {
...
                 try pbxProj.expectFile(paths: ["Sources", "A", "B", "b.swift"], buildPhase: .sources)
+                try pbxProj.expectFile(paths: ["Sources", "A", "C.D.E", "c.d.e.swift"], buildPhase: .sources)
             }
 
             $0.it("supports frameworks in sources") {
                 let directories = """
                 Sources:
-                  - Foo.framework
+                  - Foo.framework:
+                    - Foo.swift
```
toshi0383: `*.framework` is now required to be a directory, not file.

toshi0383:
@pepibumur @yonaskolb 

> 1. Add tests that cover the most common file types.

Added tests in b24a7bf .
File extensions are retrieved by this one-liner.

> `(for i in $(/System/Library/Frameworks/CoreServices.framework/Versions/A/Frameworks/LaunchServices.framework/Versions/A/Support/lsregister -dump | grep ' \.[a-zA-Z][a-zA-Z]*' | sed -n 's/.*:  *\(.*\)/\1/p' | sort | uniq | sed "s,[ '],,g" | sed 's/,/ /g'); do echo $i; done) | grep '^\.' | grep -v '^\.\.*$' | sort | uniq`

(Sorry for poor readability. The original output is not in any standard format and is hard to parse.)

> 2. Check whether the behaviour is the same after your change.

The behavior is different for **normal directory with dots**. They had been treated as `.resources` BuildPhase (as shown in #396 ), but now as normal directories.
Other things like "special directory with dots" or "normal directory without dots" are not affected.

Tests/XcodeGenKitTests/SourceGeneratorTests.swift
```diff
@@ -517,6 +488,141 @@ class SourceGeneratorTests: XCTestCase {
...
+
+                    let groupPath = pbxProj.objects.versionGroups.values.first?.path
+                    try expect(groupPath) == "Hello.xcdatamodeld"
+                }
+            }
+        }
+    }
+}
+
+// MARK: Utilities
```
toshi0383: Moved utility functions below the all test cases. What we are interested in most is the test cases themselves, not utilities.

toshi0383:
And this is the script to determine if a file extension has a special meaning or not.
https://gist.github.com/toshi0383/41b7d3bcc27ae01246558114da6bcdd7

toshi0383:
@yonaskolb 
Not sure why the test is failing. It passes locally with Xcode9.3 on HighSierra.

Sources/XcodeGenKit/PathKitExtension.swift
```diff
@@ -0,0 +1,27 @@
...
+extension Path {
+
+    /// Treat this as a file (resource or source) instead of a normal directory.
+    var isNonFolderDirectory: Bool {
+
+        if isFile || self.extension == nil {
+            return false
+        }
+
+        if let uti = try! URL(fileURLWithPath: self.string)
```
yonaskolb: I'd like to get the performance tests in (coming shortly) so this can be tested for performance

toshi0383:
Updated CHANGELOG in 9227503

Sources/XcodeGenKit/PathKitExtension.swift
toshi0383: Done in 3aac581

toshi0383:
Investigating CI failure. Maybe returned UTI is different on Mac VMs?

toshi0383:
> Path.string: TestDirectory/Sources/Hello.agilekeychain, uti: public.folder

So we need to figure out how to determine if an UTI is from System, or from third party app.

toshi0383:
Closing for now.
Feel free to take over or reopen!👋🏻

Issue #214, opened 2017-12-27
Add spec linting


yonaskolb:
This would run the current validator (which highlights errors), but also a seperate linting phase that would provide suggestions on how to simplify the spec using all it's features eg. use the `deploymentTarget` instead of setting `IPHONEOS_DEPLOYMENT_TARGET`

yonaskolb:
This could have a fixit mode that automatically applied suggestions

Lutzifer:
Almost 5 Years later: Is there something like this? I just spent 2 hours debugging, because my dependencies were indented one level too deep...

yonaskolb:
This doesn't exist yet, but I image it would share some code with what's needed in the migration command https://github.com/yonaskolb/XcodeGen/pull/735

PR #402, opened 2018-09-14
Allow specifying run script and copy files phase ordering

### LINKED ISSUE #400: Should allow specifying ordering of copy files phases
From Xcode 10 release notes:

> While a target is performing post-compilation build steps, such as linking and code signing, Xcode’s new build system will begin compiling sources from targets that depend on that target, to improve build speed by using more of your available processor cores. This compilation will not start until any “Run Script” phases from the target have completed. (38772619)

We should allow specifying of the ordering of copy file phases. Scripts will block the compilation of a dependent target, while copy files ~~(probably)~~ doesn't. Being able to determine the order of these can improve build times, but also impact correctness of a build. The copy files phase might have to come before a post-build run script or even before Compile Sources, in order to allow dependent targets to compile correctly.

brentleyjones:
Fixes #400.

- Allows specifying run scripts to run immediately after the Compile Sources phase (which should be used instead of `postbuildScripts` 99% of the time).
- Allows specifying copy files phases to be run before the Compile Sources phase
    - Adjusts modulemap and static library heady copy files phases to be run pre Compile Sources
    - Adjusts the Carthage copy-frameworks to be run sooner


Sources/ProjectSpec/TargetSource.swift
```diff
@@ -119,6 +133,7 @@ public struct TargetSource: Equatable {
         self.type = type
         self.optional = optional
         self.buildPhase = buildPhase
+        self.headerVisibility = headerVisibility
```
brentleyjones: I noticed this wasn't being properly set. Didn't necessarily intend to commit it with this.

Sources/XcodeGenKit/SourceGenerator.swift
```diff
@@ -194,7 +195,8 @@ class SourceGenerator {
                 guard targetType == .staticLibrary else { return nil }
                 return .copyFiles(TargetSource.BuildPhase.CopyFilesSettings(
                     destination: .productsDirectory,
-                    subpath: "include/$(PRODUCT_NAME)"
+                    subpath: "include/$(PRODUCT_NAME)",
+                    phaseOrder: .preCompile
```
brentleyjones: This being `.preCompile` is a prime example of why these changes are needed.

brentleyjones:
Even with these changes the new build system is broken. That doesn't make these changes not needed though. Eventually Xcode 10 will be fixed, but being fixed will still warrant these changes.

toshi0383:
@brentleyjones 

> which should be used instead of postbuildScripts 99% of the time

Question: What do you expect for remaining 1% of them?

Also I think `sources`'s `buildPhase` structure is getting too complicated. It should look as same as we see on Xcode.
Ideally I expect a format like this.

```yml
        buildPhase:
          copyFiles:
            destination: productsDirectory
            subpath: include
          compile:      // <- This simply makes the processing order clearer.
          runScript:
            path: my-script.sh
```

brentleyjones:
> Question: What do you expect for remaining 1% of them?

When you want to post-process something that has been copied into the products folder (like adjusting application icons for debug versus releases).

Sources/ProjectSpec/TargetSource.swift
yonaskolb: I'm ok with it being here

Sources/ProjectSpec/Target.swift
```diff
@@ -32,6 +32,7 @@ public struct Target: ProjectTarget {
     public var directlyEmbedCarthageDependencies: Bool?
     public var requiresObjCLinking: Bool?
     public var prebuildScripts: [BuildScript]
+    public var postCompileScripts: [BuildScript]
```
yonaskolb: Now the old `prebuildScripts` and `postbuildScript` names look funny. Maybe we should update them to `preBuildScripts` and `postBuildScript`, and use the old casing as a fallback?
What do you think?

Docs/ProjectSpec.md
```diff
@@ -373,7 +376,7 @@ targets:
 
 ### Build Script
 
-Run script build phases added via **prebuildScripts** or **postBuildScripts**. They run before or after any other build phases respectively and in the order defined. Each script can contain:
+Run script build phases added via **prebuildScripts**, **postCompileScripts**, or **postBuildScripts**. They run before any other build phases, after the Compile Sources build phase, or after any other build phases respectively and in the order defined. Each script can contain:
```
yonaskolb: Now that we have 3, maybe we need to split them into explicit sections:

Run script build phases can be added at 3 different points in the build:
- **prebuildScripts**: Before any other build phases
- **postCompileScripts** After the compile sources build phase
- **postbuildScripts**: After any other build phases

Each script can contain:

Sources/ProjectSpec/Target.swift
brentleyjones: Sounds good.

brentleyjones:
@yonaskolb Care to give this another look?

Sources/ProjectSpec/Target.swift
```diff
@@ -279,8 +283,9 @@ extension Target: NamedJSONDictionaryConvertible {
         directlyEmbedCarthageDependencies = jsonDictionary.json(atKeyPath: "directlyEmbedCarthageDependencies")
         requiresObjCLinking = jsonDictionary.json(atKeyPath: "requiresObjCLinking")
 
-        prebuildScripts = jsonDictionary.json(atKeyPath: "prebuildScripts") ?? []
-        postbuildScripts = jsonDictionary.json(atKeyPath: "postbuildScripts") ?? []
+        preBuildScripts = jsonDictionary.json(atKeyPath: "preBuildScripts") ?? jsonDictionary.json(atKeyPath: "prebuildScripts") ?? []
```
yonaskolb: Makes me think the `lint` command in https://github.com/yonaskolb/XcodeGen/issues/214 could also be used to highlight deprecations

Docs/ProjectSpec.md
```diff
@@ -373,7 +377,13 @@ targets:
 
 ### Build Script
 
-Run script build phases added via **prebuildScripts** or **postBuildScripts**. They run before or after any other build phases respectively and in the order defined. Each script can contain:
+Run script build phases can be added at 3 different points in the build:
```
yonaskolb: I guess another option would be to have a single `buildScripts` array with each item having an `order` enum. Is this cleaner and more extensible? Thoughts?

Docs/ProjectSpec.md
```diff
@@ -275,6 +276,9 @@ A source can be provided via a string (the path) or an object of the form:
 			- `sharedSupport`
 			- `plugins`
 		- [ ] **subpath**: **String** - The path inside of the destination to copy the files.
+		- [ ] **phaseOrder**: **String** - When the Copy Files phase should execute in relation to other phases. This can be one of the following values:
```
yonaskolb: Does this need to be exposed to users? Would it ever be required? I'd rather keep the options we expose low. XcodeGen is supposed to remove all those sorts of decisions.

brentleyjones: I felt it should be exposed since a user could choose to have a specific set of resources copied before compilation. We currently have those covered, AFAIK, but others might have more exotic needs.

Docs/ProjectSpec.md
brentleyjones: Since we already have the two values, I think this is cleaner. These should be the only 3 spots needed. If we did any refactor I would want to be able to specify the order of all of the phases. It's hard for me to imagine how that would look, and it's probably a big enough breaking change that it would need to be part of a 2.0 release.

Docs/ProjectSpec.md
yonaskolb: I'd rather leave this out for now, and only add it if it someone has a need for it 👍

yonaskolb:
Hi @brentleyjones. Would you be able to get this branch up to date resolve the conflicts?

brentleyjones:
@yonaskolb Yeah, I'll try to get to it today 😄.

brentleyjones:
@yonaskolb Assuming tests pass, I believe this is all merged up. I also removed the ability to specify copy files phase ordering.

Docs/ProjectSpec.md
brentleyjones: @yonaskolb Coming back from WWDC I've learned that the way I'm doing module map copying currently isn't the best. While hacking through a local solution before figuring out how XcodeGen should be updated I ran into wanting to specify the phase order of the copy phase. Essentially I need to copy files before compilation. Wondering how you feel about me exposing this option now?

PR #410, opened 2018-09-26
Make validation of Xcode projects optional

jerrymarino:
Validation requires several things of the Xcode project, one being that files are already on disk. In some cases, it's useful to generate projects containing files that don't exist on disk yet. Additionally, this is useful for testing purposes, and generating projects on a machine where the files don't exist.

Additionally, this flag is needed to run source code generators and download files at build time, but want to have an Xcode project containing entries for those files.


jerrymarino:
@yonaskolb thank you kindly for the prompt review and merging 👏

Issue #409, opened 2018-09-26
Project lock file


yonaskolb:
I've been thinking a bit about the workflow of branch switching when you use XcodeGen. Currently many people are regenerating the project after a checkout (sometimes in a git checkout hook) to make sure the project is up to date, in case the project spec has changed or files have been added or removed. In a very large project, this can potentially take a couple of seconds.

I propose adding a project lock file that is used to track whether generation is necessary.

1. Every time the project is generated, all the files it includes in the project are looped through and collected. These files would be things source files and xcconfig files. Then all these file paths along with the spec itself would be written to a `project.lock` file (name TBD).
1. A new argument can be added to `xcodegen`, something like `--check-lock` (name?). When this is provided it checks for a `project.lock` file and compares the current spec with the one included in the lock file. 
    - if it's different generate the file
    - if it's the same then collect all the files in the current spec the same as it does when writing the lock file
    - if these files differ from what is in the lock file then generate

For those using Cocoapods, a pod install has to be run after project generation. Another argument `--pod-install` could run pod install after generation. This would mean it only runs if get past the lock file.

This command could then be added as a git checkout hook (perhaps with a command in XcodeGen itself)

Thoughts?

ryohey:
Sounds good! I also want to use `--pod-install`.

However, since I think that it is also necessary to obtain a list of source files when generating projects and comparing project.lock, either way, it will take more time for more files.
I didn't measure precisely, as far as seen the log, are seen in our project it takes about 3 seconds for generating project and about 1 second for writing project. I am thinking that much time of project generation is used for scanning source files, so I am uncertain about how much we can improvements with lock file.

LinusU:
I've personally solved this with a `Makefile` that have the Xcode project dependent on the `project.yml` file. Thus, every time that `project.yml` is edited, the Xcode project will be rebuilt before continuing with anything else.

Not saying that this shouldn't be a good addition ☺️ just sharing my current setup.

keith:
We currently solve this problem by:

1. Checking if the project definition has changed via diffing the sha256 of the file vs the last time we generated the project
2. Using `git ls-files` to see if any files have been added or deleted

For those interested this is our script for this:

```sh
#!/bin/bash

set -o pipefail
set -u

touch ./tmp/project_generated.json.sha256
shasum --check ./tmp/project_generated.json.sha256 > /dev/null 2>&1
project_changed=$?
[...]

PR #412, opened 2018-09-29
Add caching

yonaskolb:
This is a start at resolving #409, and solving the problem of generating projects when they don't need to be. This makes using xcodegen in a git-checkout hook much better

How it currently works is you have to pass a `--lockfile` argument to xcodegen specifying a file it should read and write to for determining if the project has changed. eg `xcodegen --lockfile project.lock`. This file would be added to `.gitignore`.

If this file exists it is read before generation. Only if any of the following has changed will it continue with generation:
- the xcodegen version
- the fully resolved spec (including `includes`)
- any files the the spec references have been added or removed, via project filegroups, targets sources and xconfig files

I'm not sold on the idea of the name lockfile. Not sure if the filename should be configurable either or just be a default filename, or configurable via the spec itself?

Another possible enhancement would be handling only running `pod install` if no generation occurs.
This could happen via:
1. not doing anything and making the user parse the stdout manually to check
2. a `--pod-install` flag that automatically runs pod install
3. xcodegen erroring if the project isn't generated
4. ?

Tasks:
- [x] reading and writing the lockfile
- [x] name change?
- [ ] pod install story
- [ ] tests
- [x] performance tests
- [x] documentation

EDIT: Renamed feature to caching


Sources/ProjectSpec/Project.swift
```diff
@@ -159,3 +159,40 @@ extension Project {
...
+
+        for target in aggregateTargets {
+            files.append(contentsOf: target.configFilePaths)
+        }
+        
+        for target in targets {
+            files.append(contentsOf: target.configFilePaths)
+            for source in target.sources {
+                let sourcePath = basePath + source.path
+                let sourceChildren = (try? sourcePath.recursiveChildren()) ?? []
```
yonaskolb: I guess I could filter out directories?

keith:
I get this error when trying to use this:

```
Couldn't serialize spec for lockfile```

**Comment by keith** [comment]
I added the specific error to the log and got this:

```
Couldn't serialize spec for lockfile Failed to represent {
    Alpha = release;
    Beta = release;
    Debug = debug;
    Labs = release;
    Profile = release;
    QA = debug;
    Release = release;
}
```

**Comment by keith** [comment]
We can test this once this is working fine, but in our experience using the json project spec instead of the yaml one was significantly faster, that might make sense to do for the lockfile too, especially since the format of it shouldn't matter for consumers

**Comment by keith** [comment]
Ok with this change (we'd have to solve the 10.13 issue some other way):

```patch
diff --git i/Sources/XcodeGen/main.swift w/Sources/XcodeGen/main.swift
index 132c48b..dcf7caf 100644
--- i/Sources/XcodeGen/main.swift
+++ w/Sources/XcodeGen/main.swift
@@ -44,7 +44,7 @@ func fatalError(_ message: String) -> Never {
     // Lock file
     var lockFileContent: String = ""
     let lockFilePath = lockfile.isEmpty ? nil : Path(lockfile)
-    if let lockFilePath = lockFilePath {
+    if let lockFilePath = lockFilePath, #available(macOS 10.13, *) {
 
         let files = Array(Set(project.allFiles))
             .map { $0.byRemovingBase(path: project.basePath).string }
@@ -53,8 +53,9 @@ if let lockFilePath = lockFilePath {
 
         let spec: String
         do {
-            let node = try Node(projectDictionary)
-            spec = try Yams.serialize(node: node)
+            let data = try JSONSerialization.data(withJSONObject: projectDictionary,
+                                                  options: [.sortedKeys,.prettyPrinted])
+            spec = String(data: data, encoding: .utf8)!
         } catch {
             fatalError("Couldn't serialize spec for lockfile")
         }
```

I was able to use this successfully. Here are our timings:

```
% time ~/dev/XcodeGen/.build/release/xcodegen --quiet --spec project_generated.json --lockfile foo.lock
~/dev/XcodeGen/.build/release/xcodegen --quiet --spec project_generated.json   15.74s user 1.08s system 99% cpu 16.863 total
% time ~/dev/XcodeGen/.build/release/xcodegen --quiet --spec project_generated.json --lockfile foo.lock
~/dev/XcodeGen/.build/release/xcodegen --quiet --spec project_generated.json   2.39s user 0.30s system 99% cpu 2.697 total
```

We're currently on 1.11.1 and our gen time on my machine is ~13 seconds. So we're seeing a pretty serious hit here. I think the no-op time is likely a little higher than we'd want. Maybe that's different because of my change though. Happy to help profile more if you'd like.

toshi0383:
> --lockfile

Is there any reason why it's opt-in, instead of defaulting to somewhere like `~/.xcodegen/{unique-hash-from-project-path}` ?
I thought it's better to "use cache by default".

> --pod-install

I don't think we should know about how to run other third party commands. Especially cocoapods does a lot of complicated stuff, so I would leave it alone, not adding `--pod-install`.
BTW I use cocoa pods via bundler by `bundle exec pod install` to make sure to stick with specific version of cocoapods.

ryohey:
I tried it for now. It works fine with a project with 3000 source files. awesome!
I compared the generation time with lockfile with and without lockfile, there is a variation in measurement, but there seemed to be no problem.

<details>

```bash
$ time ./XcodeGen --spec project.yml --lockfile project.lock --quiet
        7.52 real         6.82 user         0.44 sys
$ time ./XcodeGen --spec project.yml --lockfile project.lock --quiet
        1.59 real         0.78 user         0.09 sys
$ time ./XcodeGen --spec project.yml --lockfile project.lock --quiet
        0.61 real         0.51 user         0.07 sys
$ time ./XcodeGen --spec project.yml --quiet
        6.87 real         6.23 user         0.39 sys
$ time ./XcodeGen --spec project.yml --quiet
       10.82 real         8.00 user         0.49 sys
$ rm project.lock
$ time ./XcodeGen --spec project.yml --lockfile project.lock --quiet
       12.68 real         9.10 user         0.49 sys
$ rm project.lock
$ time ./XcodeGen --spec project.yml --lockfile project.lock --quiet
       12.24 real         8.90 user         0.62 sys
$ rm project.lock
$ time ./XcodeGen --spec project.yml --lockfile project.lock --quiet
        6.00 real         5.58 user         0.35 sys
$ rm project.lock
$ time ./XcodeGen --spec project.yml --lockfile project.lock --quiet
        6.34 real         5.83 user         0.39 sys
```

</details>

yonaskolb:
@keith yeah json will be better 👍 

@toshi0383 it's opt in for now as it's just a test, but will be opt out if it runs performantly and without issues. I like your idea of writing the cache file to a special directory. That way people don't have to bother adding it to their `.gitignore`.
I agree to not calling `pod install` explicitly. We still need a way to run a command after project generation (or rather not run a command if it doesn't generate).

Any ideas for another name for this functionality? Maybe `cache file`, `diff file`, `project hash`?

toshi0383:
> We still need a way to run a command after project generation

Why don't we just do `xcodegen && rm -rf Pods/; pod install` ?

yonaskolb:
@toshi0383 because you wouldn't want to run an unnecessary and lengthy pod install if the project hasn't changed. 
One option is to add an argument to xcodegen that makes it fail if it doesn't generate anything due to no changes found. You could make it fail be default but that might be weird and unexpected?

toshi0383:
Or we can add option like `--exec` ?
`xcodegen --exec pod install`
`xcodegen --exec echo $(rm -rf Pods/; pod install)`

It gets a little complex when you want to execute multiple commands.

ref: https://github.com/sharkdp/fd#parallel-command-execution

keith:
If possible I think XcodeGen should avoid tying itself to CocoaPods or any other tool that isn't required for using XcodeGen itself

toshi0383:
`--exec` is too complex. Please forget.

> One option is to add an argument to xcodegen that makes it fail if it doesn't generate anything due to no changes found. You could make it fail be default but that might be weird and unexpected?

I'm 👌 to this option. Maybe with optional exit code from user.
e.g. `--gen-or-fail[=n]`

yonaskolb:
I've updated and renamed to `--use-cache` which saves the file to `~/.xcodegen/cache/{project_spec_path_hash}`.
This could be turned on by default if all is well

yonaskolb:
Also added a `--cache-path` for a custom path. This will allow more flexibility for places like a CI server

yonaskolb:
Still not sure on the naming of this feature. Does anyone have any thoughts on it?

PR #416, opened 2018-10-05
Improve performance of getCombinedBuildSettings

kastiglione:
This change refactors `getCombinedBuildSettings` for performance.

Using Instruments with our project and XCHammer (which uses XcodeGen), there was a large amount of memory management costs for calls to this method. In particular because this method was called for every target, which doesn't scale well for projects with lots of targets.

All use of `getCombinedBuildSettings()` was looking up the value for single key. For this use case, its inefficient to combine the  target, target-config, project, and project-config dictionaries into a single dictionary, since a dictionary is not the result wanted by callers. With this change, the function has been renamed to be singular (`getCombinedBuildSetting()`) and it takes a key to look for in the underlying dictionaries.

Note: Some unused parameters were also removed.

### DIFF
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -355,7 +355,7 @@ public class PBXProjGenerator {
 
             func getSingleBuildSetting(_ setting: String) -> String? {
                 let settings = project.configs.compactMap {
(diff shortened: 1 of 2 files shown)


brentleyjones:
Nice! What sort of improvements are you seeing with this change, in regards to memory and speed?

kastiglione:
I just did a few more runs and have seen times at 4s and 6s. Heap as low as 106mb but also one 141mb. Definitely not scientific here.

PR #417, opened 2018-10-05
Cache config/setting file misses too

kastiglione:
`SettingsBuilder.swift` has two caches for `BuildSettings`, for xcconfigs and settings presets. The caches ensure _existing_ files aren't redundantly reloaded, but for non-existent files they don't cache the fact that no settings file could be found, which means redundant searches for any optional settings file that does not happen to exist. Each failed search is repeated work. For the project I'm working on, this is most expensive in `SettingsPresetFile.getBuildSettings()`.

This change makes these caches also cache which config files do not exist. For just `getBuildSettings()`, this change reduces the memory use, down to 1.7mb, from a little over 10mb. The sample project has ~40 targets but real world targets for us will be bigger.

This change adds an `Optional` equivalent enum named `Cached`. This is to avoid having a dictionary value of an `Optional` type, otherwise the subscript operator would return an `Optional<Optional<T>>`, which can be awkward to work with.

### DIFF
--- a/Sources/XcodeGenKit/SettingsBuilder.swift
+++ b/Sources/XcodeGenKit/SettingsBuilder.swift
@@ -126,28 +126,43 @@ extension Project {
     /// Returns cached build settings from a config file
     private func loadConfigFileBuildSettings(path: String) -> BuildSettings? {
         let configFilePath = basePath + path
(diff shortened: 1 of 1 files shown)


yonaskolb:
Great stuff, thanks @kastiglione!

PR #419, opened 2018-10-08
Use the provided TargetSource.headerVisibility

jerrymarino:
This was missing in the init - simply use the value provided by the
user.


jerrymarino:
Thanks for the review and merging @yonaskolb 🎉

PR #423, opened 2018-10-18
Update generated project defaults to xcode 10 and swift 4.2

yonaskolb:
This updates the default Xcode project version to Xcode 10, and default Swift version to 4.2


MaxDesiatov:
@yonaskolb is there any way to set Swift version explicitly? Unfortunately, this causes build issues for my projects due to Swift version change without any warning. It doesn't look like there's any option to override Swift version in the [Project Spec](https://github.com/yonaskolb/XcodeGen/blob/master/Docs/ProjectSpec.md).

yonaskolb:
Hi @MaxDesiatov 
The swift version is just a build setting. You can change it for specific targets, or just the whole project like this:
```
settings:
   SWIFT_VERSION: `4.0`
```

PR #427, opened 2018-10-31
Fix issue with legacy target fileref

dflems:
`xcodegen` was crashing when generating a `legacy` target as a dependency of another target (in this case an application). In the case of legacy targets, `targetFileReferences[dependencyTarget.name]` seems to be `nil`, and the generator was crashing on the force-unwrap. The value was only for used targets that are embedded into the parent target's bundle, so I moved the unwrap to that location.


yonaskolb:
Fantastic, good catch @dflems! Could you add a change log entry as well please

yonaskolb:
I'll add a changelog in master 👍

PR #437, opened 2018-11-10
Add XcodeGenCLI

yonaskolb:
This refactors the cli:
- Swap `Commander` dependency for `SwiftCLI`
- Extract cli into XcodeGenCLI
- Add `generate` command
- Use `generate` command if no arguments are passed
- Sets the stage for more commands


Sources/XcodeGenCLI/GenerateCommand.swift
```diff
@@ -0,0 +1,85 @@
...
+        do {
+
+            let fileWriter = FileWriter(project: project)
+            try fileWriter.writeXcodeProject(xcodeProject, to: projectPath)
+            try fileWriter.writePlists()
+        } catch {
+            throw GenerationError.writingError(error)
+        }
+
+        success("💾  Saved project to \(projectPath)")
```
toshi0383: Floppy disk😂

yonaskolb: 😄 This was just copied over from the existing output

PR #443, opened 2018-11-23
Don't overwrite INFOPLIST_FILE setting with info path

### LINKED ISSUE #440: Generating Info.plist overwrites path specified in settings>configs
I have a project where a target has different configs. One of those configs uses a different Info.plist than the others. 
When I create a Info.plist with xcodegen it overwrites every config with the generated one, even if I have specified a path for it via INFOPLIST_FILE build setting.

Example: In this case Predifined.plist for debug gets overwritten with Generated.plist.
plistoverwrite

Suggestion: When there is a path for a predefined plist set, do not overwrite it with the created one.

Related issues: #207, #406

feischl97:
Resolves #440.
Just added a check if for a config INFOPLIST_FILE is already set in settings. If it is it will not be overwritten with the generated Info.plist.
I also updated projectspec.md to reflect this change.
I did not add any tests, are they necessary?

### DIFF
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -792,13 +792,11 @@ public class PBXProjGenerator {
                 buildSettings["CODE_SIGN_ENTITLEMENTS"] = entitlements.path
             }
 
-            // Set INFOPLIST_FILE
-            if project.targetHasBuildSetting("INFOPLIST_FILE", target: target, config: config) {
-                // do not overwrite Info.plist path when already defined
-            } else if let info = target.info {
-                buildSettings["INFOPLIST_FILE"] = info.path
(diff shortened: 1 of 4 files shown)


Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -793,7 +793,9 @@ public class PBXProjGenerator {
             }
 
             // Set INFOPLIST_FILE
-            if let info = target.info {
+            if project.targetHasBuildSetting("INFOPLIST_FILE", target: target, config: config) {
+                // do not overwrite Info.plist path when already defined
+            } else if let info = target.info {
```
yonaskolb: This could actually go inside the `if !project.targetHasBuildSetting("INFOPLIST_FILE", target: target, config: config)` below

feischl97:
Ok I will look into it! 👍 
Thanks @yonaskolb

PR #450, opened 2018-11-30
Use process unique temporary so we don't get duplicate temporary path…

### LINKED ISSUE #378: Generation failed: The folder “project.pbxproj” doesn’t exist.
We've been getting this warning sporadically with XcodeGen 1.10.2. I'm not sure how to reproduce it yet, but I was wondering if anyone else has seen this issue. It appears to be coming from this catch:  https://github.com/yonaskolb/XcodeGen/blob/6c35c332e973c4a0bfb0f87c285fcc4e23a8db55/Sources/XcodeGen/main.swift#L66

bryansum:
This commit resolves #378. The issue was that the temporary directory path was only distinguished by a second-level timestamp, which meant interleaved `xcodegen` calls would use the same temporary directory. This fix uses a globally unique string given by `NSProcessInfo`.

Test plan:
* Run multiple instances of XcodeGen at the same time in a tight loop (I used `while true; do xcodegen;  done`
* Observe that we don't see any errors reported as before.


yonaskolb:
Would you be able to add a changelog entry as well?

bryansum:
@yonaskolb Thanks! You would be the one to merge this, I take it?

yonaskolb:
Sorry thought I did 😄

PR #457, opened 2018-12-13
Add mlmodel to compiled file extensions

dwb357:
Xcode requires that '.mlmodel' files be handled as Sources and not as Resources.  They need to be processed and compiled and not copied as part of resource bundle.


brentleyjones:
Thanks @dwb357!

PR #459, opened 2018-12-19
Add includeBridgingHeader option to generate BridgingHeader easily

giginet:
# Motivation & Context

If a target includes Bridging Headers. We have to set `OBJC_SWIFT_OBJC_BRIDGING_HEADER` option. But it seems to be redundancy.

```yaml
targets:
  MyFramework:
    sources: MyFramework
    platform: [iOS, tvOS]
    type: framework
    settings:
      base:
        SWIFT_OBJC_BRIDGING_HEADER: MyFramework/MyFramework-Bridging-Header.h
```

# Description

I introduced a new option `includeBridgingHeader`.

```yaml
targets:
  MyFramework:
    sources: MyFramework
    platform: [iOS, tvOS]
    type: framework
    includeBridgingHeader: true
```

`includeBridgingHeader` helps to write `SWIFT_OBJC_BRIDGING_HEADER` instead.
The value would be defined from the target's name.
`MyFramework/MyFramework-Bridging-Header.h`


toshi0383:
Hi @giginet 👋

Thanks for the suggestion, but I don't think it's a good idea. We can't implicitly depend on specific location and filename for a bridging header. It can be in any directory with any name, right?
Maybe automating to create a bridging header would make sense, but I don't think it's possible and it's out of XcodeGen's responsibility anyways (I guess).

Any other thoughts on this?

toshi0383:
FYI, I usually use xcconfig instead of `settings`. So that's an option to simplify things.

brentleyjones:
I believe this is set automatically for you unless you set it to blank. So I’m not seeing why this is needed?

yonaskolb:
Hi @giginet, thanks for your contribution! I can see this would be usefully for quite a few cases, but as has been pointed out the value of this build setting isn’t always the same. That leaves the option to turn this from a Boolean into a path string, but then we get into the territory of defining every possible building setting into a friendly name. While that is indeed possible and potentially useful, that’s not something I feel this project can support at this time.
Having said that if lots of people chime in and say this is useful as a default case, I’d be open to adding it. 
For now, specifying the build setting manually via `setttings` or `configFiles` will have to suffice

PR #465, opened 2019-01-02
Bug carthage dependencies with different platform

raptorxcz:
When project is generated, then in iOS app are added all carthage dependencies.
This behaviour is wrong, because iOS target can have watchOS target as a dependency.

### DIFF
--- a/Tests/XcodeGenKitTests/SourceGeneratorTests.swift
+++ b/Tests/XcodeGenKitTests/SourceGeneratorTests.swift
@@ -517,6 +517,23 @@ class SourceGeneratorTests: XCTestCase {
                 try expect(sourcesBuildPhase.files.count) == 1
             }
 
+            $0.it("add only carthage dependencies with same platform") {
+                let directories = """
+                    A:
+                    - file.swift
+                """
+                try createDirectories(directories)
+
+                let watchTarget = Target(name: "Watch", type: .watch2App, platform: .watchOS, sources: ["A"], dependencies: [Dependency(type: .carthage, reference: "Alamofire_watch")])
+                let watchDependency = Dependency(type: .target, reference: "Watch")
+                let target = Target(name: "Test", type: .application, platform: .iOS, sources: ["A"], dependencies: [Dependency(type: .carthage, reference: "Alamofire"), watchDependency])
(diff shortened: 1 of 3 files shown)


yonaskolb:
Hmm, maybe this behaviour is indicative of a deeper issue in that targets shouldn't be able to specify target dependencies that aren't the same platform. That could be added to the spec validation pass.

Or do you have some weird case where one target has a dependency on a target with a different platform? Maybe there's watchOS extension case I'm not thinking of

raptorxcz:
The project generating is correct. In iOS bundle can be watchOS app. I am not sure, but I thing same behavious can be achieved with tvOS too.

yonaskolb:
Could you also add a changelog entry

Also can you explain a use case that this fixes? The tests use `Alamofire_watch` but is that a real use case? Wouldn't the watch target just use `Alamofire`?

Would another way of thinking of this be don't include carthage dependencies from app targets?

Alex293:
I have multiple examples of libs that behave like this. I’ll post them asap

raptorxcz:
Hi @yonaskolb, 

Is is possible to merge this PR or is it required to implement the solution you suggested?

Thanks

Tests/XcodeGenKitTests/SourceGeneratorTests.swift
```diff
@@ -517,6 +517,23 @@ class SourceGeneratorTests: XCTestCase {
                 try expect(sourcesBuildPhase.files.count) == 1
             }
 
+            $0.it("add only carthage dependencies with same platform", closure: {
```
yonaskolb: Can this be made a trailing closure

yonaskolb:
@raptorxcz I guess there's nothing wrong with this directly, so we can merge it. Just needs a rebase and a small tweak

Tests/XcodeGenKitTests/SourceGeneratorTests.swift
raptorxcz: Done in ec60901ff749383943ced49e3b77aeb60c50603a.

raptorxcz:
Updated in ec60901ff749383943ced49e3b77aeb60c50603a.

PR #471, opened 2019-01-03
⛳️ swift for tensorflow example

johndpope:


yonaskolb:
Thanks @johndpope. Just checking you haven’t missed the build setting docs in 
https://github.com/yonaskolb/XcodeGen/blob/master/Docs/Usage.md and 
https://github.com/yonaskolb/XcodeGen/blob/master/Docs/ProjectSpec.md

PR #472, opened 2019-01-04
⛳️ include xcconfig sample

johndpope:
I think this cuts to the heart of bridging complexity between yml and more advanced compiler flags.


yonaskolb:
Thanks for this @johndpope but I think I'd rather keep the intro example to a single file. 
How do you mean the `complexity between yml and more advanced compiler flags`? Are you aware you can also set build settings within the spec itself?

In any case the docs in `/Docs` can always be improved

johndpope:
I'

PR #480, opened 2019-01-07
 Added an abbility to generate empty directories

Beniamiiin:
I added a new option `generateEmptyDirectories`. Default to `true`.

### DIFF
--- a/Docs/ProjectSpec.md
+++ b/Docs/ProjectSpec.md
@@ -93,7 +93,7 @@ Note that target names can also be changed by adding a `name` property to a targ
   - `bottom` - at the bottom, after other files
   - `none` - sorted alphabetically with all the other files
 - [ ] **transitivelyLinkDependencies**: **Bool** - If this is `true` then targets will link to the dependencies of their target dependencies. If a target should embed its dependencies, such as application and test bundles, it will embed these transitive dependencies as well. Some complex setups might want to set this to `false` and explicitly specify dependencies at every level. Targets can override this with [Target](#target).transitivelyLinkDependencies. Defaults to `false`.
-- [ ] **generateEmptyDirectories**: **Bool** - If this is `true` then empty directories will be added to project too else will be missed. Defaults to `true`.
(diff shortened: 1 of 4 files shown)


Docs/ProjectSpec.md
```diff
@@ -93,6 +93,7 @@ Note that target names can also be changed by adding a `name` property to a targ
   - `bottom` - at the bottom, after other files
   - `none` - sorted alphabetically with all the other files
 - [ ] **transitivelyLinkDependencies**: **Bool** - If this is `true` then targets will link to the dependencies of their target dependencies. If a target should embed its dependencies, such as application and test bundles, it will embed these transitive dependencies as well. Some complex setups might want to set this to `false` and explicitly specify dependencies at every level. Targets can override this with [Target](#target).transitivelyLinkDependencies. Defaults to `false`.
+- [ ] **generateEmptyDirectories**: **Bool** - If this is `true` then empty directories will be added to project too else will be missed. Defaults to `true`.
```
yonaskolb: We'll have to keep the default as false for backwards compatibility

Beniamiiin: Yes, you are right, I'll change it to `false`.

Beniamiiin:
@yonaskolb 
> An entry in the changelog under Master/Added
I'm so sorry I don't understand what I should to add to `CHANGELOG.md`, could you give me please more details.

> Possibly a unit test if you have time in https://github.com/yonaskolb/XcodeGen/blob/master/Tests/XcodeGenKitTests/SourceGeneratorTests.swift
I tried to fix it, but unfortunately I didn't understand how can I do it. If you give me some advice, I'll do it.

yonaskolb:
> I'm so sorry I don't understand what I should to add to CHANGELOG.md, could you give me please more details.

Just an entry like the others. At the moment there is nothing new in master so it will be the first entry
```md
# Change Log

## Master

#### Added
- Added new ability to generate empty directories via `options.generateEmptyDirectories` [#480](https://github.com/yonaskolb/XcodeGen/pull/480) @Beniamiiin

## 2.1.0
```

In terms of the tests I can add those later. It seems the source generator test helpers need to be updated to generate empty directories and check for groups

PR #482, opened 2019-01-07
Update Linkage to handle additional PBXProductType case from xcodeproj

### LINKED ISSUE #481: XcodeGen 2.1.0 fails to build as an SPM dependency
When adding XcodeGen 2.1.0 as a SPM dependency, compilation fails with the following message:

```
.../XcodeGen.git-2328129159905136123/Sources/ProjectSpec/Linkage.swift:12:9: error: switch must be exhaustive
        switch type {
        ^
.../XcodeGen.git-2328129159905136123/Sources/ProjectSpec/Linkage.swift:12:9: note: add missing case: '.instrumentsPackage'
        switch type {
        ^
```

It looks like the latest release of `xcodeproj` (6.4.0) added another case to `PBXProductType` which isn't being handled.

ksulliva:
This resolves #481.

Changes:
- Update `Package.resolved` to ensure the latest version of `xcodeproj` is being used.
- Handle the newly-added `.instrumentsPackage` case of `PBXProductType`

I didn't make the change in this PR, but it might be worth pinning the version of `xcodeproj` in the `Package.swift` file to prevent additional changes to that library from affecting this project.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Master
 
+#### Added
+- Added support for the `instrumentsPackage` product type
+
 ## 2.1.0
 
 #### Added
--- a/Docs/ProjectSpec.md
+++ b/Docs/ProjectSpec.md
@@ -215,6 +215,7 @@ This will provide default build settings for a certain product type. It can be a
 - `bundle.ui-testing`
 - `bundle.ocunit-test`
 - `framework`
+- `instruments-package`
 - `library.dynamic`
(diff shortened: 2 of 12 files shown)


yonaskolb:
Great, thanks @ksulliva! 
Could you please add a changelog entry under `Master/Added` saying something like 
```
Added support for the `instrumentsPackage` product type
```

Also could you update the product list in https://github.com/yonaskolb/XcodeGen/blob/master/Docs/ProjectSpec.md#product-type

PR #489, opened 2019-01-13
Spec file relative directory expansion

### LINKED ISSUE #385: Nested project.yml includes not using relative paths (modular project.yml files)
I am importing (copy-pasting) other projects each that have their own `project.yml` files for Xcodegen. Here is the directory structure:

```
- ProjectFolder/
    - project.yml
    - App/
    - AppTests/
    - DependencyA/
        - project.yml
        - DependencyAApp/
        - DependencyATests/
```

In my `ProjectFolder/project.yml`, I define `include: [DependencyA/project.yml]`. In my `DependencyA/project.yml`, I define:

```
name: DependencyAApp
targets:
    DependencyAApp:
        sources:
            - path: DependencyAApp
        scheme:
            testTargets:
                - DependencyATests
    DependencyATests
        sources:
            - path: DependencyATests
        dependencies:
            - target: DependencyAApp
```

When I `cd ProjectFolder` and run `xcodegen`, I get an error saying that `- Target "DependencyATests" has a missing source directory "ProjectFolder/DependencyATests"` whereas I would expect `sources` defined in a `yml` file to be relative to that `yml` file and not the root one including it.

Is there an option flag that I am missing? In this project, I can't use Pods or Carthage. I'm also aware that:
1. I can define `sources` in `DependencyA/project.yml` to include that folder so it looks like `- path: DependencyA/DependencyATests` which then works
2. I can also combine all the `project.yml` files into one file and include `DependencyA/` in the `sources` path

But I don't like either of the solutions as I want to maintain the modularity of the `project.yml` files.

Is this possible?

ellneal:
Closes #385.

This PR started as a bug fix but ended up including quite a bit refactoring; so I welcome any feedback (I've tried to stay as true to the style of the existing code as I could).

### The Problem

`Included` spec files still used the root spec file's directory as their base directory.

```
Root/
    project.yml
    Sources/
        ...
    Embedded/
        project.yml
        Sources/
            ...
```

In this structure, a target declared in `Embedded/project.yml` would have to use the path `Embedded/Sources` to reference the `Sources` directory adjacent to it.

### The Fix

The first major change is that I've introduced a `Spec` struct that references the directory containing the spec file, the `JSONDictionary` content of the spec file, and any _included_ spec files.

```swift
public struct Spec {
    public let relativePath: Path
    public let jsonDictionary: JSONDictionary
    public let subSpecs: [Spec]
}
```

This struct replaces the `Project.loadDictionary(path:)` function used before, and the `Spec` struct itself is used in the initialization of a `Project`, where previously a raw `JSONDictionary` was used.

The second major change is the introduction of a `PathContainer` protocol to abstract away the process of changing a path in an embedded file.

```swift
protocol PathContainer {
    static var pathProperties: [PathProperty] { get }
}

enum PathProperty {
    case string(String)
    case dictionary([PathProperty])
    case object(String, [PathProperty])
}
```

This is implemented by the model types (e.g. `Target`) to convert any dictionary keys that reference paths relative to the spec file. It is then used during initialization of a `Project` before collapsing the tree of `Spec` objects into a single `JSONDictionary`.

### DIFF
--- a/Docs/ProjectSpec.md
+++ b/Docs/ProjectSpec.md
@@ -45,7 +45,22 @@ Required properties are marked with checkbox. Some of the YAML examples don't sh
 
 One or more specs can be included in the project spec. This can be used to split your project spec into multiple files, for easier structuring or sharing between multiple specs. Included specs can also include other specs and so on.
 
-Include can either be a list of string paths or a single string path. They will be merged in order and then the current spec will be merged on top.
+Include can either be a list of includes or a single include. They will be merged in order and then the current spec will be merged on top.
+
+An include can be provided via a string (the path) or an object of the form:
+
+**Include Object**
+
+- [x] **path**: **String** - The path to the included file.
+- [ ] **useRelativePaths**: **Bool** - Dictates whether the included spec specifies paths relative to itself (the default) or the root spec file.
+
+```yaml
+include:
+  - includedFile.yml
+  - path: path/to/includedFile.yml
+    useRelativePaths: false
+```
+
 By default specs are merged additively. That is for every value:
 
 - if existing value and new value are both dictionaries merge them and continue down the hierachy
--- a/Docs/ProjectSpec.md
+++ b/Docs/ProjectSpec.md
@@ -52,13 +52,13 @@ An include can be provided via a string (the path) or an object of the form:
 **Include Object**
 
 - [x] **path**: **String** - The path to the included file.
-- [ ] **useRelativePaths**: **Bool** - Dictates whether the included spec specifies paths relative to itself (the default) or the root spec file.
+- [ ] **relativePaths**: **Bool** - Dictates whether the included spec specifies paths relative to itself (the default) or the root spec file.
 
 ```yaml
 include:
   - includedFile.yml
   - path: path/to/includedFile.yml
-    useRelativePaths: false
+    relativePaths: false
 ```
 
 By default specs are merged additively. That is for every value:
--- a/Sources/ProjectSpec/AggregateTarget.swift
+++ b/Sources/ProjectSpec/AggregateTarget.swift
@@ -1,6 +1,6 @@
 import Foundation
 import JSONUtilities
-import struct PathKit.Path
+import PathKit
 
 public struct AggregateTarget: ProjectTarget {
     public var name: String
--- a/Sources/ProjectSpec/AggregateTarget.swift
+++ b/Sources/ProjectSpec/AggregateTarget.swift
@@ -1,5 +1,6 @@
 import Foundation
 import JSONUtilities
+import struct PathKit.Path
 
 public struct AggregateTarget: ProjectTarget {
     public var name: String
@@ -59,5 +60,19 @@ extension AggregateTarget: NamedJSONDictionaryConvertible {
         buildScripts = jsonDictionary.json(atKeyPath: "buildScripts") ?? []
         scheme = jsonDictionary.json(atKeyPath: "scheme")
         attributes = jsonDictionary.json(atKeyPath: "attributes") ?? [:]
+    }
+}
+
+extension AggregateTarget: PathContaining {
+
+    static func expandPaths(for source: [String: JSONDictionary], relativeTo path: Path) -> [String: JSONDictionary] {
+        var result = source
+
+        for (targetName, var target) in result {
+            target = expandStringPaths(from: target, forKey: "configFiles", relativeTo: path)
+            target = expandChildPaths(from: target, forKey: "buildScripts", relativeTo: path, type: BuildScript.self)
+            result[targetName] = target
+        }
+        return result
     }
 }
--- a/Sources/ProjectSpec/AggregateTarget.swift
+++ b/Sources/ProjectSpec/AggregateTarget.swift
@@ -1,6 +1,5 @@
 import Foundation
 import JSONUtilities
-import PathKit
 
 public struct AggregateTarget: ProjectTarget {
     public var name: String
--- a/Sources/ProjectSpec/AggregateTarget.swift
+++ b/Sources/ProjectSpec/AggregateTarget.swift
@@ -63,7 +63,7 @@ extension AggregateTarget: NamedJSONDictionaryConvertible {
     }
 }
 
-extension AggregateTarget: PathContaining {
+extension AggregateTarget: PathContainer {
 
     static func expandPaths(for source: [String: JSONDictionary], relativeTo path: Path) -> [String: JSONDictionary] {
         var result = source
--- a/Sources/ProjectSpec/AggregateTarget.swift
+++ b/Sources/ProjectSpec/AggregateTarget.swift
@@ -65,14 +65,12 @@ extension AggregateTarget: NamedJSONDictionaryConvertible {
 
 extension AggregateTarget: PathContainer {
 
-    static func expandPaths(for source: [String: JSONDictionary], relativeTo path: Path) -> [String: JSONDictionary] {
-        var result = source
-
-        for (targetName, var target) in result {
-            target = expandStringPaths(from: target, forKey: "configFiles", relativeTo: path)
-            target = expandChildPaths(from: target, forKey: "buildScripts", relativeTo: path, type: BuildScript.self)
-            result[targetName] = target
-        }
-        return result
+    static var pathProperties: [PathProperty] {
+        return [
+            .dictionary([
+                .string("configFiles"),
+                .object("buildScripts", BuildScript.pathProperties),
+            ]),
+        ]
     }
 }
--- a/Sources/ProjectSpec/Project.swift
+++ b/Sources/ProjectSpec/Project.swift
@@ -168,6 +168,30 @@ extension Project {
     }
 }
 
+extension Project: PathContaining {
+
+    static func expandPaths(for spec: Spec, relativeTo basePath: Path = "") -> Spec {
+        let relativePath = (basePath + spec.relativePath).normalize()
+        guard relativePath != Path() else {
+            return spec
+        }
+        
+        var jsonDictionary = spec.jsonDictionary
+
+        jsonDictionary = expandStringPaths(from: jsonDictionary, forKey: "configFiles", relativeTo: relativePath)
+        jsonDictionary = expandChildPaths(from: jsonDictionary, forKey: "options", relativeTo: relativePath, type: SpecOptions.self)
+        jsonDictionary = expandChildPaths(from: jsonDictionary, forKey: "targets", relativeTo: relativePath, type: Target.self)
+        jsonDictionary = expandChildPaths(from: jsonDictionary, forKey: "aggregateTargets", relativeTo: relativePath, type: AggregateTarget.self)
+        return Spec(
+            relativePath: spec.relativePath,
+            jsonDictionary: jsonDictionary,
+            subSpecs: spec.subSpecs.map { template in
+                return Project.expandPaths(for: template, relativeTo: relativePath)
+            }
+        )
+    }
+}
+
 extension Project {
 
     public var allFiles: [Path] {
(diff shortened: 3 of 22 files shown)


ellneal:
CC #385

Sources/ProjectSpec/AggregateTarget.swift
```diff
@@ -1,5 +1,6 @@
 import Foundation
 import JSONUtilities
+import struct PathKit.Path
```
yonaskolb: Are there any naming conflicts here, that don't allow importing the whole module?

Sources/ProjectSpec/AggregateTarget.swift
```diff
@@ -61,3 +62,17 @@ extension AggregateTarget: NamedJSONDictionaryConvertible {
...
 }
+
+extension AggregateTarget: PathContaining {
+
+    static func expandPaths(for source: [String: JSONDictionary], relativeTo path: Path) -> [String: JSONDictionary] {
+        var result = source
+
+        for (targetName, var target) in result {
+            target = expandStringPaths(from: target, forKey: "configFiles", relativeTo: path)
+            target = expandChildPaths(from: target, forKey: "buildScripts", relativeTo: path, type: BuildScript.self)
```
yonaskolb: What if this `expandPaths` just returned something like `[PathProperty]` enum. Then it's a bit more declarative. And maybe `PathContainer` is good name?

Hand typed code (might not actually work):

```swift
struct PathProperty {
	let name: String
	let type: PropertyType

	enum PropertyType {
 		case string
 		case array
 		case object(PathContaining.Type) 
	}
}
```

```swift
extension AggregateTarget: PathContainer {
	static var pathProperties: [PathProperty] {
		return [
			PathProperty(name: "configFiles", type: .array),
			PathProperty(name: "buildScripts", type: .object(BuildScript.self)),
		]
	}
}
```

Then the spec loader can loop through all of these and do the actual merging

Sources/ProjectSpec/Project.swift
```diff
@@ -132,9 +132,18 @@ extension Project: Equatable {
 
 extension Project {
 
+    @available(*, deprecated, message: "Use `Project.Spec` and `init(spec:basePath:)`.")
```
yonaskolb: Do these functions need to be deprecated? They will probably still be the main entry point. Consumers shouldn't have to care about `Spec`, that's just an implementation detail (Can even be an `internal` struct)

Sources/ProjectSpec/ProjectSpec.swift
```diff
@@ -0,0 +1,90 @@
...
+
+internal protocol SubSequenceConvertible: Sequence {
+
+    init(_ subsequence: SubSequence)
+}
+
+extension String: SubSequenceConvertible { }
+extension Substring: SubSequenceConvertible { }
+
+extension Dictionary where Key: StringProtocol, Key: SubSequenceConvertible, Value: Any {
```
yonaskolb: Can you explain the reason for the `SubSequenceConvertible` protocol?

Sources/ProjectSpec/SpecLoader.swift
```diff
@@ -4,45 +4,59 @@ import PathKit
...
+    
     public init(path: Path) throws {
-        let dictionary = try Project.loadDictionary(path: path)
-        try self.init(basePath: path.parent(), jsonDictionary: dictionary)
+        let basePath = path.parent()
+        let template = try Spec(filename: path.lastComponent, basePath: basePath)
+        try self.init(spec: template, basePath: basePath)
     }
 
+    @available(*, deprecated, message: "Use `Project.Spec` for loading files from disk.")
```
yonaskolb: Same comment as above

Sources/ProjectSpec/AggregateTarget.swift
ellneal: No, just me letting NSHipster articles get into my head 😅

Sources/ProjectSpec/AggregateTarget.swift
ellneal: That's a much more elegant implementation than mine (I was initially merging it with the `resolveProject(_:)` function but ended up breaking it out, so it looks similar). I'll refactor it.

Sources/ProjectSpec/Project.swift
ellneal: The reason I deprecated these functions is that they will continue to function in the legacy behaviour, i.e. they won't respect the relative paths of included specs. I initially erased them completely but didn't want to break the public API so put them back with deprecation. Your call though obviously.

Sources/ProjectSpec/ProjectSpec.swift
ellneal: It's to do with `String` and `Substring`, and stripping the `:REPLACE` suffix from the keys. If the `Dictionary` extension is bound to `StringProtocol: Any`, and you perform this:

```swift
let newKey = key[key.startIndex ..< key.index(key.endIndex, offsetBy: -8)]
merged[newKey] = value
```

You get a compiler error:

> Cannot subscript a value of type '[Dictionary<Key, Value>.Key : Dictionary<Key, Value>.Value]' (aka 'Dictionary<Key, Value>') with an index of type 'Key.SubSequence'

The way to convert a `Substring` to a `String` is simply `String(subtring)`. The `SubSequenceConvertible` protocol just describes that conversion and makes the above code work via:

```swift
let newKey = key[key.startIndex ..< key.index(key.endIndex, offsetBy: -8)]
merged[Key(newKey)] = value
```

Note that it could have been avoided by making the extension `where Key == String, Value: Any`

ellneal:
Thanks @yonaskolb. I would hazard a guess that this behaviour is what most people would expect, so I do think that any `option` param should be to opt into the legacy behaviour, rather than the other way around. Although I agree that it's a breaking change, so perhaps this should go out in a minor version bump rather than a patch version bump? I'll get to work implementing some of your feedback.

Sources/ProjectSpec/AggregateTarget.swift
ellneal: Actually generics is making this rather complicated, I'll have to give it some more thought.

> Protocol 'PathContainer' can only be used as a generic constraint because it has Self or associated type requirements

yonaskolb: Ahh I see. Then the .object case can just have an array of `[PathProperty]` instead of `PathContainer`

ellneal: I've managed to refactor it with just an enum:

```swift
enum PathProperty {
    case string(String)
    case dictionary([PathProperty])
    case object(String, [PathProperty])
}
```

The expansion is then handled recursively in `Project.expandPaths(_:in:relativeTo:)`.

ellneal:
@yonaskolb I've added a new option when including files to provide the legacy behaviour of not expanding paths.

Docs/ProjectSpec.md
```diff
@@ -45,7 +45,22 @@ Required properties are marked with checkbox. Some of the YAML examples don't sh
...
 
-Include can either be a list of string paths or a single string path. They will be merged in order and then the current spec will be merged on top.
+Include can either be a list of includes or a single include. They will be merged in order and then the current spec will be merged on top.
+
+An include can be provided via a string (the path) or an object of the form:
+
+**Include Object**
+
+- [x] **path**: **String** - The path to the included file.
+- [ ] **useRelativePaths**: **Bool** - Dictates whether the included spec specifies paths relative to itself (the default) or the root spec file.
```
yonaskolb: What do you think of just `relativePaths`? The names have been trying to stay concise and without `use`, `has` or `is` prefixes

Sources/ProjectSpec/AggregateTarget.swift
```diff
@@ -61,3 +62,15 @@ extension AggregateTarget: NamedJSONDictionaryConvertible {
         attributes = jsonDictionary.json(atKeyPath: "attributes") ?? [:]
     }
 }
+
+extension AggregateTarget: PathContainer {
+
+    static var pathProperties: [PathProperty] {
+        return [
+            .dictionary([
```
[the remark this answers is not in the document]
ellneal: It's for layouts like _Target_, where it's a dictionary keyed by the name of the target. It's the difference between `[String: JSONDictionary]` and `[JSONDictionary]`.

```swift
case .dictionary(let pathProperties):
    for (key, dictionary) in result {
        if let source = dictionary as? JSONDictionary {
             result[key] = expandPaths(pathProperties, in: source, relativeTo: path)
        }
    }
```

Sources/ProjectSpec/ProjectSpec.swift
```diff
@@ -0,0 +1,102 @@
+import Foundation
+import typealias JSONUtilities.JSONDictionary
+import PathKit
+
+extension Project {
+
+    public struct Spec {
```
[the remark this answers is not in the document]
ellneal: It is in its own file?

Sources/ProjectSpec/Project.swift
```diff
@@ -132,9 +132,18 @@ extension Project: Equatable {
...
     public init(basePath: Path, jsonDictionary: JSONDictionary) throws {
+        let spec = Project.Spec(relativePath: Path(), jsonDictionary: jsonDictionary)
+        try self.init(spec: spec, basePath: basePath)
+    }
+
+    public init(spec: Spec, basePath: Path) throws {
         self.basePath = basePath
-        let jsonDictionary = try Project.resolveProject(jsonDictionary: jsonDictionary)
+
+        let spec = Project.expandPaths(for: spec)
```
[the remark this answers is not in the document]
ellneal: You're right, that would be much better.

Sources/ProjectSpec/SpecOptions.swift
```diff
@@ -120,3 +121,14 @@ extension SpecOptions: JSONObjectConvertible {
         generateEmptyDirectories = jsonDictionary.json(atKeyPath: "generateEmptyDirectories") ?? false
     }
 }
+
+extension SpecOptions: PathContainer {
+
+    static var pathProperties: [PathProperty] {
+        return [
+            .string("carthageBuildPath"),
+            .string("carthageExecutablePath"),
```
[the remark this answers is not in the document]
ellneal: I had the same thought, but not the motivation to do it. 👍

Sources/ProjectSpec/ProjectSpec.swift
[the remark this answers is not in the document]
yonaskolb: I mean the `Spec` struct out of the `ProjectSpec` file and into a `Spec` file

ellneal:
My pleasure 😄

rpassis:
FWIW I just updated to `2.2.0` and noticed a regression due to this change.

Our project configuration structure has a `project.yml` on the root folder of our project, and that file includes a bunch of `targets` that are all configured as separate files under `xcodegen/*.yml`:

```
include:
  - xcodegen/version.yml
  - xcodegen/base.yml
  - xcodegen/myapp.yml
  - xcodegen/myapp-dev.yml
  - xcodegen/myframework.yml
  ... etc
```

When running `generate` with this version I get a bunch of spec validation errors similar to the below:

```
- Target "MyFramework" has a missing source directory "/Users/user/dev/MyApp/xcodegen/MyFramework"
```

It is obviously expecting the source directories to be relative to where the `yml` file is located.
Not sure if you guys were aware of it but I am glad there's an opt-out flag :)

yonaskolb:
Yeah it was a decision to make relative paths the default, as per the change log. Unfortunately this is a breaking change. You can opt out like this
```
include:
  - path: xcodegen/version.xml
    relativePaths: false
```
Would be great to know stats on which default people prefer.

brentleyjones:
While it was a pain to migrate, I prefer the new way by a long shot.

danielwilliamson:
Yeah same opinion here. It's rather a pain to migrate but I can see why it's needed.

DrewKiino:
amazing update. thank you.

PR #492, opened 2019-01-16
Fixes issue with sticker pack targets generating redundant compile sources phase

rpassis:
This PR is intended to address the issue of Sticker Pack targets getting generated with a redundant compile sources build phase.

I am not entirely happy with the testing for this here but when I started looking at testing against `pbxProject.sourcesBuildPhases` a big can of worms ensued 😬

Note: The rest of this description is (semi)unrelated to the PR so feel free to ignore.

If you take a look at the `testTarget()` test, on line 578 the code iterates over targets.

Inside that for loop I tried to add code to the effect of the below

```
for target in targets { 
    let nativeTarget = ...
    let buildPhases = nativeTarget.buildPhases
    let sourcesPhases = pbxProject.sourcesBuildPhases.filter { buildPhases.contains($0) }
}
```

For example in one instance of the for loop where `buildPhases.count == 3` the `sourcePhases.count` was returning `8`. I assume the source of the problem is due to how `hashValue` is implemented in `PBXObject.swift` but I haven't had the change to fully investigate it.

### DIFF
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -665,8 +665,13 @@ public class PBXProjGenerator {
         }
 
         let sourcesBuildPhaseFiles = getBuildFilesForPhase(.sources)
-        let sourcesBuildPhase = addObject(PBXSourcesBuildPhase(files: sourcesBuildPhaseFiles))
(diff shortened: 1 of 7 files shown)


yonaskolb:
Fantastic, thanks for fixing this @rpassis! Could you add a changelog entry as well?

yonaskolb:
No worries I'll update the changelog on master

PR #500, opened 2019-01-23
Adding input and output file lists to build script spec

### LINKED ISSUE #499: Plans to support xcfilelist?
Hey 👋, Is xcodegen planning to support the [recent Xcode 10 addition](https://developer.apple.com/documentation/xcode_release_notes/xcode_10_release_notes/build_system_release_notes_for_xcode_10#3035615) of `.xcfilelist` files for build scripts?

> Run Script build phases now support declaring input and output files in a .xcfilelist file. This file should contain a newline-seperated list of the file paths for the inputs or outputs. Build setting references may be used in these paths. The path to a copy of the .xcfilelist file with resolved build setting references will be provided in the SCRIPT_INPUT_FILE_LIST_(#) and SCRIPT_OUTPUT_FILE_LIST_(#) environment variables of the Run Script phase. Projects which use an .xcfilelist require Xcode 10. (39241255)

✌️

lukewakeford:
Either that was just super easy or I missed something important.

@yonaskolb let me know if this needs more work.

closes #499

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 #### Added
 - Added ability to generate empty directories via `options.generateEmptyDirectories` [#480](https://github.com/yonaskolb/XcodeGen/pull/480) @Beniamiiin
 - Added support for the `instrumentsPackage` product type [#482](https://github.com/yonaskolb/XcodeGen/pull/482) @ksulliva
+- Added support for `inputFileLists` and `outputFileLists` within project build scripts [#500](https://github.com/yonaskolb/XcodeGen/pull/500) @lukewakeford
 
 #### Fixed
 - Fixed `--project` argument not taking effect [#487](https://github.com/yonaskolb/XcodeGen/pull/487) @monowerker
--- a/Docs/ProjectSpec.md
+++ b/Docs/ProjectSpec.md
@@ -423,6 +423,8 @@ Each script can contain:
 - [ ] **name**: **String** - name of a script. Defaults to `Run Script`
 - [ ] **inputFiles**: **[String]** - list of input files
 - [ ] **outputFiles**: **[String]** - list of output files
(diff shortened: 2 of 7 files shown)


yonaskolb:
Awesome, thank you for the PR Luke! Yes, it is that easy 😄 

Just a couple of minor additions and we're good to go. Could you please add:
- a changelog entry under `Added`
- a decoding test in https://github.com/yonaskolb/XcodeGen/blob/master/Tests/XcodeGenKitTests/SpecLoadingTests.swift#L439
- the addition of at least one file list path to an existing build script in the `TestProject` fixture https://github.com/yonaskolb/XcodeGen/blob/master/Tests/Fixtures/TestProject/project.yml. (This is a broad integration test. A specific unit test for the generation isn't needed but feel free to add one if you feel like it)

If we wanted to add some validation about whether the paths exist that would be added to https://github.com/yonaskolb/XcodeGen/tree/master/Sources/ProjectSpec. In this case though those file lists are potentially themselves generated so failing validation because they don't exist at the time of project generation might be unwanted. What do you think?

lukewakeford:
I've made the changes you asked for, I'm not entirely sure how and when the integration test is run but i've added a couple of example `.xcfilelist` and committed the result in the TestProject.

I agree that validating the existence of those files when generating is not a good idea as they will, in most cases, be generated by a previous build script.

yonaskolb:
`TestProject` is generated during the tests as you noticed and checked in. On CI, https://github.com/yonaskolb/XcodeGen/blob/master/diff-fixtures.sh is run to check there are no changes and https://github.com/yonaskolb/XcodeGen/blob/master/build-fixtures.sh is run to make sure the project still builds.

lukewakeford:
That makes sense 👍 thanks for the pointers.

PR #514, opened 2019-02-12
Add missingFiles option to skip checking for config file existence

elliottwilliams:
In https://github.com/yonaskolb/XcodeGen/pull/220#pullrequestreview-86853813, @yonaskolb suggested that XcodeGen could have a validation setting to disable checking for the existence of config files:

> We could create a followup PR that disables `missingFiles`

I'm playing with using XcodeGen from a context where I don't have config files available, so I went ahead and added a `missingFiles` option to handle this use case.

One question I have is whether `missingFiles` should imply `missingConfigs`, or whether you think it makes sense for those two options to remain distinct. Let me know what you think!

### DIFF
--- a/Tests/XcodeGenKitTests/ProjectSpecTests.swift
+++ b/Tests/XcodeGenKitTests/ProjectSpecTests.swift
@@ -117,6 +117,13 @@ class ProjectSpecTests: XCTestCase {
                 project.configFiles = ["missingConfiguration": configPath.string]
                 try project.validate()
             }
+            
+            $0.it("allows non-existent config files") {
+                var project = baseProject
+                project.options = SpecOptions(disabledValidations: [.missingFiles, .missingConfigs])
+                project.configFiles = ["invalidConfig": "invalidConfigFile"]
+                try project.validate()
+            }
 
             $0.it("fails with invalid target") {
                 var project = baseProject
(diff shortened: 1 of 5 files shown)


yonaskolb:
Thanks for the PR @elliottwilliams! 
Perhaps `missingFiles` is a bit misleading as there are other sorts of files that are required and can cause validation errors, for example source files and dependencies)]. For this case, perhaps it should be called `missingConfigFiles`. In terms of the existing `missingConfigs` that is seperate and relates to actual config names.

Could you please add a changelog entry as well?

elliottwilliams:
That makes sense! Renamed to `missingConfigFiles`.

Tests/XcodeGenKitTests/ProjectSpecTests.swift
```diff
@@ -117,6 +117,13 @@ class ProjectSpecTests: XCTestCase {
                 project.configFiles = ["missingConfiguration": configPath.string]
                 try project.validate()
             }
+            
+            $0.it("allows non-existent config files") {
+                var project = baseProject
+                project.options = SpecOptions(disabledValidations: [.missingConfigFiles, .missingConfigs])
```
yonaskolb: Could we use just `missingConfigFiles`

elliottwilliams: Yeah, good catch!

PR #524, opened 2019-02-21
Compute relative paths to support sources outside a spec's directory

### LINKED ISSUE #318: Add sources outside project file's parent folder?
My project structure:

```
- ios_project
    - project.xcodeproj
    - example
- src
- generated_src
    - cpp
    - objc
```

I want include `example`, `../src`, `../generated_src/cpp`, `../generated_src/objc` into my project's sources, below is my `project.yml`:

``` yml
  AudioMixer:
    type: framework
    platform: iOS
    deploymentTarget: "9.0"
    sources:
      - example
      - ../src
      - ../generated_src/cpp
      - ../generated_src/objc
```

The problem is, those three groups `src`, `cpp`, and `objc` has wrong path value:

screen shot 2018-06-12 at 00 00 23

It's "Relative To Group", but has an absolute path value, which cause header search failure.

How could I fix it?

elliottwilliams:
Closes #318 

Previously, XcodeGen would determine the relative path to a source by removing the project's `basePath` from the source's path, but this doesn't work when the source's path is outside the base path. For example, given a base path `/a/b` and a path to a source `../c`, XcodeGen would unsuccessfully try to remove "/a/b" from "/a/c" and wind up creating a "relative" file reference to "/a/c".

This change adds a method to `Path` that computes the relative path between two paths, inspired by methods like [`Pathname#relative_path_from`][0] in Ruby, and uses it to determine the paths for top-level references in the project.

I'm trying to decide whether I should really be contributing this to PathKit, since this solves a generalizable path problem, but since it fixes a specific bug in XcodeGen I wanted to put it up here. I'm happy to open a PR on PathKit in addition to this one, so that you don't get stuck maintaining a pathing algorithm 🙂

[0]: http://ruby-doc.org/stdlib-2.6.1/libdoc/pathname/rdoc/Pathname.html#method-i-relative_path_from

### DIFF
--- a/Sources/XcodeGenKit/PathExtensions.swift
+++ b/Sources/XcodeGenKit/PathExtensions.swift
@@ -2,11 +2,6 @@ import Foundation
 import PathKit
 
 extension Path {
-
-    public func byRemovingBase(path: Path) -> Path {
-        return Path(normalize().string.replacingOccurrences(of: "\(path.normalize().string)/", with: ""))
-    }
-
     /// Returns a Path without any inner parent directory references.
     ///
     /// Similar to `NSString.standardizingPath`, but works with relative paths.
--- a/Sources/XcodeGenKit/PathExtensions.swift
+++ b/Sources/XcodeGenKit/PathExtensions.swift
@@ -7,6 +7,11 @@ extension Path {
         return Path(normalize().string.replacingOccurrences(of: "\(path.normalize().string)/", with: ""))
     }
     
+    /// Returns the relative path necessary to go from `base` to `self`.
+    ///
+    /// Both paths must be absolute or relative paths.
+    /// - throws: Throws an error when the path types do not match, or when `base` has so many parent path components
+    ///           that it refers to an unknown parent directory.
     public func relativePath(from base: Path) throws -> Path {
         enum PathArgumentError: Error {
             /// Can't back out of an unknown parent directory
@@ -17,11 +22,11 @@ extension Path {
         
         func pathComponents(for path: ArraySlice<String>, relativeTo base: ArraySlice<String>, memo: [String]) throws -> [String] {
             switch (base.first, path.first) {
-            // 1. Paths are equivalent
+            // Base case: Paths are equivalent
             case (.none, .none):
                 return memo
                 
-            // 2. No path to backtrack from
+            // No path to backtrack from
             case (.none, .some(let rhs)):
                 guard rhs != "." else {
                     // Skip . instead of appending it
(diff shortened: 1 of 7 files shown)


Sources/XcodeGenKit/PathExtensions.swift
```diff
@@ -6,4 +6,79 @@ extension Path {
...
+                                                   relativeTo: ArraySlice(base.normalize().components.strippingParentDirectoryReferences()),
+                                                   memo: []))
+    }
+}
+
+private extension Array where Element == String {
+    /// Removes inner `..`s from an array of path components.
+    ///
+    /// Foundation does this in `NSString.standardizingPath`, but only for absolute paths.
+    func strippingParentDirectoryReferences() -> [Element] {
```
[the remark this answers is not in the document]
elliottwilliams: I reimplemented this method in `b5dcc9c` to be a lot simpler (turns out PathKit's + operator can simplify parent references for us), but added some examples.

Sources/XcodeGenKit/SourceGenerator.swift
```diff
@@ -254,7 +254,7 @@ class SourceGenerator {
 
             let groupName = name ?? path.lastComponent
             let groupPath = isTopLevelGroup ?
-                path.byRemovingBase(path: project.basePath).string :
+                ((try? path.relativePath(from: project.basePath)) ?? path).string :
```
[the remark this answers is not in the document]
elliottwilliams: Sure!

Sources/XcodeGenKit/SourceGenerator.swift
```diff
@@ -474,7 +474,7 @@ class SourceGenerator {
         var sourcePath = path
         switch type {
         case .folder:
-            let folderPath = Path(targetSource.path)
+            let folderPath = project.basePath + Path(targetSource.path)
```
elliottwilliams: I had to change this line to fix a test that broke when I swapped out `byRemovingBase(path:)` with `relativePath(from:)`.

It seems like folder target sources were being created without the base directory as part of the path. In the tests, I saw that a folder "Sources/A" would passed to `getFileReference` as `Path("Sources/A")`, but a file "Sources/a.swift" would be passed as `Path("TestDirectory/Sources/a.swift")`.

elliottwilliams:
Thanks! It was fun to write :) Changelog updated.

Sources/XcodeGenKit/PathExtensions.swift
```diff
@@ -6,6 +6,18 @@ extension Path {
...
+    /// Returns a Path without any inner parent directory references.
+    ///
+    /// Similar to `NSString.standardizingPath`, but works with relative paths.
+    ///
+    /// ### Examples
+    /// - `a/b/../c` simplifies to `a/c`
+    /// - `../a/b` simplifies to `../a/b`
+    /// - `a/../../c` simplifies to `../c`
+    public func simplifyingParentDirectoryReferences() -> Path {
+        return normalize().components.reduce(Path(), +)
```
yonaskolb: Does normalize just do the same thing? What does adding the components together again do?

elliottwilliams: normalize calls [`NSString.standardizingPath`](https://developer.apple.com/documentation/foundation/nsstring/1407194-standardizingpath#), which _almost_ does the same thing but doesn't remove parent directory references in relative paths: 

> **For absolute paths only**, resolving references to the parent directory (that is, the component “..”) to the real parent directory if possible using resolvingSymlinksInPath. For relative paths, references to the parent directory are left in place.

PathKit's + operator backs up a level if you try to add `..` to a path, which means it can handle this normalization edge case for us. In other words:

```swift
Path("foo/../bar").normalize() == Path("foo/../bar")
Path("foo") + ".." + "bar" == Path("bar")
```

Issue #521, opened 2019-02-19
Target template path resolution


haritowa:
Hi, I’m trying to organize a monorepo project using XcodeGen. My general module structure looks like:
```
.
├── Core
│   ├── Sources
│   │   └── CoreClass.swift
│   ├── TestData
│   ├── Tests
│   └── project.yml
├── SomeFeature
│   ├── Sources
│   │   └── SomeFeature.swift
│   ├── TestData
│   ├── Tests
│   └── project.yml
└── baseProject.yml
```

Where `Sources` is lib sources, `TestData` contains mocks and stubs, `Tests` contains tests and so on. Each module will have several development targets(main, tests, test data, etc), so I’m trying to use target templates feature to formalize and automate development process. 

```yaml
name: BaseProject
targetTemplates:
  Core:
    platform: iOS
    type: framework
    info:
      path: $target_name_info.plist
    sources: Sources
    scheme:
      testTargets:
        - $target_nameUnitTests
      gatherCoverageData: true
  CoreUnitTests:
    platform: iOS
    type: bundle.unit-test
    sources: Tests
```

All templates are described in the `baseProject.yml` file(which is located in the root folder). Module spec is located in the module directory(`Core/project.yml`; `Feature/Auth/project.yml`, for example). Each module spec includes baseProject, so it looks like:

```yaml
name: Core
include:
  - path: ../baseProject.yml
    relativePaths: false
targets:
  Core:
    templates: 
      - Core
  CoreUnitTests:
    templates:
      - CoreUnitTests
```

Unfortunately, target templates are resolved after the path resolution process, so all paths are always the same, no matter where base spec is included from.

For example, let’s take a project with two modules: `A` and `B`. `B` depends on `A` ([project.zip](https://github.com/yonaskolb/XcodeGen/files/2880250/project.zip)).

```
.
├── A
│   ├── Sources
│   │   └── AClass.swift
│   └── project.yml
├── B
│   ├── Sources
│   │   └── BClass.swift
│   └── project.yml
└── baseProject.yml
```

XcodeGen for `B` spec (`xcodegen -s B/project.yml`) ends up with a messed project, where both `A` and `B` Sources folder is `B/Sources`.

![](https://cl.ly/fae3baf4589c/%255B8036a1889db031c1dc73c59c662fdd88%255D_Image%252525202019-02-19%25252520at%252525204.41.45%25252520PM.png)

[...]

PR #529, opened 2019-02-28
Fix target template path resolution

haritowa:
This PR fixes target template path resolution https://github.com/yonaskolb/XcodeGen/issues/521. 

It runs `resolvingPaths` on the target template merged dictionary. Unfortunately, I see no clear way to pass info about spec path(which target belongs to), so, as a POC I integrate relative path into targets JSON dictionary.

### DIFF
--- a/Sources/ProjectSpec/SpecFile.swift
+++ b/Sources/ProjectSpec/SpecFile.swift
@@ -94,6 +94,18 @@ public struct SpecFile {
                 .reduce([:]) { $1.merged(onto: $0) })
     }
 
+    private func setTargetsRelativePath(projectJSON: JSONDictionary, relativePath: Path) -> JSONDictionary {
+        guard var targets = projectJSON["targets"] as? [String: JSONDictionary] else {
+            return projectJSON
+        }
+
+        var result = projectJSON
+        targets.keys.forEach { targets[$0]?.updateValue(relativePath.string, forKey: "relative_path") }
+        result["targets"] = targets
+        
+        return result
+    }
+    
     func resolvingPaths(relativeTo basePath: Path = Path()) -> SpecFile {
         let relativePath = (basePath + self.relativePath).normalize()
         guard relativePath != Path() else {
@@ -101,8 +113,9 @@ public struct SpecFile {
         }
 
         let jsonDictionary = Project.pathProperties.resolvingPaths(in: self.jsonDictionary, relativeTo: relativePath)
+        
         return SpecFile(
-            jsonDictionary: jsonDictionary,
+            jsonDictionary: setTargetsRelativePath(projectJSON: jsonDictionary, relativePath: relativePath),
             relativePath: self.relativePath,
             subSpecs: subSpecs.map { $0.resolvingPaths(relativeTo: relativePath) }
         )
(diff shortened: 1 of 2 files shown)


yonaskolb:
Does this merged PR address your issue at all? https://github.com/yonaskolb/XcodeGen/pull/537

haritowa:
I'll run new version against my current setup, but, my fix addresses not only sources but other template paths too https://github.com/yonaskolb/XcodeGen/pull/529/files#diff-867759754eec9b4f3a17977bdaa3a3c2R117

haritowa:
@yonaskolb nope, and it "breaks" my "fix" :)

yonaskolb:
Could you rebase this off master, and see if you can support the new tests that come with it.

haritowa:
Huge sorry for this delay. I have tried the rebased version and found out why my “fix” is not working anymore: you have updated path resolving routine in the 32b3d8c, so now target templates are resolved like plain targets. While it is a step toward my goals, it made my approach impossible to implement. Let me describe, what behavior I want to achieve and how it differs from the current implementation.

Now target templates have two path resolution strategies, depending on `relativePaths` `include` value. Let’s use my sample project as an example:

```
.
├── A
│   ├── Sources
│   │   └── AClass.swift
│   └── project.yml
├── B
│   ├── Sources
│   │   └── BClass.swift
│   └── project.yml
└── baseProject.yml

4 directories, 5 files
```

*baseProject:*
```yaml
name: BaseProject
targetTemplates:
  Core:
    platform: iOS
    type: framework
    sources: Sources
```

*A/project.yml:*
```yaml
name: A
include:
  - ../baseProject.yml
targets:
  A:
    templates: 
      - Core
```

*A/project.yml:*
```yaml
name: B
include: 
  - ../baseProject.yml
  - ../A/project.yml
targets:
  B:
    templates: 
      - Core
    dependencies:
      - target: A
```

## Current + relativePaths: true
If we run _xcodegen_ against `A/project.yml`, we will receive a _Core_ template with `sources` path equals to `../Sources`. Unfortunately, this result is not useful for my case. During merge _A_ target receives `sources: ../Sources`.

So, `include` with `relativePaths: true` *declares a target part, which is defined relative to base file*(i.e., while we declare this kind of target, we are assuming, that there will be `Sources` folder in the same folder, as `baseproject.yml` is). We may use a target name as a parameter, but this doesn’t allow us to build nested sub-targets(`Root/Core/Common` have to be  `Root/CoreCommon`).

## Current + relativePath: false
Now let’s change `relativePaths ` to `false` in the  `A/project.yml`. The result is _Core_ template with `sources` path equals to `../A/Sources`.   During merge _A_ target receives `sources: ../A/Sources`. This looks better because now the project is buildable and works appropriately.

So, `include` with `relativePaths: false` *declares a target part, which is defined relative to including file*. Unfortunately, this is done during project parsing, which squashes a lot of data. So, if we will run _xcodegen_ against the `B/project.yml` we will end up with *one* `Core` template with sources: `../A/Sources`.  Now we have two targets with the same(wrong) list of Sources.



## Desired behavior
I need a mode, where a target template is described in terms of including target but is unique for each target. My proposal is to add new mode or fix `Current + relativePath: false` to preserve original target template JSON data until `resolveTargetTemplates` method invoked.

Target templates seem broken now, if you include spec from several specs and at least one have `relativePath: false`(as a side effect of data merges).

I can create new POC with either new mode or updated `relativePath: false` mechanics.

PR #530, opened 2019-03-01
Feature: Add path support to Scheme.ExecutionAction

macbellingrath:
## Background
I have a use-case to add a large shell script (50-100 lines) as a `postAction` for one of my schemes. I have made an initial attempt to add it using the existing `PathType` from `BuildScript`. 

Open to your feedback. Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-g<p align="center">
+<p align="center">
 <a href="https://github.com/yonaskolb/XcodeGen">
 <img src="Assets/Logo_animated.gif" alt="XcodeGen" />
 </a>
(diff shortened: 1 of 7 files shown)


macbellingrath:
I've since realized that multiline and comment-containing scripts are not supported with this change since `BuildScript` is converted into `PBXShellScriptBuildPhase` which handles this with its internal type `CommentedString`.

alternatively, Xcode accepts a path as well. I may just use something like this:

```yaml
schemes:
  MyFramework:
    build:
      parallelizeBuild: true
      buildImplicitDependencies: true
      targets:
        MyFramework: all
    archive:
      targets:
        MyFramework: all
      postActions:
        - script: $PROJECT_DIR/scripts/myframework_post_archive.sh
```

yonaskolb:
Thanks for the PR @macbellingrath, but yes referencing an external script is probably the best way to solve this 👍

PR #532, opened 2019-03-05
use link property in framework deps

dimatosaurus:
Allows using link property in framework dependencies. This is needed
if you use firebase as a framework dependency.

### DIFF
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -518,11 +518,15 @@ public class PBXProjGenerator {
                     )
                 }
 
-                let buildFile = addObject(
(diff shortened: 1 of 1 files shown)


brentleyjones:
We use Firebase at our company, and we don't need this change. Can you give a fuller example of how this is broken without this change?

dimatosaurus:
We use firebase as framework dependency (not Carthage) in a framework that we ship
with our app. Linking the firebase libs prohibits the upload to the app store, since no 
Bitcode is included in the firebase libraries. Linking is also not necessary here.

The change is similar to the Carthage change that was posted a while ago. Since we
do not use Carthage to fetch the framework, we needed a similar behavior.

yonaskolb:
Thanks @dimatosaurus. At first glance this seems to check out. Look good to you @brentleyjones?

@dimatosaurus could you add a changelog entry as well?

yonaskolb:
Added changelog in master

PR #533, opened 2019-03-13
Add ability to define `templateAttributes` within a target

tomquist:
This allows parameterizing templates. One specific problem I wanted to solve with this issue is that I have a multi-framework project where framework-targets are specified by a target template. Each framework consists of the framework itself and a corresponding test target. The framework template is specified like this:
```yaml
targetTemplates:
  Module:
    platform: iOS
    type: framework
    sources: [modules/$target_name/Sources]
```

Now I'd like to create a template for the test target where the test-files are placed in `modules/<ModuleName>/Tests`. I have no way of defining a common template for this.

When this PR is merged, I'm able to define the following template:
```yaml
targetTemplates:
  ModuleTests:
    platform: iOS
    type: bundle.unit-test
    sources: [modules/$targetUnderTest/Tests]
    dependencies:
      target: $targetUnderTest
```
and use it like this:
```yaml
targets:
  MyFramework:
    templates: [Module]
  MyFrameworkTests:
    templates: [ModuleTests]
    templateAttributes:
      targetUnderTest: MyFramework
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
 - Added ability to automatically include Carthage related dependencies via `includeRelated: true` [#506](https://github.com/yonaskolb/XcodeGen/pull/506) @rpassis
 - Added ability to define a per-platform `deploymentTarget` for Multi-Platform targets. [#510](https://github.com/yonaskolb/XcodeGen/pull/510) @ainopara
 - Added support for nested target templates [#534](https://github.com/yonaskolb/XcodeGen/pull/534) @tomquist
+- Added ability to define `templateAttributes` within a target to be able to parameterize templates. [#533](https://github.com/yonaskolb/XcodeGen/pull/533) @tomquist
+
+#### Changed
+- **DEPRECATION**: Placeholders `$target_name` and `$platform` have been deprecated in favour of `${target_name}` and `${platform}`. Support for the old placeholders will be removed in a future version [#533](https://github.com/yonaskolb/XcodeGen/pull/533) @tomquist
 
 #### Fixed
 - Sources outside a project spec's directory will be correctly referenced as relative paths in the project file. [#524](https://github.com/yonaskolb/XcodeGen/pull/524)
--- a/Docs/ProjectSpec.md
+++ b/Docs/ProjectSpec.md
@@ -203,7 +203,8 @@ Settings are merged in the following order: groups, base, configs.
   - `CFBundleVersion`
   - `CFBundlePackageType`
 - [ ] **entitlements**: **[Plist](#plist)** - If defined this will generate and write a `.entitlements` file, and use it by setting `CODE_SIGN_ENTITLEMENTS` build setting for every configuration. All properties must be provided
-- [ ] **templates**: **[String]** - A list of [Target Templates](#target-template) referenced by name that will be merged with the target in order. Any instances of `$target_name` within these templates will be replaced with the target name.
+- [ ] **templates**: **[String]** - A list of [Target Templates](#target-template) referenced by name that will be merged with the target in order. Any instances of `${target_name}` within these templates will be replaced with the target name.
+- [ ] **templateAttributes**: **[String: String]** - A list of attributes where each instance of `${attributeName}` within the templates listed in `templates` will be replaced with the value specified.
 - [ ] **transitivelyLinkDependencies**: **Bool** - If this is not specified the value from the project set in [Options](#options)`.transitivelyLinkDependencies` will be used.
 - [ ] **directlyEmbedCarthageDependencies**: **Bool** - If this is `true` Carthage dependencies will be embedded using an `Embed Frameworks` build phase instead of the `copy-frameworks` script. Defaults to `true` for all targets except iOS/tvOS/watchOS Applications.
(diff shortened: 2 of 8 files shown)


AlexisQapa:
What about going even further and define associated templates for tests into the tested target template ? This would give us the ability to define only targets of interest.

targetTemplates:
  Module:
    platform: iOS
    type: app
    sources: [modules/sources]
    tests:
        template: ModuleTest

use
targets:
  MyFramework:
    templates: [Module]

tomquist:
I think my solution solves the issue in a more general way and would also be useful for many other use cases. I agree that it would also be a good idea to have some way to automatically generate an associated test target for a framework target but this can still be done separately.

AlexisQapa:
That's what I thoughts I'll give this a try when I can and give you feedbacks

Docs/ProjectSpec.md
```diff
@@ -580,21 +581,24 @@ This is used to override settings or run build scripts in specific targets
 
 This is a template that can be referenced from a normal target using the `templates` property. The properties of this template are the same as a [Target](#target)].
 Any instances of `$target_name` within each template will be replaced by the final target name which references the template.
+Attributes defined in `templateAttributes` will be replaced by the value specified in the target which references the template.
```
yonaskolb: The `templateAttributes` are defined in the target, not the template. This sentence doesn't make that so clear.

tomquist: I'm not a native speaker. Do you have a suggestion for a better sentence?

yonaskolb: Perhaps `Any attributes defined within a targets `templateAttributes` will be used to replace any attribute references in the template using the syntax `$attribute_name`.

tomquist:
Regarding your suggestion to use `${attributeName}` instead of `$attributeName`: I totally agree but didn't want to make the change without discussing first. I'll try to implement that. Ideally xcodegen would show a warning when `$target_name` or `$platform` is used to migrate to `${target_name}` and `${platform}`
Is there anything I can reuse to generate such warnings?

yonaskolb:
Warnings would come from a linter which is not yet implemented and tracked here https://github.com/yonaskolb/XcodeGen/issues/214

tomquist:
I did the necessary change and also added warnings for old usage of `$target_name` and `$platform`. @yonaskolb Could you please have a look at this? I hope my solution is fine for now until it is replaced by the more sophisticated linter solution.

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@
 #### Added
 - Added `missingConfigFiles` to `options.disabledValidations` to optionally skip checking for the existence of config files.
 - Added ability to automatically include Carthage related dependencies via `includeRelated: true` [#506](https://github.com/yonaskolb/XcodeGen/pull/506) @rpassis
+- Added ability to define `templateAttributes` within a target to be able to parameterize templates. [#533](https://github.com/yonaskolb/XcodeGen/pull/533) @tomquist
+
+#### Changed
+- **DEPRECATION**: Placeholders `$target_name` and `$platform` changed to `${target_name}` and `${platform}` in order to streamline placeholders with `templateAttributes`. [#533](https://github.com/yonaskolb/XcodeGen/pull/533) @tomquist
```
```suggestion
- **DEPRECATION**: Placeholders `$target_name` and `$platform` have been deprecated in favour of `${target_name}` and `${platform}`. Support for the old placeholders will be removed in a future version [#533](https://github.com/yonaskolb/XcodeGen/pull/533) @tomquist
```

Sources/ProjectSpec/SpecValidation.swift
```diff
@@ -194,3 +195,36 @@ extension Project {
         }
     }
 }
+
+extension Dictionary where Key == String, Value: Any {
+
+    public func validateWarnings() throws {
+        var errors: [SpecValidationError.ValidationError] = []
```
yonaskolb: Could we move the contents of this function to `SpecLoader.validateProjectDictionaryWarnings`. This doesn't need to be on every dictionary

tomquist: I moved it into a private extension within `SpecLoader`.

Docs/ProjectSpec.md
```diff
@@ -203,7 +203,8 @@ Settings are merged in the following order: groups, base, configs.
   - `CFBundleVersion`
   - `CFBundlePackageType`
 - [ ] **entitlements**: **[Plist](#plist)** - If defined this will generate and write a `.entitlements` file, and use it by setting `CODE_SIGN_ENTITLEMENTS` build setting for every configuration. All properties must be provided
-- [ ] **templates**: **[String]** - A list of [Target Templates](#target-template) referenced by name that will be merged with the target in order. Any instances of `$target_name` within these templates will be replaced with the target name.
+- [ ] **templates**: **[String]** - A list of [Target Templates](#target-template) referenced by name that will be merged with the target in order. Any instances of `${target_name}` within these templates will be replaced with the target name.
+- [ ] **templateAttributes**: **[String: String]** - A list of attributes where each instances of `$<attributeName>` within the templates listed in `templates` will be replaced with the value specified here.
```
yonaskolb:
```suggestion
- [ ] **templateAttributes**: **[String: String]** - A list of attributes where each instance of `${attributeName}` within the templates listed in `templates` will be replaced with the value specified.
```

Docs/ProjectSpec.md
```diff
@@ -583,22 +584,25 @@ This is used to override settings or run build scripts in specific targets
...
+Any attributes defined within a targets `templateAttributes` will be used to replace any attribute references in the template using the syntax `${attribute_name}`.
 
 
 ```yaml
 targets:
   MyFramework:
     templates: 
       - Framework
+    templateAttributes:
+      frameworkName: AwsomeFramework
```
yonaskolb:
```suggestion
      frameworkName: AwesomeFramework
```

tomquist:
I'll rebase the branch on master and add additional tests that combine nested templates with targetAttributes.

tomquist:
I adjusted the `parses complex nested target templates` spec to also include `templateAttributes`.

yonaskolb:
Great work!

PR #534, opened 2019-03-14
Add support for nested templates

tomquist:
It would be convenient if templates could be nested, which means
a template can be based on another template.

This change implements support for nested templates. It avoids
cycles by just ignoring templates that have already been visited
when collecting the set of templates to use.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Added `missingConfigFiles` to `options.disabledValidations` to optionally skip checking for the existence of config files.
 - Added ability to automatically include Carthage related dependencies via `includeRelated: true` [#506](https://github.com/yonaskolb/XcodeGen/pull/506) @rpassis
 - Added ability to define a per-platform `deploymentTarget` for Multi-Platform targets. [#510](https://github.com/yonaskolb/XcodeGen/pull/510) @ainopara
+- Added support for nested target templates [#534](https://github.com/yonaskolb/XcodeGen/pull/534) @tomquist
 
 #### Fixed
 - Sources outside a project spec's directory will be correctly referenced as relative paths in the project file. [#524](https://github.com/yonaskolb/XcodeGen/pull/524)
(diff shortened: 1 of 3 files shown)


yonaskolb:
Another great PR @tomquist. Along with https://github.com/yonaskolb/XcodeGen/pull/533, target templates are becoming really powerful. Whichever one gets merged first, I'd like to see some tests that combine nesting with targetAttributes to see how they combine if a template uses `targetAttributes`

Sources/ProjectSpec/Target.swift
```diff
@@ -139,9 +139,26 @@ extension Target {
...
+                return
+            }
+            var index = index
+            for template in templates where !allTemplates.contains(template) {
+                guard let templateDictionary = targetTemplatesDictionary[template] else {
+                    continue
+                }
+                allTemplates.insert(template, at: index)
+                collectTemplates(of: templateDictionary, into: &allTemplates, at: index)
+                index += 1
```
yonaskolb: From reading the code it's not immediately obvious to me what the index does?

tomquist: The index makes sure that templates are inserted in the right order, e.g. this template tree:
 - A
   - B
   - C
 - D
   - E

will result in the following template list: B,C,A,E,D
After collecting the list templates will be processed as previously.
This ensures that the leaf templates are always processed before processing the parent templates so that parent templates can override values.

I tried to make that more clear from the code by renaming the parameter. And now that you point it out I realized that the logic was wrong. I fixed that and added a test case that covers this specific example.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 #### Added
 - Added `missingConfigFiles` to `options.disabledValidations` to optionally skip checking for the existence of config files.
 - Added ability to automatically include Carthage related dependencies via `includeRelated: true` [#506](https://github.com/yonaskolb/XcodeGen/pull/506) @rpassis
+- Added support for nested templates [#534](https://github.com/yonaskolb/XcodeGen/pull/534) @tomquist
```
```suggestion
- Added support for nested target templates [#534](https://github.com/yonaskolb/XcodeGen/pull/534) @tomquist
```

PR #541, opened 2019-03-22
Fix multi-platform target templates

### LINKED ISSUE #523: Cannot use Target.platform as array
Hello.
Thank you for the tool, it's great. 
Today I found an issue with `Target.platform`. When I try to generate Xcode project using XcodeGen 2.2.0 with the `project.yml` created for XcodeGen 2.0.0 I get this error:
**Parsing project spec failed: Decoding failed at "platform": Incorrect type, expected String found ["iOS"]**
As far as I can see [here](https://github.com/yonaskolb/XcodeGen/blob/master/Docs/ProjectSpec.md) `Target.platform` can be an array of strings or a string. Is it still correct? Can I use `platform: [iOS]` in XcodeGen 2.2.0?

yonaskolb:
Fixes #523
This was due to multi-platform targets being resolved before templates. This order is now switched

### DIFF
--- a/Tests/XcodeGenKitTests/SpecLoadingTests.swift
+++ b/Tests/XcodeGenKitTests/SpecLoadingTests.swift
@@ -613,6 +613,32 @@ class SpecLoadingTests: XCTestCase {
                 try expect(tvOSTarget?.type) == .framework
             }
 
+            $0.it("parses platform specific templates") {
+
+                let project = try getProjectSpec([
+                    "targets": [
+                        "Framework": [
+                            "type": "framework",
+                            "platform": ["iOS", "tvOS"],
+                            "templates": ["$platform"],
+                        ]
+                    ],
+                    "targetTemplates": [
+                        "iOS": [
+                            "sources": "A",
+                        ],
+                        "tvOS": [
+                            "sources": "B",
+                        ]
+                    ],
+                    ])
+
+                let iOSTarget = project.targets.first { $0.platform == .iOS }
+                let tvOSTarget = project.targets.first { $0.platform == .tvOS }
+                try expect(iOSTarget?.sources) == ["A"]
+                try expect(tvOSTarget?.sources) == ["B"]
+            }
+
             $0.it("parses aggregate targets") {
                 let dictionary: [String: Any] = [
                     "targets": ["target_1", "target_2"],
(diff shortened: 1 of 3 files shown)


yonaskolb:
The one downside of this is that you couldn’t reference a template dynamically based on the $platform. I think this is less likely than having multi platform templates though

Tests/XcodeGenKitTests/SpecLoadingTests.swift
```diff
@@ -590,6 +590,28 @@ class SpecLoadingTests: XCTestCase {
...
+                        "temp": [
+                            "platform": ["iOS", "tvOS"],
+                        ]
+                    ],
+                    ])
+                
+                let iOSTarget = project.targets.first { $0.platform == .iOS }
+                let tvOSTarget = project.targets.first { $0.platform == .tvOS }
+                try expect(iOSTarget?.type) == .framework
+                try expect(tvOSTarget?.type) == .framework
```
RomanPodymov: The solution should be more generic, something like that:
```
let platforms: [Platform] = [.iOS, .tvOS]
platforms.forEach { platform in
    let target = project.targets.first { $0.platform == platform }
    try expect(target?.type) == .framework
}
```
Or maybe you should use `Platform.all` here.

Tests/XcodeGenKitTests/SpecLoadingTests.swift
```diff
@@ -590,6 +590,28 @@ class SpecLoadingTests: XCTestCase {
...
+                let project = try getProjectSpec([
+                    "targets": [
+                        "Framework": [
+                            "type": "framework",
+                            "templates": ["temp"],
+                        ]
+                    ],
+                    "targetTemplates": [
+                        "temp": [
+                            "platform": ["iOS", "tvOS"],
```
RomanPodymov: Here you should use `"platform": platforms.map { $0.rawValue }`

Tests/XcodeGenKitTests/SpecLoadingTests.swift
```diff
@@ -590,6 +590,28 @@ class SpecLoadingTests: XCTestCase {
                 try expect(target.sources) == ["nestedTemplateSource2", "nestedTemplateSource1", "templateSource", "targetSource"] // merges array in order
                 try expect(target.configFiles["debug"]) == "Configs/Framework/debug.xcconfig" // replaces $target_name
             }
+            
+            $0.it("parses cross platform target templates") {
+
```
RomanPodymov: Probably `let platforms: [Platform] = [.iOS, .tvOS]` should be defined here.

Issue #9, opened 2017-07-24
Spec generation


yonaskolb:
Should be able to generate a spec from an existing project for easy migration

yonaskolb:
At the moment no. It’s a huge piece of work to do this properly, especially without duplicating everything (having a path to every file in your project for example)
It would require #123 and maybe #214 to be completed first

ladislas:
thanks! :) 

do you have a roadmap of things to be worked on and how to tackle them for people willing to give a hand?

yonaskolb:
There's no real roadmap documented. All the issues are far game. If you want to work on any just mention it in the issue 👍

toshi0383:
FYI, if you extract the "Build Settings" into xcconfigs, migration is a lot easier.
I use my own tool and it works in most cases.
https://github.com/toshi0383/xcconfig-extractor

ryohey:
@yonaskolb I will try to work on this 🤘

yonaskolb:
@ryohey nice! 👍

yonaskolb:
Draft PR here https://github.com/yonaskolb/XcodeGen/pull/682

2jumper3:
Hi all! ) 
Any updates?)

shreesha-kedilaya:
Any updates on this.

PR #545, opened 2019-03-26
Add ability to encode ProjectSpec to JSON

### LINKED ISSUE #123: Serialize `ProjectSpec` to JSON
Good afternoon!

Thanks so much for creating XcodeGen, the project is great.

I'd like to use the `ProjectSpec` and children as a protocol to write out a JSON file. My custom generator would be allocating data structures in `ProjectSpec` directly

Without being able to write the `ProjectSpec` and dependencies to JSON, I need to implement my own version of the spec ( currently on Swift 4 codeable ).

Is this something Xcode gen would support?

ryohey:
I first tried to implement spec generation, but as you mentioned, I needed to be able to serialize ProjectSpec. https://github.com/yonaskolb/XcodeGen/issues/9#issuecomment-368647138

I defined `JSONDictionaryEncodable` protocol like `JSONObjectConvertible` of JSONUtilities, and implemented it in ProjectSpec and related classes. I did not use Codable because some classes do not have a simple correspondence with the JSON structure.

I wrote a test that encodes ProjectSpec into JSON and decodes back and to test if they are equal. However, it may not be perfect because the tests do not cover all patterns and implementations are written by a lot of manual work.

fixes #123


Sources/ProjectSpec/AggregateTarget.swift
```diff
@@ -62,6 +62,35 @@ extension AggregateTarget: NamedJSONDictionaryConvertible {
...
+    public func toJSONDictionary() -> JSONDictionary {
+        var dict: JSONDictionary = [:]
+
+        let settingsDict = settings.toJSONDictionary()
+        if settingsDict.count > 0 {
+            dict["settings"] = settingsDict
+        }
+        
+        if targets.count > 0 {
+            dict["targets"] = targets
```
yonaskolb: All these `if array.count > 0 {` are not really required. It might be useful for some use cases to generate out all the empty arrays. For the compact spec you have now, there could be a function `JSONDictionary.removeEmptyArrays` that loops through and removes all empty arrays

Sources/ProjectSpec/DeploymentTarget.swift
```diff
@@ -78,3 +78,24 @@ extension DeploymentTarget: JSONObjectConvertible {
...
+        if let iOS = iOS {
+            dict["iOS"] = iOS.string
+        }
+        if let tvOS = tvOS {
+            dict["tvOS"] = tvOS.string
+        }
+        if let watchOS = watchOS {
+            dict["watchOS"] = watchOS.string
+        }
+        if let macOS = macOS {
```
yonaskolb: All these if statements for optional values could be removed if the dictionary was `[String: Any?]`.

Sources/ProjectSpec/Encoding.swift
```diff
@@ -0,0 +1,19 @@
...
+    func toJSONArray() -> JSONArray
+}
+
+public protocol JSONPrimitiveEncodable {
+    func toJSONValue() -> JSONRawType
+}
+
+public protocol JSONDynamicEncodable {
+    // returns JSONDictionary or JSONArray or JSONRawType
+    func toJSONValue() -> Any
```
yonaskolb: I'm not sure we need all these protocols as they aren't really used on their own. I think all the functions could just be the `func toJSONValue() -> Any`. Maybe just the top level spec would have a function that returned a dictionary.

Tests/XcodeGenKitTests/ProjectSpecTests.swift
```diff
@@ -256,6 +256,248 @@ class ProjectSpecTests: XCTestCase {
             }
         }
     }
+
+    func testJSONEncodable() {
+        describe {
+            $0.it("encodes to json") {
+                let proj = Project(basePath: Path.current,
```
yonaskolb: Nice object 😄 This might be useful for other tests

Sources/ProjectSpec/AggregateTarget.swift
AlexisQapa: Definitively a good approach generating empty template really help getting started.

ryohey:
@yonaskolb Thank you for your reviewing! I fixed them up. As you said, It would be nice to add the process of removing nil and empty arrays / dictionaries in the process of generating JSON for human reading.

Sources/ProjectSpec/TargetSource.swift
```diff
@@ -181,6 +181,32 @@ extension TargetSource: JSONObjectConvertible {
...
+            "excludes": excludes,
+            "name": name,
+            "headerVisibility": headerVisibility?.rawValue,
+            "type": type?.rawValue,
+            "buildPhase": buildPhase?.toJSONValue(),
+            "createIntermediateGroups": createIntermediateGroups,
+        ]
+
+        if optional {
+            dict["optional"] = true
```
yonaskolb: I'm a bit worried about the maintainability of all these properties, and that regressions might occur in the future by accident. 

To aid with the properties that have defaults, and because we're now mapping those defaults in 3 places (property init, decoding, and now encoding), it would be good if we have static properties for all of these. In this example:
```
static let defaultOptional = true
```

ryohey: @yonaskolb It's make sense. Since I changed it, could you please review it?

Sources/ProjectSpec/Scheme.swift
```diff
@@ -294,8 +304,8 @@ extension Scheme.Test.TestTarget: JSONObjectConvertible {
 
     public init(jsonDictionary: JSONDictionary) throws {
         name = try jsonDictionary.json(atKeyPath: "name")
-        randomExecutionOrder = jsonDictionary.json(atKeyPath: "randomExecutionOrder") ?? false
-        parallelizable = jsonDictionary.json(atKeyPath: "parallelizable") ?? false
+        randomExecutionOrder = jsonDictionary.json(atKeyPath: "randomExecutionOrder") ?? Scheme.Test.TestTarget.randomExecutionOrderDefault
+        parallelizable = jsonDictionary.json(atKeyPath: "parallelizable") ?? Scheme.Test.TestTarget.parallelizableDefault
```
yonaskolb: Very minor and doesn't need to change, but not sure you need to traverse the whole namespace path, I think you can just use `TestTarget.parallelizableDefault`

Sources/ProjectSpec/TargetSource.swift
yonaskolb: Looks good! Could you please rebase to fix the merge conflict and add a changelog?

Sources/ProjectSpec/Scheme.swift
ryohey: Thanks!
After trying it out, it seems necessary to specify the complete namespace in the extension.

ryohey:
@yonaskolb Thank you for your review! I rebased it and fix conflicts, added changelog.

yonaskolb:
Thanks @ryohey! Not sure how you want to follow this PR up and how it will be used (you mentioned project migration), but I think it would be useful having a parameter like `Project.toDictionary(style: ProjectEncodingStyle)` that does the empty array and dictionary removal
```swift
enum ProjectEncodingStyle {
  /// removes empty dictionaries and arrays
  case minimal
  case full
}
```
Feel free to do that here or in another PR. Perhaps would be better here

ryohey:
@yonaskolb Thanks! I received a pull-request from @AndreiVidrasco that adds feature to filter out empty values from dictionary. and we talked about this pr #545. So I merge this pr this time.
@AndreiVidrasco I would like you to make a pull-request that reflects this conversation https://github.com/yonaskolb/XcodeGen/pull/545#issuecomment-485240367

PR #547, opened 2019-03-27
Fix creating optional file

### LINKED ISSUE #546: Optional path not working
Hi.
Following project spec should make optional file reference, but it's not made.

```yml
name: MyProject
targets:
  MyFramework:
    platform: iOS
    type: framework
    sources:
      - path: Sources
      - path: Sources/File.swift
        optional: true
```

#527 might be related.

xcodegen 2.3.0

Thanks.

KyoheiG3:
Fixes #546.

### DIFF
--- a/Sources/XcodeGenKit/SourceGenerator.swift
+++ b/Sources/XcodeGenKit/SourceGenerator.swift
@@ -463,11 +463,6 @@ class SourceGenerator {
     /// creates source files
     private func getSourceFiles(targetType: PBXProductType, targetSource: TargetSource, path: Path) throws -> [SourceFile] {
 
-        if targetSource.optional && !Path(targetSource.path).exists {
-            // Is optional so if it doesn't exist just return an empty array
-            return []
-        }
-
         // generate excluded paths
         targetSourceExcludePaths = getSourceExcludes(targetSource: targetSource)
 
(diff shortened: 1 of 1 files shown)


yonaskolb:
Hi @KyoheiG3. I'm not sure what this fixes?

Sources/XcodeGenKit/SourceGenerator.swift
```diff
@@ -521,6 +516,11 @@ class SourceGenerator {
             sourceFiles.append(sourceFile)
 
         case .group:
+            if targetSource.optional {
+                // Is optional so if it doesn't exist just return an empty array
+                return []
```
yonaskolb: This isn't checking if a directory exists, so it would always return nothing

ra1028:
Hi @yonaskolb 

This pull request is my colleague's.
We want to fix to reference auto-generated files that hasn't been generated yet when runs xcodegen.
For example, case of generate localized strings file using SwiftGen in build phases.
Requires to generate a Xcode project before runs SwiftGen in the pre-build phase.
So we expect like following: 
Screen Shot 2019-03-28 at 23 38 47

How should we fix?

AlexisQapa:
I think there is a pre gen hook or at least it was mentioned in another issue. Right now I deal with this by either having blank files (we don't use swiftgen for storyboard so files are known in advance) or running swiftgen once before Xcode gen.

yonaskolb:
Comments left in https://github.com/yonaskolb/XcodeGen/issues/546
@AlexisQapa suggestions should resolve this issue.
Closing this for now

Issue #525, opened 2019-02-22
Subgroups are flattened when Sources paths contain ".." in YAML file


nzhuk:
xcodegen 2.2.0

Test setup:

```
.
├── Sources
│   ├── Subfolder
│   │   └── file2.swift
│   └── file1.swift
└── spec
    └── project.yml
```

`spec/project.yml`:
```
name: MyProject
targets:
  MyFramework:
    type: framework
    platform: iOS
    sources:
      - path: ../Sources
```

Command: `xcodegen --spec spec/project.yml`

`MyProject.xcodeproj` looks like this:

![screenshot 2019-02-22 10 43 46](https://user-images.githubusercontent.com/294154/53230169-c16e5480-368e-11e9-9c58-25de94d9986c.png)

Current: `Subfolder` is not a subgroup of `Sources` group
Expected: `Subfolder` is a subgroup of `Sources` group

Note: This seems to be triggered when path of Sources in YAML file contains at least one `../`

yonaskolb:
Thanks for the report @nzhuk. This will most likely be fixed by #524

sroebert:
I just checked, but it does not seem to solve this issue.

alejandro-isaza:
As an alternative I tried moving the project.yml file one level up and using `codegen --project subfolder` but then the references in the project are broken: `
Build input file cannot be found: '...'`

yonaskolb:
Can you try again in Xcode 2.5.0, now that it includes #550

nzhuk:
@yonaskolb The reported issue seems to be fixed in 2.5.0.

`xcodegen --spec spec/project.yml` produces correct project:

![Screenshot 2019-04-10 12 42 32](https://user-images.githubusercontent.com/294154/55868930-2114b680-5b8e-11e9-8ac7-8be9a4553180.png)


However, if the Xcode project is generated to the parent directory of `spec` directory, the project is still broken (probably another issue, not sure where it's tracked):

`xcodegen --spec spec/project.yml -p .`:

![Screenshot 2019-04-10 12 44 13](https://user-images.githubusercontent.com/294154/55869040-5faa7100-5b8e-11e9-94e5-64e02cfa5df2.png)

(here, "Sources" group refers to invalid folder, thus "test.swift" is not found)

yonaskolb:
[...]

PR #550, opened 2019-03-27
Fixed flattening of relative paths (#525)

sroebert:
Currently when adding relative paths as a source root, all folders are flattened as root groups in the Xcode project (#525). This fix makes sure the folder structure is kept without flattening.

### DIFF
--- a/Sources/XcodeGenKit/SourceGenerator.swift
+++ b/Sources/XcodeGenKit/SourceGenerator.swift
@@ -247,7 +247,7 @@ class SourceGenerator {
             let isOutOfBasePath = !path.absolute().string.contains(project.basePath.absolute().string)
 
             // has no valid parent paths
-            let isRootPath = isOutOfBasePath || path.parent() == project.basePath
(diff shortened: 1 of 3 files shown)


yonaskolb:
Thanks for this @sroebert. Are you able to add a unit test to `SourceGeneratorTests` that demonstrates the issue and the fix.

You'll also need to rebase to fix the conflicts

sroebert:
Done!

Sources/XcodeGenKit/SourceGenerator.swift
```diff
@@ -247,7 +247,7 @@ class SourceGenerator {
             let isOutOfBasePath = !path.absolute().string.contains(project.basePath.absolute().string)
 
             // has no valid parent paths
-            let isRootPath = isOutOfBasePath || path.parent() == project.basePath
+            let isRootPath = (isBaseGroup && isOutOfBasePath) || path.parent() == project.basePath
```
yonaskolb: The `isBaseGroup` param is a newer addition than the rest of this line. I wonder if that's all we need

PR #557, opened 2019-04-04
Fix optional file sources

### LINKED ISSUE #546: Optional path not working

yonaskolb:
This fixes files marked as optional from being added to the project. Directories will still not be added

Fixes #546


ra1028:
Looks good.
Thanks!

Issue #511, opened 2019-02-08
Carthage watch copy phase


raptorxcz:
Hi,

When I generate project with apple Watch App which uses Carthage, the Carthage copy phase is in the wrong target.

Here is example project:
```yml
name: MyApp
options:
  carthageBuildPath: "../../Carthage/Build"
targets:
  MyApp:
    sources:
      - path: "sources"
    platform: iOS
    type: application
    dependencies:
      - carthage: Charts
      - target: MyApp WatchApp
  MyApp WatchApp:
    sources:
      - path: "sources"
    platform: watchOS
    type: 'application.watchapp2'
    dependencies:
      - target: MyApp WatchApp Extension
  MyApp WatchApp Extension:
    deploymentTarget: "3.0"
    sources:
      - path: "sources"      
    platform: watchOS
    type: 'watchkit2-extension'
    dependencies:
      - carthage: Charts
```

<img width="565" alt="screenshot 2019-02-08 at 16 27 41" src="https://user-images.githubusercontent.com/3967087/52487751-9bd94980-2bbe-11e9-831b-fbb12c062b60.png">
<img width="530" alt="screenshot 2019-02-08 at 16 27 46" src="https://user-images.githubusercontent.com/3967087/52487752-9bd94980-2bbe-11e9-830a-2e95fec15841.png">

PR #558, opened 2019-04-05
Fix Carthage dependencies for watchOS target

KhaosT:
It appears that App Store does not allow embedding additional frameworks directly inside WatchKit app's framework directory (ITMS-90539). Instead, all embedded frameworks should live inside the app extension.

This PR updated the behavior to embed Carthage dependencies to watch extension, instead of the WatchKit app. Should address #511.

https://forums.developer.apple.com/thread/76615


yonaskolb:
Thanks @KhaosT! Could you also please add a changelog entry

PR #564, opened 2019-04-09
support sdks with dylib extension

wag-miles:
Attempting to build a project with a dependency like `- sdk: libz.dylib` causes an error: `Spec validation error: Target "TARGET_NAME" has invalid sdk dependency: "libz.dylib". It must be a full path or have the following extensions: .framework, .dylib, .tbd`

### DIFF
--- a/Sources/ProjectSpec/SpecValidation.swift
+++ b/Sources/ProjectSpec/SpecValidation.swift
@@ -149,6 +149,7 @@ extension Project {
                     if !dependency.reference.contains("/") {
                         switch path.extension {
                         case "framework"?,
+                            "dylib"?,
                              "tbd"?:
                             break
                         default:


yonaskolb:
Hi @wag-miles. Is the expectation that if you provide just a simple filename that the `dylib` should be looked up in `/usr/lib`? In that case you would need to add code here as well https://github.com/yonaskolb/XcodeGen/blob/master/Sources/XcodeGenKit/PBXProjGenerator.swift#L551

The docs here would need to be updated https://github.com/yonaskolb/XcodeGen/blob/master/Sources/XcodeGenKit/PBXProjGenerator.swift#L551, as well as the Changelog

wag-miles:
Would this be in keeping with the intentions of the project? I was confused since the error mentions the `.dylib` extension. Is there another way to accomplish linking to one?

-Miles

> On Apr 10, 2019, at 2:27 AM, Yonas Kolb <notifications@github.com> wrote:
> 
> Hi @wag-miles. Is the expectation that if you provide just a simple filename that the dylib should be looked up in /usr/lib? In that case you would need to add code here as well https://github.com/yonaskolb/XcodeGen/blob/master/Sources/XcodeGenKit/PBXProjGenerator.swift#L551
> 
> The docs here would need to be updated https://github.com/yonaskolb/XcodeGen/blob/master/Sources/XcodeGenKit/PBXProjGenerator.swift#L551, as well as the Changelog
> 
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or mute the thread.

-- 
The information contained in this message is intended for the addressee 
only and may contain classified information. If you are not the addressee, 
please delete this message and notify the sender; you should not copy or 
distribute this message or disclose its contents to anyone.

yonaskolb:
From what I understand the use of `dylib` is deprecated in favour of `.tbd`, so in your case you should be able to use `- sdk: libz.tbd`. At the moment the error mentions `dylib` even though it's not really supported.

But if `dylib` is still required by anyone we should definitely add support for it, which includes the list of changes I mentioned above.

yonaskolb:
@wag-miles would you like to finish this PR off to add support for `.dylib`?

yonaskolb:
@wag-miles this HD been added here https://github.com/yonaskolb/XcodeGen/pull/650

Issue #574, opened 2019-04-25
Target template duplications


haritowa:
I have an issue when target template properties are duplicated each time `yml` file is included. 
Here is my sample setup:

```.
├── A
│   ├── Sources
│   │   └── AClass.swift
│   └── project.yml
├── B
│   ├── Sources
│   │   └── BClass.swift
│   └── project.yml
└── baseProject.yml

4 directories, 5 files
```

`baseProject.yml` contains target template `Core`:
```name: BaseProject
targetTemplates:
  Core:
    platform: iOS
    type: framework
    preBuildScripts:
      - name: Swifty Mocky
        script: echo "123"
```

Both `A/project.yml` and `B/project.yml` has an `- ../baseProject.yml` include. `xcodegen` invocation against `B/project.yml` leads to duplicated properties(build scripts, for example) for any target, that have the `Core` template.

![Duplicated Build Scripts](https://cl.ly/55fa3f4a7d5b/Image%2525202019-04-25%252520at%2525208.21.46%252520PM.png)

[Sample Project](https://github.com/yonaskolb/XcodeGen/files/3118139/project.zip)

haritowa:
Hello everyone, I have made some progress:

- This issue relates to any spec element, not only target templates
- Any array value is multiplied

As far as I understand, the problem is that we have duplicated `subSpecs`, which forces this method https://github.com/yonaskolb/XcodeGen/blob/b490851c737809ee3cb545ee8ea41742cfb5a84d/Sources/ProjectSpec/SpecFile.swift#L85-L95 to merge spec onto itself.

I propose to merge each spec *file* only once. I already have the fix, which solves this issue and doesn't break any existing tests. 

Before any PR fuss, I want to be sure that this issue is actually a bug. So, after @yonaskolb provides details, I'll produce tests cases and open PR.

haritowa:
Meanwhile, @nivanchikov, could you share your setup, which encounters the same problem? I want to test my fix against your case.

As far as I understand, you may bypass this issue adding `:REPLACE` suffix to any property you don't want to be merged. (Try to replace `preBuildScripts` with `preBuildScripts:REPLACE` in your target template.)

yonaskolb:
@haritowa your setup indeed includes the base spec twice. If you want to handle a scenario like this instead of changing your includes, then a fix would be required to only include a path a single time. If you'd like to create a PR that would be great

nivanchikov:
[...]

PR #577, opened 2019-05-03
Fix duplicate build scripts being generated when using target templates

nivanchikov:
This PR fixes duplicate builds scripts issue described in #574.

I haven't experienced any issues with other properties being duplicated except for the build scripts when using the same target template for several targets.

The change attempts to generate a unique set of build scripts for each build phase based on `BuildScript` equality checks.

### DIFF
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -273,6 +273,14 @@ public class PBXProjGenerator {
         )
         return targetDependency
     }
+    
+    func generateUniqueBuildScripts(targetName: String, from buildScripts: [BuildScript]) throws -> [PBXShellScriptBuildPhase]  {
+        let uniqueBuildScripts = buildScripts.reduce([BuildScript](), { acc, script in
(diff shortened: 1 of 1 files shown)


yonaskolb:
Thanks @nivanchikov. Do you have a test case that would break without this change?

nivanchikov:
@yonaskolb sorry, you mean a unit test or a test project?

yonaskolb:
A unit test would be great

haritowa:
@nivanchikov hi, thanks a lot. As far as I can see, your PR solves duplicates issue only for build scripts. Unfortunately, this issue spreads on other Target props. I'll try to find time to write some test cases, so it will be easier to narrow the root cause and deliver proper fixes ^_^

yonaskolb:
As @haritowa says this may just be a symptom of a larger issue. 
@nivanchikov does your setup include a spec multiple times? We can take the conversation to #574

nivanchikov:
@yonaskolb yes, exactly - I have a base spec which is included by several other specs. Thus I suppose this PR can be closed in favor of work being done in #574

PR #581, opened 2019-05-07
Change compatibilityVersion to Xcode 10.0

acecilia:
I experienced that with the `compatibilityVersion` set to `Xcode 9.3`, Xcode automatically sets the `ObjectVersion` from `51` to `50`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 - Added ability to encode ProjectSpec to JSON [#545](https://github.com/yonaskolb/XcodeGen/pull/545) @ryohey
 
 #### Changed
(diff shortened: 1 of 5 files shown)


yonaskolb:
Interesting. Is the downgrade in object version affecting you?
I wonder if anyone is still using Xcode 9.3 with the latest XcodeGen, probably not

acecilia:
Yes it is. I am not sure at which point is happening (I did some tests but could not figure out in which exact moment Xcode decides to update the `ObjectVersion`), but after working with a project for some time and going to commit, I see that `ObjectVersion` changed from `51` to `50`.

To check this you can make a test:
- Create a new `Single View App` project with Xcode. It will be created with `ObjectVersion = 50` and `compatibilityVersion = "Xcode 9.3"`.
- With the project file selected, in the `file inspector` (using the dropdown) change the compatibility to `Xcode 10.0-compatible`: the `ObjectVersion` is automatically changed to 51.
- Changing the compatibility back to `Xcode 9.3-compatible` also automatically changes  the `ObjectVersion` to 50.

Due to that, seems that `ObjectVersion = 51` is only used with `compatibilityVersion = "Xcode 10.0"`.

PR #582, opened 2019-05-08
Allow skipping tests in test targets

kadarandras:
Since the used xcodeproj project allows an extra `skippedTests` parameter inside the `TestableReference.init`, with a small change in the project I was able to implement the same functionality here.

https://github.com/tuist/xcodeproj/blob/master/Sources/xcodeproj/Scheme/XCScheme%2BTestableReference.swift#L20

### DIFF
--- a/Sources/ProjectSpec/Scheme.swift
+++ b/Sources/ProjectSpec/Scheme.swift
@@ -102,25 +102,30 @@ public struct Scheme: Equatable {
         public struct TestTarget: Equatable, ExpressibleByStringLiteral {
             public static let randomExecutionOrderDefault = false
             public static let parallelizableDefault = false
+            public static let skippedTestsDefault: [String] = []
 
             public let name: String
             public var randomExecutionOrder: Bool
             public var parallelizable: Bool
+            public var skippedTests: [String]
 
             public init(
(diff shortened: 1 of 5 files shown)


Sources/ProjectSpec/Scheme.swift
```diff
@@ -102,25 +102,30 @@ public struct Scheme: Equatable {
         public struct TestTarget: Equatable, ExpressibleByStringLiteral {
             public static let randomExecutionOrderDefault = false
             public static let parallelizableDefault = false
+            public static let skippedTestsDefault: [String] = []
```
yonaskolb: We probably don't need to add empty arrays as explicit default members

yonaskolb:
Great, thanks @kadarandras! Could you add a changelog entry as well

kadarandras:
@yonaskolb I have added & modified the requested changes.

yonaskolb:
Thanks @kadarandras! 😄

thii:
@yonaskolb Could you cut a new release that includes this?

yonaskolb:
@thii done! Was going to do a release today anyway, so good timing 😄

PR #583, opened 2019-05-08
Add support for customizing attributes on a target source file

min:
When adding to an intents definition file to a target, Xcode will do some code gen at build time. In some cases we need to add the definition file to multiple targets (shared frameworks), but only have the code generation happen for one of the targets to avoid duplicated code.

Per apple [docs](https://developer.apple.com/documentation/sirikit/soup_chef_accelerating_app_interactions_with_shortcuts):

> To use a custom intent in your app, Xcode needs to generate source code for the items defined in the Intent Definition file. However, for apps that have a shared framework, like Soup Chef, code generation should only happen for the framework target. Generating the code for the shared framework and app causes conflicts due to duplicated code.
**To specify which target has the generated source code, use the Intent Classes setting in the Target Membership panel. For Soup Chef, only the SoupKit target uses the Intent Classes setting; all other targets use the No Generated Classes setting.** And because SoupKit shares its code with other targets, the other targets can use the order soup intent.

Screen Shot 2019-05-08 at 6 52 18 AM

This adds a `noCoden` property to target source which internal sets `settings = {ATTRIBUTES = (no_codegen, ); };` on the intents definition PBXBuildFile.


min:
@yonaskolb this is the last remaining issue after migrating our app to xcodegen. not sure if this is the correct solution, would love sone feedback .

yonaskolb:
Thanks for the PR @min!
Perhaps another way to approach this would be to allow custom attributes on a `TargetSource` in the form of `[String: Any]`, similar to `Target`.
That would allow for more flexibility going forward. I would probably prefer not to implement every possible attribute manually like this, though I do know it makes it a bit more user friendly. What do you think?

min:
@yonaskolb That makes sense, changed to a `[String]` as that's what's expected for `ATTRIBUTES`.

min:
@yonaskolb would love to get this merged in as we're using a forked version for now :)

Docs/ProjectSpec.md
```diff
@@ -326,6 +326,7 @@ A source can be provided via a string (the path) or an object of the form:
 	- `public`
 	- `private`
 	- `project`
+- [ ] **attributes**: **[String]** - Additional settings attributes for the file reference.
```
yonaskolb:
```suggestion
- [ ] **attributes**: **[String]** - Additional settings attributes that will be applied to any build files.
```

min:
@yonaskolb all set

CHANGELOG.md
```diff
@@ -2,10 +2,9 @@
 
 ## Next Version
 
-#### Added
```
yonaskolb: Was removing this heading an accident? Let's keep it around. It's fine to remove the blank line

min: @yonaskolb that was a merge error, should be fixed now.

PR #594, opened 2019-06-05
Add environment variable expansion

### LINKED ISSUE #587: Allow using env vars in project.yml
Similar to how SwiftLint does: https://github.com/realm/SwiftLint/pull/1703

tomquist:
closes #587

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Added ability to encode ProjectSpec to JSON [#545](https://github.com/yonaskolb/XcodeGen/pull/545) @ryohey
 - Added ability to skip tests [#582](https://github.com/yonaskolb/XcodeGen/pull/582) @kadarandras
 - Added ability to set `attributes` on build files [#583](https://github.com/yonaskolb/XcodeGen/pull/583) @min
+- Allow using environment variables in the form of `${SOME_VARIABLE}`. This might be a **breaking** change when a target template attribute is also defined as an environment variable [#594](https://github.com/yonaskolb/XcodeGen/pull/594) @tomquist
 
 #### Fixed
 - Fixed `.pch` files being bundled as resources [#597](https://github.com/yonaskolb/XcodeGen/pull/597) @thii
--- a/Sources/ProjectSpec/SpecFile.swift
+++ b/Sources/ProjectSpec/SpecFile.swift
@@ -82,9 +82,9 @@ public struct SpecFile {
         }
     }
 
-    public func resolvedDictionary(environmentVariables: [String: String] = ProcessInfo.processInfo.environment) -> JSONDictionary {
+    public func resolvedDictionary(variables: [String: String] = [:]) -> JSONDictionary {
         var resolvedSpec = resolvingPaths().mergedDictionary()
-        for (key, value) in environmentVariables {
+        for (key, value) in variables {
             resolvedSpec = resolvedSpec.replaceString("${\(key)}", with: value)
         }
         return resolvedSpec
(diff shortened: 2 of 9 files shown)


Sources/ProjectSpec/SpecFile.swift
```diff
@@ -82,9 +82,12 @@ public struct SpecFile {
         }
     }
 
-    public func resolvedDictionary() -> JSONDictionary {
-        let resolvedSpec = resolvingPaths()
-        return resolvedSpec.mergedDictionary()
+    public func resolvedDictionary(environmentVariables: [String: String] = ProcessInfo.processInfo.environment) -> JSONDictionary {
```
yonaskolb:
```suggestion
    public func resolvedDictionary(variables: [String: String] = [:]) -> JSONDictionary {
```
This lives in the project spec so we can allow it to handle variables coming from anywhere.

Sources/XcodeGenKit/SpecLoader.swift
```diff
@@ -15,9 +15,9 @@ public class SpecLoader {
         self.version = version
     }
 
-    public func loadProject(path: Path) throws -> Project {
+    public func loadProject(path: Path, environmentVariables: [String: String] = ProcessInfo.processInfo.environment) throws -> Project {
```
yonaskolb:
```suggestion
    public func loadProject(path: Path, variables: [String: String] = [:]) throws -> Project {
```
The generate command can pass in `ProcessInfo.processInfo.environment` here https://github.com/yonaskolb/XcodeGen/blob/master/Sources/XcodeGenCLI/GenerateCommand.swift#L61

Tests/XcodeGenKitTests/SpecLoadingTests.swift
```diff
@@ -954,10 +965,10 @@ fileprivate func getProjectSpec(_ project: [String: Any], file: String = #file,
     }
 }
 
-fileprivate func loadSpec(path: Path, file: String = #file, line: Int = #line) throws -> Project {
+fileprivate func loadSpec(path: Path, environmentVariables: [String: String] = [:], file: String = #file, line: Int = #line) throws -> Project {
```
yonaskolb:
```suggestion
fileprivate func loadSpec(path: Path, variables: [String: String] = [:], file: String = #file, line: Int = #line) throws -> Project {
```

tomquist:
@yonaskolb Good idea, makes it much cleaner.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - Added ability to encode ProjectSpec to JSON [#545](https://github.com/yonaskolb/XcodeGen/pull/545) @ryohey
 - Added ability to skip tests [#582](https://github.com/yonaskolb/XcodeGen/pull/582) @kadarandras
 - Added ability to set `attributes` on build files [#583](https://github.com/yonaskolb/XcodeGen/pull/583) @min
+- Allow using environment variables in the form of `${SOME_VARIABLE}`. This might be a **breaking** change when a target template attribute is also defined as an environment variable [#594](https://github.com/yonaskolb/XcodeGen/pull/594) @tomquist
```
yonaskolb: Would this actually be a breaking change? It seems the variables are replaced AFTER template attributes, so they wouldn’t interfere. Thinking about this now, could you add a test that shows who wins, in this case the template attributes.

tomquist: Unfortunately it's the other way round since variable expansion is done on the JSON dictionary and template attributes get expanded later when the JSON dictionary is converted into a `Project` value.

I added a spec that documents this behavior. 

In general I think this order is good because it allows greatest flexibility when using variables. E.g. you could define which template to use via environment variable which would not be possible if variables were replaced after template attributes.

We could also think about using a different placeholder scheme for variables, e.g. `$(SOME_VARIABLE)`. Not sure if it is a good idea to have two different placeholder types though.

PR #597, opened 2019-06-10
Don't bundle `.pch` files as resources

### LINKED ISSUE #593: `.pch` files are bundled as resources
`.pch` files are prefix headers but they are bundled as resources.

thii:
Fixes #593


yonaskolb:
Thanks @thii! Could you add a changelog entry as well

PR #601, opened 2019-06-19
Added disableMainThreadChecker to TargetScheme and Run/Test Actions

wag-miles:
Added capability to disable the main thread checker. Added values to test cases where appropriate.


Sources/ProjectSpec/Scheme.swift
```diff
@@ -294,6 +306,7 @@ extension Scheme.Test: JSONEncodable {
     public func toJSONValue() -> Any {
         return [
             "gatherCoverageData": gatherCoverageData,
+            "disableMainThreadChecker": disableMainThreadChecker,
```
yonaskolb: It looks like `gatherCoverageData` accidentally missed out on this before, so it was easy to do, but these values should only be written if they differ from the default

yonaskolb: Also is `Scheme.Run` writing this property?

Docs/ProjectSpec.md
```diff
@@ -686,6 +687,7 @@ A multiline script can be written using the various YAML multiline methods, for
 ### Test Action
 
 - [ ] **gatherCoverageData**: **Bool** - a boolean that indicates if this scheme should gather coverage data. This defaults to false
+- [ ] **disableMainThreadChecker**: **Bool** - a boolean that indicates if this scheme should disable disable the Main Thread Checker. This defaults to false
```
yonaskolb: From the code it looks like this is also for the `Run Action` yes?

wag-miles:
I believe I've covered your suggested changes. Thanks!

yonaskolb:
Thanks for the changes @wag-miles. One more thing: Could you edit a scheme in here to set this property https://github.com/yonaskolb/XcodeGen/blob/master/Tests/Fixtures/TestProject/project.yml and then re-run the tests and commit the project diff.
This will serve as an integration test

wag-miles:
> Thanks for the changes @wag-miles. One more thing: Could you edit a scheme in here to set this property https://github.com/yonaskolb/XcodeGen/blob/master/Tests/Fixtures/TestProject/project.yml and then re-run the tests and commit the project diff.
> This will serve as an integration test

Done. Thanks!

bartzet:
Hi @yonaskolb 👋

Do you have an idea when this PR will be merged? We need this flag to be available as after migrating to Xcode 11 we have noticed some tests failing and to fix them we need to disableMainThreadChecker when we set up the project with XcodeGen. 

Please provide more details.

yonaskolb:
@bartzet this is now released in 2.7.0 👍

bartzet:
Amazing. Thank you for acting that quickly @yonaskolb!

PR #607, opened 2019-07-09
Add support for custom groups in target sources

### LINKED ISSUE #478: Add a custom group for a target source
This would add another property to `TargetSource` to be able to specify a custom group that this source should be put into.
By default sources get added to the root of the project, or go up the tree if `options.createIntermediateGroups` is on. This will allow for more flexibility

eg:
```yml
targets:
  App:
    sources:
      - path: Sources/Files
        group: App/CustomName
```

rcari:
Currently, the groups layout in the Xcode project is bound to the filesystem organisation of source files. This patch adds a new 'group' property for target sources which allows the customization of the group where said source should appear in.

Resolves #478

### DIFF
--- a/Docs/ProjectSpec.md
+++ b/Docs/ProjectSpec.md
@@ -302,7 +302,8 @@ A source can be provided via a string (the path) or an object of the form:
 #### Target Source
 
 - [x] **path**: **String** - The path to the source file or directory.
-- [ ] **name**: **String** - Can be used to override the name of the source file or directory. By default the last component of the path is used for the name
+- [ ] **name**: **String** - Can be used to override the name of the source file or directory. By default the last component of the path is used for the name.
+- [ ] **group**: **String** - A path-like string representing the project group hierarchy in which the source file or directory should appear.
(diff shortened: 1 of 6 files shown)


sroebert:
I just tried this with my project and it seems to either flatten all files when a group is specified. Also when I add a `/` in the group name, the script never finishes running, it goes on forever.

rcari:
Hi @sroebert it will flatten all the files when a group is specified, but you can have finer grained control if you want to. For the / are you talking about a trailing one ? Why would you specify that ?

sroebert:
I would prefer not to flatten, but keep the structure, just in a different group if that is possible.

For the `/`, if I just have the group `Test/SomeOtherName` the script already does not finish.

sroebert:
For me, running this with my project resulted in some recursion when adding a group with a `/` in it, so for example `group: Some/Group`. Having previously worked on fixing the flattening of groups, I figured it would be easier to handle the grouping in a slightly different way, which is why I created #611.

rcari:
Will track #611 instead

PR #613, opened 2019-07-23
Add ability to set an order of groups

Beniamiiin:
Hi @yonaskolb.

I implemented the option called `groupOrder`. It's for setting an order of groups. Currently, we can just set the `groupSortPosition`, but sometimes we wanna set directly order of groups.

**Some examples**

1. Set the order of the main group, i.e. the project.

I have a project with a structure:
```
Test.xcodeproj
  Sources
  Resources
  Tests
  Configurations
  Products
  Frameworks
```
Without `groupOrder` option it will be sorted alphabetically, to set the order as above we can use the new option in this way:
```
options:
  groupsOrder: 
    - order: [Sources, Resources, Tests, Support files, Configurations]
```

2. Set the order of groups whose names end with `Screen`.

I have a group with a structure:
```
Test.xcodeproj
  Sources
    PresentationLayer
      SplashScreen
        View
        Presenter
        Interactor
        Entities
        Assembly
  Resources
  Tests
  Configurations
  Products
  Frameworks
```
With out `groupOrder` option, it will be sorted alphabetically, to set the order as above we can use the new option in this way:
```
options:
  groupsOrder:
    - order: [Sources, Resources, Tests, Support files, Configurations]
    - pattern: '^.*Screen$'
      order: [View, Presenter, Interactor, Entities, Assembly]
```


brentleyjones:
Is this only at the top level? How about being able to do it on sub-levels as well?

Beniamiiin:
> Is this only at the top level? How about being able to do it on sub-levels as well? Maybe that is covered by your second example, I'm not sure.

No, it's work on any level. You are right, it's covered by the second example.

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -376,23 +376,33 @@ public class PBXProjGenerator {
...
-                let sortOrder2 = child2.getSortOrder(groupSortPosition: project.options.groupSortPosition)
-
-                if sortOrder1 != sortOrder2 {
-                    return sortOrder1 < sortOrder2
-                } else {
-                    if child1.nameOrPath != child2.nameOrPath {
-                        return child1.nameOrPath.localizedStandardCompare(child2.nameOrPath) == .orderedAscending
+        if let groupOrder = project.options.groupsOrder.first(where: { group.nameOrPath == $0.pattern || group.nameOrPath.isMatch(to: $0.pattern) }) {
+            let files = group.children.filter { $0 is PBXFileReference }
+            let groups = groupOrder.order
```
yonaskolb: If someone adds a directory or target, it won't be included as a group unless specifically specified in the groupsOrder. Is that intended?

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -376,23 +376,33 @@ public class PBXProjGenerator {
...
-                let sortOrder1 = child1.getSortOrder(groupSortPosition: project.options.groupSortPosition)
-                let sortOrder2 = child2.getSortOrder(groupSortPosition: project.options.groupSortPosition)
-
-                if sortOrder1 != sortOrder2 {
-                    return sortOrder1 < sortOrder2
-                } else {
-                    if child1.nameOrPath != child2.nameOrPath {
-                        return child1.nameOrPath.localizedStandardCompare(child2.nameOrPath) == .orderedAscending
+        if let groupOrder = project.options.groupsOrder.first(where: { group.nameOrPath == $0.pattern || group.nameOrPath.isMatch(to: $0.pattern) }) {
+            let files = group.children.filter { $0 is PBXFileReference }
```
yonaskolb: Is there any reason files couldn't be sorted? What if all the sorts were sorted first and then everything else came after? I guess there could be an option to put everything else first as well

Sources/ProjectSpec/StringExtensions.swift
```diff
@@ -0,0 +1,11 @@
+import Foundation
+
+public extension String {
+    
+    func isMatch(to regex: String) -> Bool {
+        guard let regex = try? NSRegularExpression(pattern: regex) else { return false }
```
yonaskolb: Creating a `NSRegularExpression` everytime is slow. The `GroupOrder` could just have an optional `pattern: NSRegularExpression?` instead of a `pattern: String`

Docs/ProjectSpec.md
```diff
@@ -114,6 +114,7 @@ Note that target names can also be changed by adding a `name` property to a targ
   - `none` - sorted alphabetically with all the other files
   - `top` - at the top, before files
   - `bottom` (default) - at the bottom, after other files
+- [ ] **groupsOrder**: **[[GroupOrder]](#groupOrder)** - An order of groups.
```
yonaskolb: Not in love with the name `groupsOrder`, but I also can't think of something better right now.

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -376,23 +376,33 @@ public class PBXProjGenerator {
...
-            .sorted { child1, child2 in
-                let sortOrder1 = child1.getSortOrder(groupSortPosition: project.options.groupSortPosition)
-                let sortOrder2 = child2.getSortOrder(groupSortPosition: project.options.groupSortPosition)
-
-                if sortOrder1 != sortOrder2 {
-                    return sortOrder1 < sortOrder2
-                } else {
-                    if child1.nameOrPath != child2.nameOrPath {
-                        return child1.nameOrPath.localizedStandardCompare(child2.nameOrPath) == .orderedAscending
+        if let groupOrder = project.options.groupsOrder.first(where: { group.nameOrPath == $0.pattern || group.nameOrPath.isMatch(to: $0.pattern) }) {
```
yonaskolb: `nameOrPath` evaluates to `name ?? path ?? ""`. That means groups who have a different path than just the name won't be supported.
In either case doing something like `Group1/Group2/*` won't work as the name is just the end and the path is a relative path.

brentleyjones:
>How important is it to your team to have this sorting? Both for the top level and for sub groups?

I'll say for us that the top level is pretty important. We've gone the route of adding `1.`, `2. `, etc. to the directory names to get the sorting. For context, we are grouping similar targets into directories (applications, experiences, capabilities, components, services, core frameworks, etc.), since we have a massive number of targets.

Beniamiiin:
> Thanks for the PR @Beniamiiin. How important is it to your team to have this sorting? Both for the top level and for sub groups?

For us, it's important because currently, we use my fork with the implementation of this feature.

Sources/XcodeGenKit/PBXProjGenerator.swift
Beniamiiin: No, it was not supposed, I'll think how to fix it.

Sources/XcodeGenKit/PBXProjGenerator.swift
Beniamiiin: You know, I just forgot to sort files. I divided files and groups and sorted groups but forgot to sort files. So I can sort files and groups and after this merge it by `fileSortPosition` option.

Sources/ProjectSpec/StringExtensions.swift
Beniamiiin: Yes, you are right, I'll fix it.

Docs/ProjectSpec.md
Beniamiiin: Not only you, I don't like it too, but I couldn't find more lovely name :( I'll try to think about it again.

Sources/XcodeGenKit/PBXProjGenerator.swift
Beniamiiin: Okay, so should I to use just `name`?

Sources/XcodeGenKit/PBXProjGenerator.swift
Beniamiiin: I've updated PR and it works this way.
1. Everything sort as was previously, i.e. using `sortGroups` functions.
2. Setup groups order for provided groups
3. Not provided groups put at the end of the main group.

**Example**

I have a project with a structure:

```
Test.xcodeproj
  Sources
  Resources
  Tests
  Support files
  Configurations
  Products
  Frameworks
```

Xcodegen spec:

```
options:
  groupsOrder: 
    - order: [Sources, Resources, Tests]
```

The result will be looks like:

```
Test.xcodeproj
  Sources
  Resources
  Tests
  Configurations
  Support files
  Products
  Frameworks
```

**Explanation**

First, we sort groups alphabetically, i.e. `Configurations, Resources, Support files, Sources, Tests` after this we setup group order with provided group names `[Sources, Resources, Tests]` and get result `Sources, Resources, Tests` and remaining groups `Configurations, Support files` put at the end.

Sources/XcodeGenKit/PBXProjGenerator.swift
Beniamiiin: @yonaskolb I need your help. I don't understand what is the correct way to do it.

brentleyjones:
@Beniamiiin @yonaskolb What is needed to move this forward. We personally would love this feature landed.

Beniamiiin:
@brentleyjones firstly I need some help from @yonaskolb to fix all review issues and secondly @yonaskolb should review the PR again.

Beniamiiin:
@yonaskolb are we gonna do something with this PR?

Beniamiiin:
@yonaskolb are we gonna do something with this PR?

brentleyjones:
I'm still a big fan of this change as well. @yonaskolb, since you had some PR comments I'll leave it to you to re-review. Thanks!

Beniamiiin:
@yonaskolb I'm so sorry can you give us some information about this PR?

Docs/ProjectSpec.md
brentleyjones: How about `groupOrdering`?

brentleyjones:
It runs nicely with our project, so just looking for the requested changed and we should be good to go.

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -566,6 +567,45 @@ public class PBXProjGenerator {
...
+            var filteredGroups = [PBXFileElement]()
+            
+            for groupName in order {
+                guard let group = groups.first(where: { $0.nameOrPath == groupName }) else {
+                    continue
+                }
+                
+                filteredGroups.append(group)
+                groups.removeAll { $0 == group }
+            }
```
yonaskolb, line 599: I think this will cause performance issues in large projects that use a regex pattern, as there are a lot of nested loops here, but I guess it's up to the project author to take on that cost

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -566,6 +567,45 @@ public class PBXProjGenerator {
...
+        }
+        
+        if let order = groupOrdering?.order, let fileSortPosition = groupOrdering?.fileSortPosition {
+            let files = group.children.filter { $0 is PBXFileReference }
+            var groups = group.children.filter { $0 is PBXGroup }
+            
+            var filteredGroups = [PBXFileElement]()
+            
+            for groupName in order {
+                guard let group = groups.first(where: { $0.nameOrPath == groupName }) else {
```
yonaskolb, line 593: If this worked on just `group.children` it would be possible to include file names as well

Docs/ProjectSpec.md
```diff
@@ -133,6 +134,27 @@ options:
...
+### GroupOrdering
+
+Describe an order of groups. Available parameters:
+
+- [ ] **pattern**: **String** - A group name pattern. Can be just a single string and also can be a regex pattern. Optional option, if you don't set it, it will pattern for the main group, i.e. the project.
+- [ ] **order**: **[String]** - An order of groups.
+- [ ] **fileSortPosition**: **String** - Where files are sorted in relation to other groups. Either:
+  - `top` (default) - at the top, before groups
+  - `bottom` - at the bottom, after other groups
+
```
yonaskolb, line 146: Is this required? Would someone ever want it to be different than `options.groupSortPosition`

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -566,6 +567,45 @@ public class PBXProjGenerator {
         let childGroups = group.children.compactMap { $0 as? PBXGroup }
         childGroups.forEach(sortGroups)
     }
+    
+    func setupGroupOrdering(group: PBXGroup) {
```
yonaskolb, line 571: Would be great to have a unit test for this

brentleyjones:
@Beniamiiin after @yonaskolb's feedback is addressed and this is rebased on master, we can merge it. Thanks!

Beniamiiin:
> @Beniamiiin: after @yonaskolb's feedback is addressed and this is rebased on master, we can merge it. Thanks!

@brentleyjones I'll do it at this weekend.

Sources/XcodeGenKit/PBXProjGenerator.swift
Beniamiiin: If we try to do the ordering for files as well, we'll face the intersection of `groupOrdering` and `groupSortPossition` options.

For example, we have the following directory:
```
MyFile1(File)
MyFile2(File)
Sources(Group)
Resources(Group)
Tests(Group)
Support files(Group)
Configurations(Group)
```
If I set up options the following way:
```
groupSortPossition: top
groupOrdering:
    - order: [MyFile1, Sources, Support files]
```
Do we expect to see the following result?
```
MyFile2(File)
MyFile1(File)
Sources(Group)
Support files(Group)
Resources(Group)
Tests(Group)
Configurations(Group)
```
And if we setup in this way?
```
groupSortPossition: bottom
groupOrdering:
    - order: [MyFile1, Sources, Support files]
```
Do we expect to see the following result?
```
MyFile1(File)
Sources(Group)
Support files(Group)
Resources(Group)
Tests(Group)
Configurations(Group)
MyFile2(File)
```

It's a bit tricky, I'd advise not to implement file ordering.

brentleyjones: I agree. File ordering is way less useful.

Beniamiiin:
@yonaskolb can you review it again please

mthole:
@Beniamiiin Found this feature today. Very useful, thank you!

Issue #576, opened 2019-04-30
Allow static libraries to link dynamic frameworks


peter-tomaselli:
We have an iOS app that we’re trying to migrate over to XcodeGen. Our `App` target links to many `Feature`s built as static libraries. Features can each link a common set of `CoreLibrary`s (for example), which are built as dynamic frameworks.

I’ve been digging in to the code and issue history here trying to understand why this [seems to be specifically disallowed](https://github.com/yonaskolb/XcodeGen/blob/master/Sources/XcodeGenKit/PBXProjGenerator.swift#L507) in XcodeGen. But I don’t understand why! It seems to be something related to the new build system perhaps? But as far as I can tell, whatever restriction used to be in place has been removed in Xcode 10.2, as this setup has been working fine for us.

For reference, if we comment out the line linked above, we get exactly the behavior we need. It also looks like there’s a `link` flag on `Dependency` that may be intended to override this behavior (eg. if the line above is changed to
```swift
if target.type == .staticLibrary && dependency.link != true { break }
```
we could also get what we need), but I’m not sure what the original design intention was there. 

Is either of these changes a viable approach? If so, I’d be happy to submit a PR. Thank you for your time!

yonaskolb:
@brentleyjones do you have any comments as you worked on this? I'd be happy for either of these solutions, but as I'm unclear on what the restriction was for, maybe the second `link` flag solution is a safer default

gomsu:
I have a similar problem where I have a framework that embeds multiple features that are static libraries. These static libraries are using a "core static library" for base features. I can't make the xcodegen to generate the "Link Binary With Libraries" build phase for the "feature static libraries" to link to the "core static library", so that the core would always get built first with any of the "feature static libraries". I can though add this build phase from the Xcode and add the "core static library"  to it and everything works after that, so it should be possible to generate it also.

brentleyjones:
Static libraries can't link dynamic libraries (Xcode will correctly warn about it). You can depend on them, and as long as Xcode can find the headers and module maps you can use them in your static library, but they have to be linked to the same executable (which is possibly a dynamic library) that the static library ends up being linked to.

Our targets are almost exclusively static libraries, and we can depend on dynamic libraries just fine with XCG as it is now.

vlozko:
> Static libraries can't link dynamic libraries or framework (Xcode will correctly warn about it).

[...]

PR #616, opened 2019-07-24
Add bundle dependency type and re-enable static library linking against frameworks

bsmith11:
This PR does two things:

- Add a `bundle` dependency type. When using static libraries, in order to include resources (`.xib`, `.storyboard`, etc), we must also have a separate bundle target. Anytime the static library is linked against, we must also copy its associated bundle target. Conceptually, both are dependencies. When all targets are within the same project, we can use the `target` dependency type. This adds support for static lib + bundle targets that are part of a separate project from the root project that depends on them

- This also re-adds support for static libraries linking against dynamic frameworks. When they are each  a part of a separate projects, simply adding the framework to the framework search paths doesn't suffice, since it does not create the dependency in Xcode which tells the build system to build one before the other.

https://github.com/yonaskolb/XcodeGen/issues/576


TomasLinhart:
Thank you for your changes. I face the same problem as you, so it is great to see some changes to make it work. Unfortunately, your changes work for me only when the bundle is part of the same project. I am unable to use implicit bundles that are created as part of another project.

bsmith11:
Could you post the relevant snippet from your `project.yml` file?

TomasLinhart:
I have `dependencies:` with the following content in a target with a type `bundle.unit-test`.
```
- bundle: BundleThatIsPartOfThisProject # This one appears in Copy Bundle Resourses
- bundle: AnotherBundleInAnotherProject # This one does not apper in the copy phase
```

bsmith11:
Try adding the full filename of the bundle product:

```
dependencies:
     - bundle: MyBundle.bundle
```

TomasLinhart:
🤦‍♂ I had a wrong path to the xcodegen CLI, so it was not using the one I built from your source code. It works now. Sorry for the panic. 😄

yonaskolb:
Thanks @bsmith11! Looks good to me. Could you:
- add 2 changelog entries
- add a bundle and a static library including a dynamic framework to Fixtures/TestProjec and check there are no warnings. CI will also build the iOS and macOS app targets to make sure it builds properly

yonaskolb:
The test is failing because it looks like the `Bundle.bundle` file reference is not having its UUID resolved for some reason.

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -194,6 +195,17 @@ public class PBXProjGenerator {
             )
             derivedGroups.append(group)
         }
+        
+        if !bundleFiles.isEmpty {
```
bsmith11: This creates a derived group for bundles that are dependencies, similar to the `Frameworks` and `Products` derived groups. We need to do this because otherwise the file reference temporary IDs (`TEMP_blahblahblah`) will not be deterministically generated by the `XcodeProj.ReferenceGenerator`, thus resulting in a new reference value each time `XcodeGen` is run. This is how `XcodeProj` generates these IDs for implicitly linked frameworks as well.

Tests/Fixtures/TestProject/build.sh
```diff
@@ -1,13 +0,0 @@
-#!/bin/bash
```
bsmith11: Moved to directory with the workspace

Tests/Fixtures/TestProject/project.yml
```diff
@@ -91,6 +91,11 @@ targets:
       - target: App_watchOS
       - target: iMessageApp
       - sdk: Contacts.framework
+      - framework: libProject2StaticLibrary.a
```
bsmith11: New features are specifically to add multi-project support

Tests/Fixtures/build.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+set -e
+
+cd TestProject
```
bsmith11: I added this cd and the one after the carthage line

Tests/Fixtures/build.sh
```diff
@@ -0,0 +1,16 @@
...
+carthage bootstrap --cache-builds
+cd ../
+
+echo "
+⚙️ Building iOS app"
+xcodebuild -quiet -workspace TestWorkspace.xcworkspace -scheme "App_iOS Test" -configuration "Test Debug" CODE_SIGN_IDENTITY="" CODE_SIGNING_REQUIRED=NO CODE_SIGN_ENTITLEMENTS="" CODE_SIGNING_ALLOWED="NO"
```
bsmith11: Updated this to specify `-workspace` instead of `-project`

Docs/ProjectSpec.md
```diff
@@ -364,24 +364,25 @@ targets:
...
 ### Dependency
 
-A dependency can be one of a 3 types:
+A dependency can be one of a 5 types:
 
 - `target: name` - links to another target
 - `framework: path` - links to a framework
 - `carthage: name` - helper for linking to a Carthage framework
 - `sdk: name` - links to a dependency with the SDK. This can either be a relative path within the sdk root or a single filename that references a framework (.framework) or lib (.tbd)
+- `bundle: name` - adds the pre-built bundle for the supplied name to the copy resources build phase. This is useful when a dependency exists on a static library target that has an associated bundle target, both existing in a separate project. Only usable in targets types which can copy resources. Always use `implicit: true`. This will will not cause the bundle to be explicitly built.
```
moayes: For better readability: Only usable in `target` types which ...
Typo: `This will will ...`

yonaskolb: 👍

yonaskolb:
THanks @bsmith11, the workspace changes look good. Just a typo to fix as @moayes pointed out

Docs/ProjectSpec.md
yonaskolb: What happens if someone doesn't put `implicit`? Should this be the implied default, so people don't have to set it?

bsmith11:
👍 , I should have some time to clean these things up this week

yonaskolb:
Hi @bsmith11, would be great to get this into the release this weekend!

CHANGELOG.md
```diff
@@ -3,22 +3,26 @@
...
+#### Changed
+- Updated static library targets to be able to link against dynamic frameworks to support dependencies in other projects [#616](https://github.com/yonaskolb/XcodeGen/pull/616) @bsmith11
 
 ## 2.7.0
 
 #### Added
-
 - Added Bash 4 style recursive globbing (`**/*`) in target sources `excludes` [#636](https://github.com/yonaskolb/XcodeGen/pull/636) @bclymer
 - Added ability to disable main thread checker in Schemes [#601](https://github.com/yonaskolb/XcodeGen/pull/601) @wag-miles
+- Added new dependency type, `bundle`. This allows targets to copy bundles from other projects [#616](https://github.com/yonaskolb/XcodeGen/pull/616) @bsmith11
```
yonaskolb: I think there was an incorrect conflict resolution here

yonaskolb:
The tests are failing because the fixtures need to be regenerated

bsmith11:
I think its all good to go now. I had to update my stuff to work with https://github.com/yonaskolb/XcodeGen/pull/635 (great change btw!)

Tests/Fixtures/TestProject/Project.xcodeproj/project.pbxproj
```diff
@@ -3371,6 +3409,10 @@
 		66DBF782276810342031776D /* Staging Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
+				FRAMEWORK_SEARCH_PATHS = (
+					"$(inherited)",
+					"\".\"",
```
yonaskolb: Where does this change come from and is it required?

simonpierreroy:
@bsmith11 Can I become a collaborator on that Fork of the repo? :)  I would like to help bring that PR to the finish line. c.c. @yonaskolb

yonaskolb:
This PR adds the bundle dependency type https://github.com/yonaskolb/XcodeGen/pull/723

I'm closing this for now. If someone wants to add support for static libraries linking against dynamic frameworks, then create a new PR for that. @simonpierreroy, sounds like you might want to give this a go?

simonpierreroy:
Great. Thank you.

Issue #606, opened 2019-06-29
SPM packages and Test Plans


ethanjdiamond:
Hi! I was wondering if XcodeGen was going to support two new features I'd like to try out with Xcode 11.
- [SPM packages](https://developer.apple.com/videos/play/wwdc2019/408/)
- [Test Plans](https://developer.apple.com/videos/play/wwdc2019/413/). 

I think I saw that xcodeproj support packages now.

yonaskolb:
Yeah I've been working a bit on SPM packages in my spare time. Here's a preview

<img width="458" alt="Screen Shot 2019-06-30 at 1 54 28 pm" src="https://user-images.githubusercontent.com/2393781/60392395-193d3080-9b45-11e9-910c-c8c88d09b734.png"> 
<img width="983" alt="Screen Shot 2019-06-30 at 1 55 09 pm" src="https://user-images.githubusercontent.com/2393781/60392394-193d3080-9b45-11e9-8abc-2c57c4760976.png">

Test Plans first needs XcodeProj support

kdubb:
@yonaskolb Any thoughts on a alpha/beta release with SPM support. This solves _a lot_ of issues for us.

On a general note... kudos to you for creating and maintaining this project. Once discovered we've found it completely indispensable. Thank you!

AlexisQapa:
Started using Xcode 11 with SPM thinking this is great but it will be hard to forget XcodeGen and then I saw this issue.

bonkey:
Wonderful! :)

rvanmelle:
Yeah ... thanks.  Funny, also was playing around and had a working prototype locally, very similar syntax, but I think I like yours better: 
```packages:
  Yams:
    package: https://github.com/jpsim/yaml
    version: 2.0.0 < NextMajor
  Yams:
    package: https://github.com/jpsim2/yaml
    version: 2.0.0 < NextMinor
  Yams:
    package: https://github.com/jpsim2/yaml
    version: 2.0.0 < 3.0.0
  Yams:
    package: https://github.com/jpsim2/yaml
    version: 2.0.0
  Yams:
    package: https://github.com/jpsim2/yaml
    branch: feature/cool
  Yams:
    package: https://github.com/jpsim2/yaml
    commit: afqwerqwefasdf
```

zdnk:
What about the issue with resolving packages since `Package.resolved` file is inside the generated `xcodeproj`?

rvanmelle:
[...]

PR #617, opened 2019-07-26
Support Swift Packages

toshi0383:
#606

I personally needed this feature, so implemented.
Looks like Yonas already has an implementation, so feel free to just close this one, but otherwise people can use this branch for now.

If you want me to keep going, I will clean up the dirty codes and add tests etc.👌

## project spec consideration

I chose yml format to be rather **simple than elegant**, using `XCRemoteSwiftPackageReference` directly.

Sample `project.yml`:

```yml
name: App
targets:
  App:
    type: application
    platform: iOS
    dependencies:
      - swiftpm:
        repositoryURL: "https://github.com/tuist/xcodeproj"
        versionRequirement:
          kind: upToNextMajorVersion
          minimumVersion: 7.0.0
```

See: https://github.com/tuist/xcodeproj/blob/58c6acf1111b92a9acf633359a10249cf7940045/Sources/xcodeproj/Objects/SwiftPackage/XCRemoteSwiftPackageReference.swift

## swiftpm process

SwiftPM's package resolving and fetching process begins when you open the generated xcodeproj with Xcode11beta.

### DIFF
--- a/Sources/ProjectSpec/Dependency.swift
+++ b/Sources/ProjectSpec/Dependency.swift
@@ -1,5 +1,6 @@
 import Foundation
 import JSONUtilities
+import XcodeProj
 
 public struct Dependency: Equatable {
     public static let removeHeadersDefault = true
@@ -37,6 +38,7 @@ public struct Dependency: Equatable {
         case target
         case framework
         case carthage(findFrameworks: Bool?)
+        case swiftpm(config: XCRemoteSwiftPackageReference?)
         case sdk(root: String?)
     }
 }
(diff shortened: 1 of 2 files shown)


yonaskolb:
Thanks for the PR @toshi0383. I have indeed gone another approach, most notably separating the definition of a package from linking to it. This allows for multiple targets to use the same package. Coming back from a break so I'll open this soon

yonaskolb:
Swift Package support has been added here https://github.com/yonaskolb/XcodeGen/pull/624

Closing this PR now

PR #611, opened 2019-07-22
Added custom group support for target sources (#478)

sroebert:
As I really need custom grouping support in my team, for keeping a proper structure in our large project, I was looking forward to using #607. Unfortunately I saw some issues with this implementation, specifically the flattening of groups of directories, which was originally solved by #550. Also having multiple target sources in a custom group with a depth larger than 1, did not really seem to work.

For this reason I created this pull request where both issues are solved. This should implement the request mentioned in #478, while keeping the same behaviour for target sources as before.

PR #621, opened 2019-08-07
Add support for source groups

### LINKED ISSUE #478: Add a custom group for a target source

rcari:
Rework of #611 / Fixes #478

Given that @sroebert is on vacation and I really need this, I provided the fixes requested by @yonaskolb.
I also fixed some behaviors to something I believe makes more sense:
- If a file has a custom group, it should appear only in the custom hierarchy and no other containing folder should be added to that hierarchy
- If a folder has a custom group, it should appear as a subfolder of that custom hierarchy
- If a custom group hierarchy maps to the filesystem hierarchy it should be folder references and not simple groups


Sources/XcodeGenKit/SourceGenerator.swift
```diff
@@ -526,6 +527,9 @@ class SourceGenerator {
                 sourcePath = path
                 sourceReference = fileReference
                 rootGroups.insert(fileReference)
+            } else if hasCustomParent {
+                sourcePath = path
```
yonaskolb: If the path is something like `File.file` then the parent would be the base path, and then it would fall into the first if and not this one. Maybe `hasCustomParent` should be checked for first?

Sources/XcodeGenKit/SourceGenerator.swift
```diff
@@ -618,6 +626,29 @@ class SourceGenerator {
         }
     }
 
+    private func fixRelativePath(for fileElement: PBXFileElement, at path: Path) throws {
```
yonaskolb: What is this trying to fix?

rcari: It fixes the fileElement path and name based on the parent's aggregated paths so that its path becomes relative to the parent and not to the project. Xcode then rebuilds the actual path for the file based on the hierarchy this fileElement lives in.

Sources/XcodeGenKit/SourceGenerator.swift
rcari: I'm looking into it!

Sources/XcodeGenKit/SourceGenerator.swift
yonaskolb: Could you add some comments to that effect

Sources/XcodeGenKit/SourceGenerator.swift
rcari: Fixed!

Sources/XcodeGenKit/SourceGenerator.swift
rcari: Comments added

rcari:
@yonaskolb do you have further comments on this PR ?

CHANGELOG.md
```diff
@@ -6,9 +6,10 @@
 
 - Added support for Swift Package dependencies [#624](https://github.com/yonaskolb/XcodeGen/pull/624) @yonaskolb
 - Added `includes` to `sources` for a Target. This follows the same glob-style as `excludes` but functions as a way to only include files that match a specified pattern. Useful if you only want a certain file type, for example specifying `**/*.swift`. [#637](https://github.com/yonaskolb/XcodeGen/pull/637) @bclymer
-- Support `dylib` SDK. [#650](https://github.com/yonaskolb/XcodeGen/pull/650) @kateinoigakukun 
-- Added `language` and `region` options for `run` and `test` scheme [#654](https://github.com/yonaskolb/XcodeGen/pull/654) @kateinoigakukun 
-- Added `debugEnabled` option for `run` and `test` scheme [#657](https://github.com/yonaskolb/XcodeGen/pull/657) @kateinoigakukun 
+- Support `dylib` SDK. [#650](https://github.com/yonaskolb/XcodeGen/pull/650)
```
yonaskolb: Looks like some names were lost in the merge conflict resolution

Sources/XcodeGenKit/SourceGenerator.swift
```diff
@@ -391,6 +391,8 @@ class SourceGenerator {
 
         let children = try getSourceChildren(targetSource: targetSource, dirPath: path, excludePaths: excludePaths, includePaths: includePaths)
 
+        let createIntermediateGroups = targetSource.createIntermediateGroups ?? project.options.createIntermediateGroups
```
yonaskolb: Now that this property is created, can we remove the parameter, or is that still required?
On the other hand, on hindsight I'm ok with just using the parameter, so the fallback logic is in a single place

Sources/XcodeGenKit/SourceGenerator.swift
```diff
@@ -532,7 +534,7 @@ class SourceGenerator {
 
         let type = targetSource.type ?? (path.isFile || path.extension != nil ? .file : .group)
 
-        let customParentGroups = (targetSource.group ?? "").split(separator: "/")
+        let customParentGroups = (targetSource.group ?? "").split(separator: "/").map{ String.init($0) }
```
yonaskolb: nitpick: don't need the `.init`

Sources/XcodeGenKit/SourceGenerator.swift
yonaskolb: It's still not really clear what this is doing, and why it's required. `Fix` implies something is broken. Could it have a better name, like `syncDirectories` or something, and also include _why_ this is required.

What do you mean by `Xcode then rebuilds the actual path for the file based on the hierarchy this fileElement lives in` Does this mean there is a diff before and after Xcode saves the project?

Also could you add a test that would fail if this wasn't applied?

Sources/XcodeGenKit/SourceGenerator.swift
rcari: I preferred removing the parameter altogether and get the data from where it comes the target source itself or the project.

rcari:
@yonaskolb everything is fixed!

Sources/XcodeGenKit/SourceGenerator.swift
rcari: I updated the method name to better reflect what it's doing. It's making paths relative to the containing folders instead of to the projet root.

PR #624, opened 2019-08-13
Support for Swift Packages

yonaskolb:
This adds support for defining and linking to Swift Packages

```yaml
packages:
  Yams:
    url: https://github.com/jpsim/Yams
    majorVersion: 2.0.0
  SwiftPM:
    url: https://github.com/apple/swift-package-manager
    branch: swift-5.0-branch
targets:
  App:
    dependencies:
      - package: Yams 
      - package: SwiftPM
        product: SPMUtility
```
See the ProjectSpec diff for more documentation https://github.com/yonaskolb/XcodeGen/pull/624/files#diff-938b5cf412212f1137a53f072512963a

Please test this out and report any issues. 

One thing that is broken as of Xcode 11 beta 5 is that packages don't work if you are using configurations other than the default `Debug` and `Release`. That SPM bug is tracked here https://bugs.swift.org/browse/SR-10927


Docs/ProjectSpec.md
```diff
@@ -747,3 +769,28 @@ schemes:
       customArchiveName: MyTarget
       revealArchiveInOrganizer: false
 ```
+
+## Swift Package
+Swift packages are defined at a project level, and then linked to individual targets via a [Dependency](#dependency).
+
+> Note that Swift Packages don't work in projects with configurations other than Debug and Release
```
brentleyjones: Can we add a link to https://bugs.swift.org/browse/SR-10927?

mmmilo:
Is it possible to allow for adding multiple products?

One use case is that I am looking to add a target that contains mock/test data.

brentleyjones:
Because these are statically linked, we should treat them like `target` dependencies that are static libraries. What this means is that if I declare the following:

```yaml
packages:
  Yams:
    url: https://github.com/jpsim/Yams
    majorVersion: 2.0.0
targets:
  App:
    dependencies:
      - target: StaticLibraryA
      - package: Yams
  StaticLibraryA:
    dependencies:
      - package: Yams
```

Then `Yams` should only be linked to `App`, not `StaticLibraryA`, otherwise we will get duplicated symbols.

We also need to add the packages as Target Dependencies (especially once you make the above change, since Xcode won't know to build the SPM package before building `StaticLibraryA`):

image

brentleyjones:
For example, without the above change we get the following on one of our static library targets:

image

kdubb:
This works great for us! Thanks!

Supporting local packages would help us immensely when it comes to project maintenance. This seems like it should be fairly easy since Xcode 11 already fully supports local packages.

A little background for use case...

Our main usage of this feature is to generate xcode projects for our Swift Package Manager projects that need to be tested on multiple platforms via command line; especially those with special requirements.

For example, our security framework requires the keychain access entitlement on all platforms except macOS to run unit tests. To facilitate CI we generate a project using `xcodegen` that has host applications setup for each of the test targets and those host applications are granted keychain access.  We then simply use `xcodebuild` to build and run tests for each platform.

Now that xcodegen supports package dependencies we can generate a fairly straightforward project for the framework, then we add our test & test host targets.  Unfortunately this means we are duplicating package dependencies that are already defined in our `Package.swift`. While this is _way_ better than previous, it requires us to keep dependency versions up to date in 2 places, `Package.swift` and `project.yml`.

Supporting local packages would allow us to generate an `xcodeproj` that contains _only_ test targets & test host targets.  They would reference a package dependency to the local folder (e.g. `.`) which defines the framework itself as a package. This would be solve maintaining dependencies in two places as well as test the framework _as_ as Swift package which is how it will be consumed.

mbuchetics:
With Xcode 11 Final right around the corner ... are there any plans on merging this soon into master with an official release?

Docs/ProjectSpec.md
yonaskolb: Done

yonaskolb:
Ok I've:
- rebased
- added some more docs
- prevented linking to static libs
- added a commented out stub for adding package target dependencies. It requires this PR I created in XcodeProj first https://github.com/tuist/XcodeProj/pull/481. Interestingly Xcode doesn't do this if you add a dependency via it's UI, so I wouldn't say that should stop this PR from being merged and released.

@brentleyjones could you look this over again, and make sure it works with your static library setup

---
@mmmilo 
> Is it possible to allow for adding multiple products?

To accomplish that you can use multiple dependency declarations. This allows us to control settings for a single linkage if needed
```yaml
dependencies:
  - package: SomeDependency
    product: SomeDependency # this isn't required
  - package: SomeDependency
    product: SomeDependencyTestData
```

---
@kdubb
> Supporting local packages would help us immensely when it comes to project maintenance. This seems like it should be fairly easy since Xcode 11 already fully supports local packages.

Yes this will be possible. It isn't in this PR but will come in a followup one. It's just adding a path as a PBXFileReference and Xcode takes care of the rest.

---
@mbuchetics 
> With Xcode 11 Final right around the corner ... are there any plans on merging this soon into master with an official release?

Yes, I'd like to get this released as soon as possible. If you could test it on your setup that would help.

[Mint](https://github.com/yonaskolb/Mint) makes this easy:
```
mint run yonaskolb/xcodegen@swift_packages xcodegen generate
```

yonaskolb:
I've added support for package target dependencies for proper build dependency resolution.
I've also added local swift package support. For the moment it doesn't support linking these unless they mirror remote packages

mbuchetics:
I just experimented with it (see https://github.com/allaboutapps/ios-starter/tree/spm) and it works great! 
However, I ran into https://bugs.swift.org/browse/SR-10927 again ... this is a huge issue for us and we won't be able to use SPM until this is fixed. Can't believe this made it into 11.1 :(

brentleyjones:
@yonaskolb I'll start reviewing this shortly. I'll let you know!

yonaskolb:
Released in 2.8.0!

Issue #310, opened 2018-05-22
Default exclude git merge files


m-ruhl:
Hi 👋 ,

Does it makes sense to default exclude all git merge files `*.orig` ?
I'm not sure anymore would use does in a Xcode Project.

I can create a PR, but I wanted to check first 😉 

Greets
Michael

yonaskolb:
Hi @m-ruhl 
I agree git merge files shouldn't be added to the project. But are there workflows where you want to keep these around anyway? Wouldn't they get deleted once you've completed your merge?

I see no harm in adding them to the default ignored files however. Feel free to create a PR if it's an issue for you 😄 👍

keith:
I submitted a fix for this here https://github.com/yonaskolb/XcodeGen/pull/627

PR #627, opened 2019-08-15
Ignore files with orig extension

keith:
This is useful for multi-developer projects where you don't want to
delete these files for everyone, since they might actually want them,
but they still aren't useful to show in Xcode.

Fixes https://github.com/yonaskolb/XcodeGen/issues/310


yonaskolb:
Thanks @keith! Could you please:
- add a changelog entry
- add an `ignores files` test here https://github.com/yonaskolb/XcodeGen/blob/master/Tests/XcodeGenKitTests/SourceGeneratorTests.swift

keith:
Updated with a new test case and changelog entry!

Issue #180, opened 2017-12-02
Support for recursive Source excludes


peymankh:
Currently Source excludes are implemented with using glob patterns.

#### Examples

- `ios/*.swift`: all files with full path that ends with .swift
- `ios/configs`: all files and directories under ios/configs
- `README.md`: the README.md file in the root directory of your repository (doesn’t match ios/README.md)
- `ios/*.[mh]`: files with the extension .m or .h only under the ios directory (but not its subdirectories)
- `configs/server[0-2].json`: the files server0.json, server1.json, server2.json in the directory configs/

Aim is to add a more robust excludes support using `**`. 

#### Examples

- `ios/**`: all files with full path that starts with ios/ (i.e. any files in the ios directory)
- `**/README.md`: the README.md file in all directories (including README.md in root directory of your repository)
- `ios/**/*.[mh]`: all files with the extension .m or .h, in the ios directory and all of its subdirectories

[Examples Disclaimer](https://docs.buddybuild.com/builds/selective_builds.html#examples)

faken:
+1

keehun:
Are there any plans for supporting the `**` rule?

PR #636, opened 2019-08-28
Add recursive glob support

bclymer:
Fixes https://github.com/yonaskolb/XcodeGen/issues/180.

`Glob.swift` comes from R.swift https://github.com/mac-cain13/R.swift/blob/master/Sources/RswiftCore/Util/Glob.swift, which was previously taken from other OS projects. Credits are provided in the header.

`Glob` itself comes with numerous tests, but I was also able to activate the previously commented out recursive globstar tests in XcodeGen. In addition, I added another scenario of a recursive globstar in a subdirectory, ensuring it does not match all directories, only those specified.


Sources/XcodeGenKit/SourceGenerator.swift
```diff
@@ -317,6 +317,21 @@ class SourceGenerator {
...
     }
 
+    private func expandPattern(_ pattern: String) -> [Path] {
+        let filePaths = listFilePaths(pattern: pattern)
+        let urls = filePaths.map { Path($0) }
+
+        return urls
+    }
+
+    private func listFilePaths(pattern: String) -> [String] {
```
yonaskolb: This class already has too many functions (needs a refactor). For now can we just combine these two functions into one? Our basic currency type is always Path

acecilia:
Thanks! @bclymer would it be possible to have the Glob code and tests as an independent Github project and swift package, so instead of having the source code in the project we can just add it to the `Package.swift` manifest?

bclymer:
@yonaskolb sorry for the delay, I made the changes you requested. 

@acecilia While I think that's a good idea, it feels a bit strange to me since it's not actually my code. Tracing back the links in the header comment it looks like it's been adapted from a number of gists, and then it moved to R.swift and has been modified there since.

yonaskolb:
Thanks @bclymer! I just noticed some errors in the log from the new glob tests:

```
Error parsing file system item: Error Domain=NSCocoaErrorDomain Code=260 "The folder “dir2” doesn’t exist." UserInfo={NSFilePath=/tmp/glob-test.XBQ0m/dir2/, NSUserStringVariant=(
    Folder
), NSUnderlyingError=0x7fa8cdb00f20 {Error Domain=NSOSStatusErrorDomain Code=-43 "fnfErr: File not found"}}
```
https://github.com/yonaskolb/XcodeGen/pull/636/checks?check_run_id=208818534

bclymer:
@yonaskolb seeing that locally as well. I just didn't scroll up enough to see it originally. I'll get it resolved.

acecilia:
@bclymer okey. I also think is a good idea, since I have seen already several projects rolling their own Glob implementations.

bclymer:
Pushed a fix. The error was semi-intentional. It was testing a double globstar (`tempDir/**/dir2/**/*`), which evaluates all found directories in `tempDir`, but `**` also means nonexistent, so it tested `tempDir/dir2/**/*` as well. 

I've mainly silenced the warning by just checking if the path exists first, and then breaking out that rather large directory iteration block into it's own function.

Issue #631, opened 2019-08-21
Support sources glob patterns


PaulTaykalo:
## **Issue**
Currently, when we're adding sources, we're doing this in the way we get all files inside source and then exclude files, we don't want them to be there
```
 sources: 
      - path: "MyProject/Sources"
        createIntermediateGroups: true
        excludes:      
          - "*/*Tests.*"
```

In development often it is nice to have Tests files right near files being tested
```
SomeFile.swift
SomeFileTests.swift
```

This will allow Xcode to jump between related sources as well

While we can easily exclude Test files from the original target, we cannot create test target which will include all files passing glob pattern

## **Suggestion**

to have an option that will allow only to include files, passing glob pattern

```
 sources: 
      - path: "MyProject/Sources"
        createIntermediateGroups: true
        includes-only:      
          - "*/*Tests.*"
```

## **Impact on existing realisization**
- Since this is an optional value, no projects will be affected
- undefined behaviour if `include` and `exclude` will be defined at the same time

P.S. This could be potentially done by using negation of `excludes`, but it seems that `glob` that is used in xxodegen doesn't like negation match

yonaskolb:
#636 has been merged. Excited for your `includes` PR @bclymer 😄👍

PR #644, opened 2019-09-01
Add support for includes filter in target sources

bclymer:
# Purpose

Fixes https://github.com/yonaskolb/XcodeGen/issues/631.

This PR adds a the new property `includes` to `Sources` on `Target`. It's of type `[String]` and follows the same patterns and syntax that `excludes` does. In fact it's just the opposite of `excludes` in functionality.

If both `includes` and `excludes` match on a file, `excludes` will take priority. For example...

```
includes:
  - **/*.swift
excludes:
  - TestFile.swift
```

This will result in `TestFile.swift` not being included, but all other `.swift` files being included.

# Other Changes

I removed the instance level `targetSourceExcludePaths` var and replaced it with a local variable that gets passed around, along with `targetSourceIncludesPaths`. From what I could tell there was a single `SourceGenerator` used to generate the project, so if this project ever went multi-threaded (something I might look into eventually) having these target specific variables in a global context could have created issues.


Sources/XcodeGenKit/SourceGenerator.swift
```diff
@@ -320,14 +319,19 @@ class SourceGenerator {
     }
 
     /// Checks whether the path is not in any default or TargetSource excludes
-    func isIncludedPath(_ path: Path) -> Bool {
+    func isIncludedPath(_ path: Path, targetSourceExcludePaths: Set<Path>, targetSourceIncludePaths: Set<Path>) -> Bool {
```
yonaskolb: For readability could we rename all the uses of these params to just excludePaths and includePaths

Sources/XcodeGenKit/SourceGenerator.swift
```diff
@@ -400,7 +411,7 @@ class SourceGenerator {
 
         if let baseLocalisedDirectory = baseLocalisedDirectory {
             for filePath in try baseLocalisedDirectory.children()
-                .filter(isIncludedPath)
+                .filter({ self.isIncludedPath($0, targetSourceExcludePaths: targetSourceExcludePaths, targetSourceIncludePaths: targetSourceIncludePaths) })
```
yonaskolb: Nitpick: trailing closures

Sources/XcodeGenKit/SourceGenerator.swift
```diff
@@ -420,7 +431,7 @@ class SourceGenerator {
         for localisedDirectory in localisedDirectories {
             let localisationName = localisedDirectory.lastComponentWithoutExtension
             for filePath in try localisedDirectory.children()
-                .filter(isIncludedPath)
+                .filter({ self.isIncludedPath($0, targetSourceExcludePaths: targetSourceExcludePaths, targetSourceIncludePaths: targetSourceIncludePaths) })
```
yonaskolb: Nitpick: trailing closures

Sources/XcodeGenKit/SourceGenerator.swift
bclymer: I actually had to include the `()` because this is part of the `for in` statement. I just extracted the initialization of this sequence into its own line for clarity anyway.

PR #650, opened 2019-09-09
Allow dylib extension for SDK linkage

kateinoigakukun:
`invalidSDKDependency` says extension of SDK dependency must be one of `.framework`, `.dylib` or `.tbd`.
https://github.com/yonaskolb/XcodeGen/blob/6620187d8ab0f9eb2f524d328ac95856bacb06d0/Sources/ProjectSpec/SpecValidationError.swift#L36-L37

But in current implementation prevents to use `dylib` extension.


kateinoigakukun:
I found same PR https://github.com/yonaskolb/XcodeGen/pull/564

yonaskolb:
If you’d like to address my comments from that PR here, that would be great

PR #651, opened 2019-09-09
Expand template variable in Array of Any

kateinoigakukun:
For example, `sources` contains `String` and `JSONDictionary`, the array type is `[Any]`.
But it's not expanded even if it contains template variable


yonaskolb:
Thanks @kateinoigakukun! Could you please add a changelog entry?
If you had time to add a unit test as well that would be awesome

PR #655, opened 2019-09-14
Support External Target Reference

kateinoigakukun:
## Motivation

XcodeGen supports generating schemes but it only allows to reference it's own targets in project.yml.
We often want to execute unit tests over xcodeprojs at once.

## Changed

This pr added `externalProject` property which enables to build other project targets.

And added new syntax to reference other project target.

```yaml
projectReferences:
  FooLib:
    path: path/to/FooLib.xcodeproj
schemes:
  TestTarget:
    build:
      targets:
        FooLib/Target7: ["run"]
```


yonaskolb:
I'd like to move towards being able to reference other projects in more places in an easy way. This might be a good place to start to doing that. I think `target: String` could be turned into `target: TargetReference`
```swift
struct TargetReference {
  let name: String
  enum Location {
  	case local // better name? spec?
  	case project(path: String)
  }

  init(string: String) {
	..
  }	
}
```
(In the future this could include a case for named projects when the spec supports multiple projects.) We just need some syntax to describe it in a string so it can be defined and parsed. Something like `MyProject.xcodeproj/target`

kateinoigakukun:
@yonaskolb I implemented `ProjectName/TargetName` syntax to represent other project target.



```yaml
externalProjects:
  FooLib:
    path: Libraries/FooLib.xcodeproj
schemes:
  TestTarget:
    build:
      targets:
        FooLib/Target7: ["run"]
```

kateinoigakukun:
@yonaskolb Sorry for pinging you, cloud you review again?

Docs/ProjectSpec.md
```diff
@@ -50,6 +50,7 @@ You can also use environment variables in your configuration file, by using `${S
 - [ ] **targetTemplates**: **[String: [Target Template](#target-template)]** - a list of targets that can be used as templates for actual targets which reference them via a `template` property. They can be used to extract common target settings. Works great in combination with `include`.
 - [ ] **packages**: **[String: [Swift Package](#swift-package)]** - a map of Swift packages by name
 - [ ] **localPackages**: **[String]** - A list of paths to local Swift Packages. The paths must be directories with a `Package.swift` file in them. This is used to override `packages` with a local version for development purposes.
+- [ ] **externalProjects**: **[String: [External Project](#external-project)]** - a map of external projects by name
```
yonaskolb: What about naming this just `projects` of type `ProjectReference`. Thoughts? Or `projectReferences`

Sources/ProjectSpec/Scheme.swift
```diff
@@ -239,16 +244,60 @@ public struct Scheme: Equatable {
...
-        public init(target: String, buildTypes: [BuildType] = BuildType.all) {
+        public init(target: TargetReference, buildTypes: [BuildType] = BuildType.all) {
             self.target = target
             self.buildTypes = buildTypes
         }
     }
 }
 
+
+public struct TargetReference: Equatable, Hashable {
```
yonaskolb: Lets move this into it's own file

Sources/ProjectSpec/Scheme.swift
```diff
@@ -239,16 +244,60 @@ public struct Scheme: Equatable {
...
+        case 1:
+            location = .local
+            name = String(paths[0])
+        default: fatalError("unreachable")
+        }
+    }
+}
+
+extension TargetReference {
+    public func toString() -> String {
```
yonaskolb: Let's make this a `reference: String` property, and also conform to CustomStringConvertable that uses this in it's description

Sources/ProjectSpec/TargetScheme.swift
```diff
@@ -42,7 +42,7 @@ extension TargetScheme: JSONObjectConvertible {
         if let targets = jsonDictionary["testTargets"] as? [Any] {
             testTargets = try targets.compactMap { target in
                 if let string = target as? String {
-                    return .init(name: string)
+                    return .init(stringLiteral: string)
```
yonaskolb: Let's keep the old init around

Tests/XcodeGenKitTests/ProjectSpecTests.swift
```diff
@@ -207,7 +207,7 @@ class ProjectSpecTests: XCTestCase {
                 var project = baseProject
                 project.schemes = [Scheme(
                     name: "scheme1",
-                    build: .init(targets: [.init(target: "invalidTarget")]),
+                    build: .init(targets: [.init(target: .init(name: "invalidTarget", location: .local))]),
```
yonaskolb: Let's make `TargetReference` conform to StringLiteralInitializable or whatever it's called. Then all these string inits can stay the same

Tests/Fixtures/TestProject/Project.xcodeproj/xcshareddata/xcschemes/Framework.xcscheme
```diff
@@ -80,7 +80,7 @@
          <BuildableReference
             BuildableIdentifier = "primary"
             BlueprintIdentifier = "AE3F93DB94E7208F2F1D9A78"
-            BuildableName = "Framework.framework"
+            BuildableName = "Framework_iOS.framework"
```
yonaskolb: What made this change?

Tests/XcodeGenKitTests/SchemeGeneratorTests.swift
```diff
@@ -43,7 +43,7 @@ class SchemeGeneratorTests: XCTestCase {
     func testSchemes() {
         describe {
 
-            let buildTarget = Scheme.BuildTarget(target: app.name)
+            let buildTarget = Scheme.BuildTarget(target: .init(name: app.name, location: .local))
```
yonaskolb: Lets have a static helper func on `TargetReference`
```
static func local(_ name: String) -> TargetReference {
  return TargetReference(name: name, location: .local)
}
```
This this becomes `Scheme.BuildTarget(target: .local(app.name))`

Tests/XcodeGenKitTests/SpecLoadingTests.swift
```diff
@@ -797,7 +799,7 @@ class SpecLoadingTests: XCTestCase {
                     targets: [
                         "Target1",
                         Scheme.Test.TestTarget(
-                            name: "Target2",
+                            targetReference: .init(name: "Target2", location: .project("ExternalProject")),
```
yonaskolb: All this can use the string literal init `ExternalProject/Target2`

Tests/Fixtures/scheme_test/test_project.yml
```diff
@@ -0,0 +1,5 @@
+name: TestProject
```
yonaskolb: Where is this new Fixture used? Do we need it?

Docs/ProjectSpec.md
kateinoigakukun: `projectReferences` seems good! Thanks!

Tests/Fixtures/scheme_test/test_project.yml
kateinoigakukun: This project file is used to generate test project in https://github.com/yonaskolb/XcodeGen/pull/655/files#diff-c6ba86026b68e8323d7e329491a61492R264-R271

Tests/Fixtures/TestProject/Project.xcodeproj/xcshareddata/xcschemes/Framework.xcscheme
kateinoigakukun: Sorry, it's mistake. I reverted buildableName generation logic.

Sources/ProjectSpec/TargetReference.swift
```diff
@@ -0,0 +1,66 @@
...
+//  TargetReference.swift
+//  ProjectSpec
+//
+//  Created by Yuta Saito on 2019/10/15.
+//
+
+import Foundation
+import JSONUtilities
+
+public struct TargetReference: Equatable, Hashable {
```
giginet: should be only `Hashable`.
Because `Hashable` is inherited by `Equatable`

giginet:
```suggestion
public struct TargetReference: Hashable {
```

Sources/ProjectSpec/TargetReference.swift
```diff
@@ -0,0 +1,66 @@
...
+    public let name: String
+    public let location: Location
+
+    public enum Location: Equatable, Hashable {
```
giginet:
```suggestion
    public enum Location: Hashable {
```

Sources/ProjectSpec/TargetReference.swift
```diff
@@ -0,0 +1,66 @@
...
+            throw SpecParsingError.invalidTargetReference(string)
+        }
+        switch paths.count {
+        case 2:
+            location = .project(String(paths[0]))
+            name = String(paths[1])
+        case 1:
+            location = .local
+            name = String(paths[0])
+        default: fatalError("unreachable")
```
giginet: This is a throwable initializer, so you can throw errors instead of calling `fatalError`.

Sources/XcodeGenKit/SchemeGenerator.swift
```diff
@@ -83,34 +83,56 @@ public class SchemeGenerator {
...
+            switch buildTarget.target.location {
+            case .project(let project):
+                guard let projectReference = self.project.getProjectReference(project) else {
+                    fatalError("Unable to find project reference named \"\(project)\" in project.yml")
+                }
+                pbxProj = try XcodeProj(pathString: projectReference.path).pbxproj
+                projectFilePath = projectReference.path
+            case .local:
+                pbxProj = self.pbxProj
+                projectFilePath = "\(self.project.name).xcodeproj"
```
giginet: Is this true?

Can you try `project.basepath` to get xcodeproj path.

Sources/XcodeGenKit/SchemeGenerator.swift
```diff
@@ -83,34 +83,56 @@ public class SchemeGenerator {
...
+            let buildableName: String
+
+            switch buildTarget.target.location {
+            case .project:
+                buildableName = pbxTarget.productNameWithExtension() ?? pbxTarget.name
+            case .local:
+                guard let _buildableName =
+                        project.getTarget(buildTarget.target.name)?.filename ??
+                        project.getAggregateTarget(buildTarget.target.name)?.name else {
+                    fatalError("Unable to determinate \"buildableName\" for build target: \(buildTarget.target)")
```
giginet: should be throwable error

Sources/XcodeGenKit/SchemeGenerator.swift
kateinoigakukun: Yes, `ReferencedContainer` is relative to a directory which has `xcodeproj`

Sources/ProjectSpec/ProjectReference.swift
```diff
@@ -0,0 +1,34 @@
...
+//  ProjectSpec
+//
+//  Created by Yuta Saito on 2019/10/15.
+//
+
+import Foundation
+import JSONUtilities
+
+public struct ProjectReference {
+    public let name: String
```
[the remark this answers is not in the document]
yonaskolb: @giginet I actually disagree with this change. Mutability of structs can be locked off wherever that struct is referenced, it doesn't need to be immutable an instance variable level. Also mutating structs is not a potentally dangerous operation like it is with classes, as you're not mutating data out from under anyone. Because of this the rest of the project spec is all vars for easy editing.

Sources/XcodeGenKit/SchemeGenerator.swift
```diff
@@ -83,34 +83,56 @@ public class SchemeGenerator {
...
+            let buildableName: String
+
+            switch buildTarget.target.location {
+            case .project:
+                buildableName = pbxTarget.productNameWithExtension() ?? pbxTarget.name
+            case .local:
+                guard let _buildableName =
+                        project.getTarget(buildTarget.target.name)?.filename ??
+                        project.getAggregateTarget(buildTarget.target.name)?.name else {
+                    throw SchemeGenerationError.unresolvableBuildableName(buildTarget)
```
yonaskolb: While there's nothing bad with having specific errors thrown here, checks like this occur in the SpecValidation phase. Adding a new set of SchemeGenerationErrors just introduces more maintenance. The rest of the codebase gracefully fails at points like this. Thoughts @kateinoigakukun, @giginet?

Sources/ProjectSpec/ProjectReference.swift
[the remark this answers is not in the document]
giginet: I understood your opinion. anything will be fine.

@kateinoigakukun  Could you revert these changes 🙏

Sources/XcodeGenKit/SchemeGenerator.swift
giginet: You're right this condition will be not caused by user inputs, seems to be caused by an implementation error.
Other exceptions will be occurred by user inputs, however, here is not. 
It's better to call `fatalError` instead of throwing the exceptions here.

yonaskolb: Sorry @kateinoigakukun, could you revert to how you had it before then

Sources/XcodeGenKit/SchemeGenerator.swift
```diff
@@ -89,7 +89,7 @@ public class SchemeGenerator {
             switch buildTarget.target.location {
             case .project(let project):
                 guard let projectReference = self.project.getProjectReference(project) else {
-                    throw SchemeGenerationError.missingProject(project)
+                    fatalError("Unable to find project reference named \"\(project)\" in project.yml")
```
giginet: Sorry for confusing you. I think it's better L92 and L102 should be still throwing exceptions.
(I agreed to revert L114 only)

The reason why, these(L92 and L102) are user errors, which they would be caused by user inputs.
If users mistook the project names, these errors are raised in regular situations.

There is one more reason. `func generateScheme` is a public throwable function. function users expect to throw errors by wrong usages.
So I think you're better to throw exceptions instead of calling `fatalError`.

How do you think about my opinion? @kateinoigakukun @yonaskolb

Sources/ProjectSpec/SpecValidation.swift
```diff
@@ -180,8 +180,9 @@ extension Project {
 
         for scheme in schemes {
             for buildTarget in scheme.build.targets {
-                if getProjectTarget(buildTarget.target) == nil {
-                    errors.append(.invalidSchemeTarget(scheme: scheme.name, target: buildTarget.target))
+                guard buildTarget.target.location == .local else { continue }
```
yonaskolb: We should switch on location and also check that project reference is correct

Sources/ProjectSpec/Scheme.swift
```diff
@@ -320,7 +325,7 @@ extension Scheme.Test: JSONObjectConvertible {
         if let targets = jsonDictionary["targets"] as? [Any] {
             self.targets = try targets.compactMap { target in
                 if let string = target as? String {
-                    return TestTarget(name: string)
+                    return TestTarget(stringLiteral: string)
```
yonaskolb: Use the non string literal init

Sources/ProjectSpec/TargetReference.swift
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    public init(name: String, location: Location) {
+        self.name = name
+        self.location = location
+    }
+}
+
+extension TargetReference {
+    public init(string: String) throws {
```
yonaskolb: We can probably drop the `string` param label
```suggestion
    public init(_ string: String) throws {
```

Sources/XcodeGenKit/SchemeGenerator.swift
```diff
@@ -83,34 +83,56 @@ public class SchemeGenerator {
...
-                blueprintName: buildTarget.target
+                blueprintName: buildTarget.target.name
             )
             return XCScheme.BuildAction.Entry(buildableReference: buildableReference, buildFor: buildTarget.buildTypes)
         }
 
         let testTargets = scheme.test?.targets ?? []
         let testBuildTargets = testTargets.map {
-            Scheme.BuildTarget(target: $0.name, buildTypes: BuildType.testOnly)
+            Scheme.BuildTarget(target: TargetReference(name: $0.name, location: .local), buildTypes: BuildType.testOnly)
```
yonaskolb:
```suggestion
            Scheme.BuildTarget(target: .local($0.name), buildTypes: BuildType.testOnly)
```

Sources/XcodeGenKit/SchemeGenerator.swift
```diff
@@ -219,11 +241,26 @@ public class SchemeGenerator {
...
+        }
+    }
+}
+
 extension Scheme {
     public init(name: String, target: Target, targetScheme: TargetScheme, debugConfig: String, releaseConfig: String) {
         self.init(
             name: name,
-            build: .init(targets: [Scheme.BuildTarget(target: target.name)]),
+            build: .init(targets: [Scheme.BuildTarget(target: TargetReference(name: target.name, location: .local))]),
```
yonaskolb:
```suggestion
            build: .init(targets: [Scheme.BuildTarget(target: .local(target.name))]),
```

Sources/XcodeGenKit/SchemeGenerator.swift
```diff
@@ -83,34 +83,56 @@ public class SchemeGenerator {
...
-        func getBuildEntry(_ buildTarget: Scheme.BuildTarget) -> XCScheme.BuildAction.Entry {
+        func getBuildEntry(_ buildTarget: Scheme.BuildTarget) throws -> XCScheme.BuildAction.Entry {
+            let pbxProj: PBXProj
+            let projectFilePath: String
+            switch buildTarget.target.location {
+            case .project(let project):
+                guard let projectReference = self.project.getProjectReference(project) else {
+                    throw SchemeGenerationError.missingProject(project)
+                }
+                pbxProj = try XcodeProj(pathString: projectReference.path).pbxproj
```
yonaskolb: This could become a performance issue, especially if there are lots of target references. We could at least cache the projects in this class 
```swift
var projects: [String: XcodeProj]
```
As far as I understand you're only loading these projects to get the proper file extension for the target right? Does the buildable name require that? Cause if it doesn't we can just do away with loading the projects.

kateinoigakukun: > We could at least cache the projects in this class

Right, I added cache logic.

> As far as I understand you're only loading these projects to get the proper file extension for the target right? Does the buildable name require that? Cause if it doesn't we can just do away with loading the projects.

`PBXProj` is also used to get blueprint of target https://github.com/yonaskolb/XcodeGen/pull/655/files#diff-b0109f56886f826d76bacb6b15dbf86dR132

PR #657, opened 2019-09-16
Add `debugEnabled` option for run and test

kateinoigakukun:
When we use some libraries like [johnno1962/Fortify](https://github.com/johnno1962/Fortify), we need to disable debugger.
So I added the option which enables us to disable it explicitly.


yonaskolb:
I tested this by editing a scheme in Xcode and noticed that when deselecting `debug executable` that the `selectedLauncherIdentifier` also changed from `Xcode.DebuggerFoundation.Launcher.LLDB` to `Xcode.IDEFoundation.Launcher.PosixSpawn`

Docs/ProjectSpec.md
```diff
@@ -687,6 +687,7 @@ The different actions share some properties:
 - [ ] **disableMainThreadChecker**: **Bool** - `run` and `test` actions can define a boolean that indicates that this scheme should disable the Main Thread Checker. This defaults to false
 - [ ] **language**: **String** - `run` and `test` actions can define a language that is used for Application Language
 - [ ] **region**: **String** - `run` and `test` actions can define a language that is used for Application Region
+- [ ] **debugEnabled**: **Bool** - `run` and `test` actions can define a whether debugger should be used.
```
yonaskolb: Can you just add `This defaults to true` and then we can merge

kateinoigakukun: Updated!

sassiwalid:
Hello, but I think that debugEnabled not work, it should be checked and unchecked automatically based on debugEnabled property. I opened this issue https://github.com/yonaskolb/XcodeGen/issues/1000 and I really why when It doesn't work :(

PR #658, opened 2019-09-16
Performance Tuning

kateinoigakukun:
- https://github.com/yonaskolb/XcodeGen/commit/fa8c71569a0a34b8618734a535d9721f6fea78b8
  While investigating performance issue, I found equality checking for `PBXObject` is very slow.
  So I changed to check only `path` property.
- https://github.com/yonaskolb/XcodeGen/pull/658/commits/0dc4a8de417084b08f800d63effc134da835f80a
  `cachedGroup.children` is called for each loop but it needs some computations every access https://github.com/tuist/XcodeProj/blob/368fb87695e5dbc09c6035ffdb873d4831f5e861/Sources/XcodeProj/Objects/Files/PBXGroup.swift#L11-L18

| before| after|
|:-:|:-:|
|スクリーンショット 2019-09-16 15 39 22 | スクリーンショット 2019-09-21 20 05 55 |

This change make `getGroup` 4.6 times faster than now in my project.

### DIFF
--- a/Sources/XcodeGenKit/SourceGenerator.swift
+++ b/Sources/XcodeGenKit/SourceGenerator.swift
@@ -245,8 +245,8 @@ class SourceGenerator {
             var cachedGroupChildren = cachedGroup.children
             for child in children {
                 // only add the children that aren't already in the cachedGroup
-                // Check equality by path because XcodeProj.PBXObject.== is very slow.
-                if !cachedGroupChildren.contains(where: { $0.path == child.path }) {
+                // Check equality by path and sourceTree because XcodeProj.PBXObject.== is very slow.
+                if !cachedGroupChildren.contains(where: { $0.path == child.path && $0.sourceTree == child.sourceTree }) {
                     cachedGroupChildren.append(child)
                 }
             }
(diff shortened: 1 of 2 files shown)


bclymer:
Based on the surrounding code it doesn't seem like full equality needs to be checked, so this feels like a pretty free win. I'd like @yonaskolb to validate that assumption though.

@kateinoigakukun could you add a code comment explaining why we aren't using `contains(element)`, and then also add a reference to this in the `CHANGELOG.md`?

bclymer:
Also @kateinoigakukun just because I'm curious, I'd love if you could provide some before and after benchmarks.

kateinoigakukun:
Hmmm, why CI triggered for push? It seems jobs for push are not triggered for other PRs.

yonaskolb:
Yeah not sure. I re-ran it earlier to see if it would work. Might be a GitHub actions bug. We can merge without it. Later today I hope to review some of these PRs 👍

Sources/XcodeGenKit/SourceGenerator.swift
```diff
@@ -242,12 +242,15 @@ class SourceGenerator {
         let groupReference: PBXGroup
 
         if let cachedGroup = groupsByPath[path] {
+            var cachedGroupChildren = cachedGroup.children
             for child in children {
                 // only add the children that aren't already in the cachedGroup
-                if !cachedGroup.children.contains(child) {
-                    cachedGroup.children.append(child)
+                // Check equality by path because XcodeProj.PBXObject.== is very slow.
+                if !cachedGroupChildren.contains(where: { $0.path == child.path }) {
```
yonaskolb: This _should_ be safe to do. Perhaps we can check for `sourceTree` as well just in case the group has a different root.

PR #666, opened 2019-09-30
Specify platform

giginet:
I specify platform on `Package.swift` and update dependencies

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -69,7 +69,7 @@
         "repositoryURL": "https://github.com/tuist/xcodeproj.git",
         "state": {
           "branch": null,
-          "revision": "0f563e2d7d604499e7b57a28c78ff23d5c545acd",
+          "revision": "aefcbf59cea5b0831837ed2f385e6dfd80d82cc9",
           "version": "7.1.0"
         }
       },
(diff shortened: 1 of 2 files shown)


Package.swift
```diff
@@ -4,6 +4,7 @@ import PackageDescription
 
 let package = Package(
     name: "XcodeGen",
+    platforms: [.macOS(.v10_13)],
```
giginet: It should be 10.13. Because `.sortedKey` option is used on `FileCache` 
https://github.com/yonaskolb/XcodeGen/blob/master/Sources/XcodeGenKit/CacheFile.swift#L17

Package.resolved
```diff
@@ -69,7 +69,7 @@
         "repositoryURL": "https://github.com/tuist/xcodeproj.git",
         "state": {
           "branch": null,
-          "revision": "0f563e2d7d604499e7b57a28c78ff23d5c545acd",
+          "revision": "aefcbf59cea5b0831837ed2f385e6dfd80d82cc9",
```
giginet: This hash is updated automatically. This change cause no effects.
Because, they seem to be almost same 😄  https://github.com/tuist/XcodeProj/compare/0f563e2d7d604499e7b57a28c78ff23d5c545acd...aefcbf59cea5b0831837ed2f385e6dfd80d82cc9

PR #667, opened 2019-09-30
Update SwiftCLI

giginet:
## Motivation

In the current version of SwiftCLI raises a lot of warnings.

```console
/Users/giginet/.ghq/github.com/yonaskolb/XcodeGen/.build/checkouts/SwiftCLI/Sources/SwiftCLI/Validation.swift:40:5: warning: 'public' modifier is redundant for static method declared in a public extension
    public static func allowing(_ values: T..., message: String? = nil) -> Validation {
    ^~~~~~~

/Users/giginet/.ghq/github.com/yonaskolb/XcodeGen/.build/checkouts/SwiftCLI/Sources/SwiftCLI/Validation.swift:45:5: warning: 'public' modifier is redundant for static method declared in a public extension
    public static func rejecting(_ values: T..., message: String? = nil) -> Validation {
    ^~~~~~~

/Users/giginet/.ghq/github.com/yonaskolb/XcodeGen/.build/checkouts/SwiftCLI/Sources/SwiftCLI/Validation.swift:54:5: warning: 'public' modifier is redundant for static method declared in a public extension
    public static func greaterThan(_ value: T, message: String? = nil) -> Validation {
    ^~~~~~~

/Users/giginet/.ghq/github.com/yonaskolb/XcodeGen/.build/checkouts/SwiftCLI/Sources/SwiftCLI/Validation.swift:58:5: warning: 'public' modifier is redundant for static method declared in a public extension
    public static func lessThan(_ value: T, message: String? = nil) -> Validation {
    ^~~~~~~

/Users/giginet/.ghq/github.com/yonaskolb/XcodeGen/.build/checkouts/SwiftCLI/Sources/SwiftCLI/Validation.swift:62:5: warning: 'public' modifier is redundant for static method declared in a public extension
    public static func within(_ range: ClosedRange<T>, message: String? = nil) -> Validation {
    ^~~~~~~
```

## Description

Use latest SwiftCUI(5.3.2) to avoid warning.

SwiftCLI 5.3 took breaking changes.
https://github.com/jakeheis/SwiftCLI/compare/5.2.2...5.3.2
So I use newer API for the frontend.


Sources/XcodeGenCLI/CommandRouter.swift
```diff
@@ -1,20 +0,0 @@
-import Foundation
-import SwiftCLI
-
-class CommandRouter: Router {
```
giginet: `Router` seems to be removed from latest SwiftCLI. Using `ArgumentListManipulator` instead.

Sources/XcodeGenCLI/XcodeGenCLI.swift
```diff
@@ -3,6 +3,21 @@ import ProjectSpec
 import SwiftCLI
 
 public class XcodeGenCLI {
+    private class Manipulator: ArgumentListManipulator {
```
giginet: struct

giginet:
Ready for review ✨

Sources/XcodeGenCLI/XcodeGenCLI.swift
```diff
@@ -15,7 +26,9 @@ public class XcodeGenCLI {
             description: "Generates Xcode projects",
             commands: [generateCommand]
         )
-        cli.parser = Parser(router: CommandRouter(defaultCommand: generateCommand))
+        let manipulator = Manipulator(commandName: generateCommand.name)
+        cli.argumentListManipulators.insert(manipulator, at: 0)
+        cli.parser.routeBehavior = .searchWithFallback(generateCommand)
```
yonaskolb: Does this new routeBehaviour replace the need for the `Manipulator`? The purpose of that was to make `generate` the default command

Sources/XcodeGenCLI/XcodeGenCLI.swift
```diff
@@ -3,6 +3,17 @@ import ProjectSpec
 import SwiftCLI
 
 public class XcodeGenCLI {
+    private struct Manipulator: ArgumentListManipulator {
+        let commandName: String
```
yonaskolb: Can we name this `defaultCommand`

Package.swift
```diff
@@ -16,7 +16,7 @@ let package = Package(
         .package(url: "https://github.com/kylef/Spectre.git", from: "0.9.0"),
         .package(url: "https://github.com/onevcat/Rainbow.git", from: "3.0.0"),
         .package(url: "https://github.com/tuist/xcodeproj.git", .exact("7.1.0")),
-        .package(url: "https://github.com/jakeheis/SwiftCLI.git", .exact("5.2.2")),
+        .package(url: "https://github.com/jakeheis/SwiftCLI.git", .exact("5.3.2")),
```
yonaskolb: We can make this `from`, incase other packages using XcodeGen also use SwiftCLI

Sources/XcodeGenCLI/XcodeGenCLI.swift
yonaskolb: Looks like it works fine with just the routeBehaviour

Package.swift
giginet: I prefer to use `upToNextMinor(from:)` than `from`. Because SwiftCLI brought breaking change when minor versions are updated.

Sources/XcodeGenCLI/XcodeGenCLI.swift
giginet: I tried getting rid of any custom manipulators. However, behavior is changed. So it is needed.

```console
$ swift run xcodegen # works fine
$ swift run xcodegen generate # behaviors are changed.

Usage: xcodegen [options]

Generate an Xcode project from a spec

Options:
  --cache-path <value>     Where the cache file will be loaded from and save to. Defaults to ~/.xcodegen/cache/{SPEC_PATH_HASH}
  -c, --use-cache          Use a cache for the xcodegen spec. This will prevent unnecessarily generating the project if nothing has changed
  -h, --help               Show help information
  -p, --project <value>    The path to the directory where the project should be generated. Defaults to the directory the spec is in. The filename is defined in the project spec
  -q, --quiet              Suppress all informational and success output
  -s, --spec <value>       The path to the project spec file. Defaults to project.yml

Error: command requires exactly 0 arguments
```

giginet: I tried again, it will works as you said. Manipulators are not needed.

PR #669, opened 2019-10-02
Add .rcproject path to sources build phase

Qusic:
.rcproject directories are projects of Reality Composer app bundled in Xcode 11 or later.


yonaskolb:
Thanks @Qusic! Could you add a changelog entry as well please?

PR #670, opened 2019-10-02
Fix tests in case-sensitive file system

Qusic:
Replaced `IMessage` with `iMessage` in test project files (to be consistent with `project.yml`).
Also use some different names to enforce renaming on case-insensitive file systems.


keith:
Moving these files will break this repo in case insensitive file systems. Doing a git pull won't work.

Qusic:
I agree. But when I cloned this repo, the tests did fail to run. Do you have any suggestions to fix that?

Qusic:
Updated cases in `project.yml` this time. It works for me and I think it shouldn't break tests elsewhere.

PR #671, opened 2019-10-02
Stabilize sorting of groups with duplicate names/paths

ChristopherRogers:
For example, previously a group with (name: nil, path: "Sources") would be considered equal to (name: "Sources", path: "../Sources"), even though they are distinct groups.

I think this was causing churn in the UUIDs for the files contained in such groups, but I'm not 100% sure.

### DIFF
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -185,7 +185,7 @@ public class PBXProjGenerator {
             for (platform, files) in carthageFrameworksByPlatform {
                 let platformGroup: PBXGroup = addObject(
                     PBXGroup(
-                        children: files.sorted(),
+                        children: files.sorted(by: PBXFileElement.areNamePathInIncreasingOrder(_:_:)),
                         sourceTree: .group,
                         path: platform
                     )
@@ -230,7 +230,7 @@ public class PBXProjGenerator {
         // add derived groups at the end
         derivedGroups.forEach(sortGroups)
         mainGroup.children += derivedGroups
-            .sorted { $0.localizedStandardCompare($1) == .orderedAscending }
+            .sorted { $0.namePathLocalizedStandardCompare($1) == .orderedAscending }
             .map { $0 }
 
         let projectAttributes: [String: Any] = ["LastUpgradeCheck": project.xcodeVersion]
@@ -416,7 +416,7 @@ public class PBXProjGenerator {
                     return sortOrder1 < sortOrder2
                 } else {
                     if (child1.name, child1.path) != (child2.name, child2.path) {
-                        return child1.localizedStandardCompare(child2) == .orderedAscending
+                        return child1.namePathLocalizedStandardCompare(child2) == .orderedAscending
                     } else {
                         return child1.context ?? "" < child2.context ?? ""
                     }
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -185,7 +185,7 @@ public class PBXProjGenerator {
             for (platform, files) in carthageFrameworksByPlatform {
                 let platformGroup: PBXGroup = addObject(
                     PBXGroup(
-                        children: files.sorted { $0.nameOrPath < $1.nameOrPath },
+                        children: files.sorted(),
                         sourceTree: .group,
                         path: platform
                     )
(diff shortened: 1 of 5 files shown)


Sources/XcodeGenKit/XCProjExtensions.swift
```diff
@@ -4,6 +4,59 @@ import XcodeProj
 
 extension PBXFileElement {
 
+    public struct NamePathPair: Comparable {
```
giginet: Should this need to be public?

Move this struct into `PBXProjectGenerator` and you can change to `fileprivate`.

Sources/XcodeGenKit/XCProjExtensions.swift
```diff
@@ -4,6 +4,59 @@ import XcodeProj
...
+            case (_?, nil):
+                return true
+            case (nil, nil):
+                return nil
+            }
+        }
+    }
+
+    public var namePathPair: NamePathPair {
+        return NamePathPair(name: name, path: path)
```
yonaskolb: I think this could be simplified by having a simple `sortString` on `PBXFileElement`, which removes the need for the `NamePathPair`

```swift
var sortString: String { return "\(name ?? "")\(path ?? "")" }
```

ChristopherRogers: You could possibly have two pairs that have the same name/path pair, like `(name: "pa", path: "th"), and `(name: nil, path: "path")`. If you want to take this approach (which is a lot simpler), you'd have to introduce a character between the two that cannot be used in either name or path. I guess \t would suffice. I don't think you can use it in filenames (although I could be wrong). I'll modify the code to do that if you don't have any problems with that.

giginet: Do you use tuples instead of the struct?

```swift
typealias Pair = (String, String)
func compare(_ lhs: Pair, _ rhs: Pair) -> ComparisonResult
```

tuples can't conform to any protocols, so you can use function to compare instead.

yonaskolb: I'm happy with using a simple search string. As you say @ChristopherRogers, you can add a random separator. Here's its `-`
```
var sortString: String { return "\(name ?? "")-\(path ?? "")" }
```

ChristopherRogers: You could create a duplicate with that, though. For example, ("a-b", "c") and ("a", "b-c") would both result in the sort string, "a-b-c".

yonaskolb: \t is fine then

yonaskolb: Or even
```
var sortString: String { return "name: \(name ?? ""), path: \(path ?? "")" }
```
It doesn’t really matter, it’s an implementation detail. Would be good to make it file private

ChristopherRogers:
Sorry for taking so long to getting back to this PR. I added tests and am just using a generated, unique string to sort as we discussed here. Let me know if you have any other concerns/feedback.

ChristopherRogers:
Looks like the order in the generated fixtures have changed. I'll take a look.

ChristopherRogers:
I’ve taken care of that and added another test case for it.

ChristopherRogers:
I’ve added the changelog entry and made a small change as well.

Sources/XcodeGenKit/XCProjExtensions.swift
```diff
@@ -3,9 +3,21 @@ import PathKit
...
     public var nameOrPath: String {
-        name ?? path ?? ""
+        return name ?? path ?? ""
+    }
+
+    func namePathLocalizedStandardCompare(_ rhs: PBXFileElement) -> ComparisonResult {
+        return sortString.localizedStandardCompare(rhs.sortString)
+    }
+
+    static func areNamePathInIncreasingOrder(_ lhs: PBXFileElement, _ rhs: PBXFileElement) -> Bool {
```
yonaskolb: I don't think we need this function

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -185,7 +185,7 @@ public class PBXProjGenerator {
             for (platform, files) in carthageFrameworksByPlatform {
                 let platformGroup: PBXGroup = addObject(
                     PBXGroup(
-                        children: files.sorted { $0.nameOrPath < $1.nameOrPath },
+                        children: files.sorted(by: PBXFileElement.areNamePathInIncreasingOrder(_:_:)),
```
yonaskolb: I don't think we actually don't need to sort this at all as the files are sorted later
```suggestion
                        children: Array(files),
```

yonaskolb:
Thanks @ChristopherRogers! Will make changes in master 👍

Issue #643, opened 2019-09-01
Support Scheme Templates


nevillco:
As mentioned in the project spec, `targetScheme` is a convenience API that has a subset of the functionality of declaring top-level `scheme`s. I'm a user who requires the extra API surface of the top level `scheme`, but now, because they are no longer part of the target, I have no way of reusing common code between schemes like I had with `targetTemplates`. It would be nice to be able to define reusable schemes in that way.

yonaskolb:
Yeah, this is a great idea @nevillco! Happy to accept a PR for this. I can help you out along the way

bclymer:
PR that has this working. Needs review / docs / cleanup yet.

bclymer:
Closing as this implemented here - https://github.com/yonaskolb/XcodeGen/pull/672

PR #672, opened 2019-10-02
Scheme Templates

bclymer:
Addresses https://github.com/yonaskolb/XcodeGen/issues/643

This adds support for `schemeTemplates` at the project level. It functions identically to `targetTemplates` (in fact it's been refactored to use the exact same code paths). It also sets up the project to support the `template` structure for any property. Not sure what else would be useful besides `target` and `scheme`, but it's there.

**Remaining TODOs**
- [x] Allow `templateAttributes` to impact target names.
- [x] Update Docs
- [x] Modify CHANGELOG.md

Example Usage:

This is in some `base_project.yml` file that you include in various places.
```yml
schemeTemplates:
  FeatureModuleSchemeMin:
    templates:
      - TestScheme
    build:
      targets:
       ${moduleName}: build

  TestScheme:
    test:
      gatherCoverageData: true
      targets:
        - name: ${moduleName}Tests
          parallelizable: true
          randomExecutionOrder: true
```
Then within your normal `project.yml`
```yml
schemes:
  MyModule-ExampleApp:
    templates:
      - FeatureModuleSchemeMin
    templateAttributes:
      moduleName: MyModule
```

### DIFF
--- a/Docs/ProjectSpec.md
+++ b/Docs/ProjectSpec.md
@@ -23,6 +23,7 @@
 - [Aggregate Target](#aggregate-target)
 - [Target Template](#target-template)
 - [Scheme](#scheme)
+  - [Scheme Template](#scheme-template)
 - [Swift Package](#swift-package)
 
 ## General
@@ -780,6 +781,39 @@ schemes:
       customArchiveName: MyTarget
       revealArchiveInOrganizer: false
 ```
+
+### Scheme Template
+
+This is a template that can be referenced from a normal scheme using the `templates` property. The properties of this template are the same as a [Scheme](#scheme). This functions identically in practice to [Target Template](#target-template).
+Any instances of `${scheme_name}` within each template will be replaced by the final scheme name which references the template.
+Any attributes defined within a targets `templateAttributes` will be used to replace any attribute references in the template using the syntax `${attribute_name}`.
+
+```yaml
+schemes:
+  MyModule:
+    templates:
+      - FeatureModuleScheme
+    templateAttributes:
+      testTargetName: MyModuleTests
+
+schemeTemplates:
+  FeatureModuleScheme:
+    templates:
+      - TestScheme
+    build:
+      targets:
+       ${scheme_name}: build
+
+  TestScheme:
+    test:
+      gatherCoverageData: true
+      targets:
+        - name: ${testTargetName}
+          parallelizable: true
+          randomExecutionOrder: true
+```
+
+The result will be a scheme that builds `MyModule` when you request a build, and will test against `MyModuleTests` when you request to run tests. This is particularly useful when you work in a very modular application and each module has a similar structure.
 
 ## Swift Package
 Swift packages are defined at a project level, and then linked to individual targets via a [Dependency](#dependency).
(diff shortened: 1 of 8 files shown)


bclymer:
@yonaskolb I haven't updated the docs yet because I'm not sure I consider the PR completely finished. Do you think I should integrated support with `targetScheme`? How do you think that would look with templates?

Sources/ProjectSpec/SpecFile.swift
```diff
@@ -158,7 +158,9 @@ extension Dictionary where Key == String, Value: Any {
     func replaceString(_ template: String, with replacement: String) -> JSONDictionary {
         var replaced: JSONDictionary = self
         for (key, value) in self {
-            replaced[key] = replace(value: value, template, with: replacement)
+            let newKey = key.replacingOccurrences(of: template, with: replacement)
+            replaced.removeValue(forKey: key)
+            replaced[newKey] = replace(value: value, template, with: replacement)
```
bclymer: This change worries me. Here's the scenario I'm looking to support with it:

```yml
schemeTemplates:
  StandardScheme:
    build:
      targets:
       ${moduleName}: build
    test:
      gatherCoverageData: true
      targets:
        - name:${moduleName}Tests
          parallelizable: true
          randomExecutionOrder: true

schemes:
  MyApp:
    templates:
      - StandardScheme
    templateAttributes:
      moduleName: MyApp
```
Since `schemeTemplates.StandardScheme.build.targets` has a key that needs to be configurable, the current implementation of `replace` didn't work. However, I think it will be extremely common when using scheme templates to want to be able to dynamically generate the name of the target you're building and testing against.

yonaskolb: Should be ok. Though, perhaps we can only remove and set the new key if it has changed (`if newKey != key`)

Sources/ProjectSpec/Target.swift
```diff
@@ -105,6 +105,12 @@ public struct Target: ProjectTarget {
     }
 }
 
+struct TemplateStructure {
+    let baseKey: String
+    let templatesKey: String
+    let nameToReplace: String
+}
```
yonaskolb: Let's move this to a `Template.swift`

Sources/ProjectSpec/Target.swift
```diff
@@ -236,6 +190,72 @@ extension Target {
...
+    }
+
+    static func resolveSchemeTemplates(jsonDictionary: JSONDictionary) -> JSONDictionary {
+        return resolveTemplates(jsonDictionary: jsonDictionary,
+                                templateStructure: TemplateStructure(baseKey: "schemes",
+                                                                     templatesKey: "schemeTemplates",
+                                                                     nameToReplace: "scheme_name"))
+    }
+
+    private static func resolveTemplates(jsonDictionary: JSONDictionary, templateStructure: TemplateStructure) -> JSONDictionary {
```
yonaskolb: Let's move all these template functions to `Template.swift`

Sources/ProjectSpec/SpecFile.swift
bclymer: Good idea, I added that.

bclymer:
@yonaskolb CR comments address and docs / changelog added.

Docs/ProjectSpec.md
```diff
@@ -781,6 +782,39 @@ schemes:
       revealArchiveInOrganizer: false
 ```
 
+### Scheme Template
+
+This is a template that can be referenced from a normal scheme using the `templates` property. The properties of this template are the same as a [Scheme](#scheme). This functions identically in practice to [Target Template](#target-template).
+Any instances of `${scheme_name}` within each template will be replaced by the final scheme name which references the template.
+Any attributes defined within a targets `templateAttributes` will be used to replace any attribute references in the template using the syntax `${attribute_name}`.
```
yonaskolb: Still referencing `target`

PR #673, opened 2019-10-02
Select suitable configs for auto-generated schemes

giginet:
# Motivation and Context

In a situation, there is the following project definition.

```yaml
name: MyApp
configs:
  Beta: debug
  Debug: debug
  Production: release
  Release: release
targets:
  MyFramework:
    type: framework
    platform: iOS
    sources: MyFramework
    scheme:
      testTargets:
        - MyFrameworkTests
  MyFrameworkTests:
    type: bundle.unit-test
    platform: iOS
    sources: MyFrameworkTests
```

Auto-generated schemes select `Beta` as debug config and `Production` as release config.
Because there are prior ones by alphabetical order.

Screen Shot 2019-10-03 at 0 19 24

However, we'll expect they select `Debug` and `Release` configuration.
This behavior seems to be ambigious.

# Description

This PR make to use `Debug` and `Release` piror to first ones.

And I added test cases for `SchemeGenerator`.

|Before|After|
|------|------|
|Screen Shot 2019-10-03 at 0 19 24|Screen Shot 2019-10-03 at 0 19 59|

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 #### Fixed
 - Fixed macOS unit test setting preset [#665](https://github.com/yonaskolb/XcodeGen/pull/665) @yonaskolb
 - Add `rcproject` files to sources build phase instead of resources [#669](https://github.com/yonaskolb/XcodeGen/pull/669) @Qusic
+- Fixed configuration selection behaviors for auto-generated schemes [#673](https://github.com/yonaskolb/XcodeGen/pull/673) @giginet
 
 ## 2.8.0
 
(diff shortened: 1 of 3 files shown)


yonaskolb:
Thanks @giginet!

philippelatulippe:
Hi! Is there a way to disable this change, or to explicitly state which build configuration I want to use for a target's generated change? My "Release" targets used the build configuration "AdHoc", but with this change they suddenly started using the wrong build configuration. 

I understand that it's just by luck of alphabetical sorting that it worked in the past, but now I can't find a way to set the build configuration of the convenience target scheme.

giginet:
@philippelatulippe Hi ✋ 

I think you're better to specify configurations expressly on project scheme section.
https://github.com/yonaskolb/XcodeGen/blob/master/Docs/ProjectSpec.md#scheme

```yaml
schemes:
  Release:
    build:
      targets:
        Release: all
    archive:
      config: AdHoc
```

giginet:
Your previous usage may be a bit fragile.

Issue #260, opened 2018-03-01
Recursive path issues


iamtomcat:
When I move to the parent directory and then try and use other folder outside of the current directory, everything is picked up fine, however all the folders are not included inside of the named directory or the last folder.

I'm getting around this by using a symlink to pull the folders in correctly.

Should also note this happens when using a full path as well.

Using a symlink:
<img width="272" alt="screen shot 2018-03-01 at 3 21 21 pm" src="https://user-images.githubusercontent.com/6033483/36875334-62183080-1d64-11e8-89c4-f8b23e77d8e1.png">

Without symlink and using a parent path: `../otherfolder/sources`
<img width="266" alt="screen shot 2018-03-01 at 3 22 27 pm" src="https://user-images.githubusercontent.com/6033483/36875345-6e86bee0-1d64-11e8-832c-5edb4a1bd305.png">

yonaskolb:
Hi @iamtomcat 
Do you have an example spec, showing how you use source paths.
Are you using `fileGroups` or `options. createIntermediateGroups`?

iamtomcat:
Here is my spec. No I'm not using filegroups. Maybe that is what I'm missing in this case?
```
name: Generated
options:
  deploymentTarget: 
    iOS: "8.0"
  bundleIdPrefix: com.paywith
  usesTabs: false
  indentWidth: 4
  tabWidth: 4
configs:
  Debug: debug
  Release: release
targets:
  PayWithSDKCoreFramework:
    type: framework
    platform: iOS
    settings:
      INFOPLIST_FILE: /Users/tclark/projects/PayWithSDKCore/PayWithSDKCoreFramework/Info.plist
    sources: 
      - path: /Users/tclark/projects/PayWithSDKCore/Sources
        name: CheeseTown
      - path: ../PayWithSDKCore/PayWithSDKCore
  PayWithSDK:
    type: library.static
    platform: iOS
  PayWithSDKFramework:
    type: framework
    platform: iOS
  PayWithSDKMallRewards:
    type: framework
    platform: iOS
    #com.paywith.PayWithSDKMallRewards
    settings:
      INFOPLIST_FILE: "PayWithSDKMallRewards/Info.plist"
      GCC_PREPROCESSOR_DEFINITIONS: ["$(inherited)", "IS_NEW=1"]
    dependencies:
      - carthage: AirshipKit
      - target: PayWithSDKCoreFramework
    sources: 
      - Sources/Shared
[...]

PR #681, opened 2019-10-08
Resolving relative paths with custom project destination

giginet:
# Motivation and Context

In the situation generating with `--project` option different with spec's directory, All containing files are missing on the generated projects.

```console
$ xcodegen -p ./GeneratedProjects
```

Screen Shot 2019-10-08 at 21 19 41

There is following directory structure.

```
.
├── GeneratedProjects
│   └── MyFramework.xcodeproj
├── MyFramework
│   ├── SubGroup
│   │   └── A.swift
│   └── main.swift
└── project.yml
```

# Description

This PR solve this issue. Top Level group's location must be the relative path from the generated project.

|Before|After|
|---------|-----------|
|Screen Shot 2019-10-08 at 21 19 41|Screen Shot 2019-10-08 at 21 20 08|

Before : `MyFramework` group's location was `MyFramework`. This value comes from a relative path **from the project spec**. It should be a relative path from `PROJECT_ROOT`.

After: `MyFramework` group's location chages to `../MyFramework`. It's a relative path **from `PROJECT_ROOT`**.

### DIFF
--- a/Sources/XcodeGenKit/ProjectGenerator.swift
+++ b/Sources/XcodeGenKit/ProjectGenerator.swift
@@ -13,11 +13,11 @@ public class ProjectGenerator {
         self.project = project
     }
 
-    public func generateXcodeProject(to projectDestinationDirectory: Path? = nil) throws -> XcodeProj {
+    public func generateXcodeProject(to projectDirectory: Path? = nil) throws -> XcodeProj {
 
         // generate PBXProj
         let pbxProjGenerator = PBXProjGenerator(project: project,
-                                                projectDestinationDirectory: projectDestinationDirectory)
+                                                projectDirectory: projectDirectory)
         let pbxProj = try pbxProjGenerator.generate()
 
         // generate Schemes
--- a/Sources/XcodeGenKit/ProjectGenerator.swift
+++ b/Sources/XcodeGenKit/ProjectGenerator.swift
@@ -13,7 +13,7 @@ public class ProjectGenerator {
         self.project = project
     }
 
-    public func generateXcodeProject(to projectDirectory: Path? = nil) throws -> XcodeProj {
+    public func generateXcodeProject(in projectDirectory: Path? = nil) throws -> XcodeProj {
 
         // generate PBXProj
         let pbxProjGenerator = PBXProjGenerator(project: project,
--- a/Sources/XcodeGenKit/ProjectGenerator.swift
+++ b/Sources/XcodeGenKit/ProjectGenerator.swift
@@ -13,10 +13,11 @@ public class ProjectGenerator {
         self.project = project
     }
 
-    public func generateXcodeProject() throws -> XcodeProj {
+    public func generateXcodeProject(to projectDestinationDirectory: Path? = nil) throws -> XcodeProj {
 
         // generate PBXProj
-        let pbxProjGenerator = PBXProjGenerator(project: project)
+        let pbxProjGenerator = PBXProjGenerator(project: project,
+                                                projectDestinationDirectory: projectDestinationDirectory)
         let pbxProj = try pbxProjGenerator.generate()
 
         // generate Schemes
--- a/Sources/XcodeGenKit/SourceGenerator.swift
+++ b/Sources/XcodeGenKit/SourceGenerator.swift
@@ -13,6 +13,7 @@ struct SourceFile {
 class SourceGenerator {
 
     var rootGroups: Set<PBXFileElement> = []
+    private let projectDestinationDirectory: Path?
     private var fileReferencesByPath: [String: PBXFileElement] = [:]
     private var groupsByPath: [Path: PBXGroup] = [:]
     private var variantGroupsByPath: [Path: PBXVariantGroup] = [:]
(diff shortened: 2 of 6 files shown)


Sources/XcodeGenKit/SourceGenerator.swift
```diff
@@ -286,9 +288,16 @@ class SourceGenerator {
             let isTopLevelGroup = (isBaseGroup && !createIntermediateGroups) || isRootPath
 
             let groupName = name ?? path.lastComponent
-            let groupPath = isTopLevelGroup ?
-                ((try? path.relativePath(from: project.basePath)) ?? path).string :
```
giginet: The previous implementation seems to be wrong.

According to the `PBXGroup`'s documentation, `path` argument must be relative path from source root.
https://github.com/tuist/XcodeProj/blob/aefcbf59cea5b0831837ed2f385e6dfd80d82cc9/Sources/XcodeProj/Objects/Files/PBXGroup.swift#L28

However, in some situation, `groupPath` become `path.string`. It is absolute path.

Tests/XcodeGenKitTests/ProjectGeneratorTests.swift
```diff
@@ -1004,4 +1004,38 @@ class ProjectGeneratorTests: XCTestCase {
...
+        )
+        
+        describe("generateXcodeProject") {
+            $0.context("without projectDestinationDirectory") {
+                $0.it("generate groups") {
+                    let project = Project(name: "test", targets: [frameworkWithSources])
+                    let generator = ProjectGenerator(project: project)
+                    let generatedProject = try generator.generateXcodeProject()
+                    let group = generatedProject.pbxproj.groups.first(where: { $0.nameOrPath == groupName })
+                    try expect(group?.path) == "App_iOS"
```
giginet: This value should be relative path from source root.
https://github.com/yonaskolb/XcodeGen/pull/681/files#r332485507

Tests/XcodeGenKitTests/ProjectGeneratorTests.swift
```diff
@@ -1004,4 +1004,38 @@ class ProjectGeneratorTests: XCTestCase {
...
+            }
+            
+            $0.context("with projectDestinationDirectory") {
+                $0.it("generate groups") {
+                    let destinationPath = fixturePath
+                    let project = Project(name: "test", targets: [frameworkWithSources])
+                    let generator = ProjectGenerator(project: project)
+                    let generatedProject = try generator.generateXcodeProject(to: destinationPath)
+                    let group = generatedProject.pbxproj.groups.first(where: { $0.nameOrPath == groupName })
+                    try expect(group?.path) == "TestProject/App_iOS"
```
giginet: Test fixing this issue.

Sources/XcodeGenKit/SourceGenerator.swift
```diff
@@ -286,9 +288,13 @@ class SourceGenerator {
...
-                path.lastComponent
+            let groupPath: String
+            switch try? path.relativePath(from: projectDirectory ?? project.basePath).string {
+            case .some(let relativePath) where isTopLevelGroup:
+                groupPath = relativePath
+            default:
+                groupPath = path.lastComponent
```
yonaskolb: Could this be simplified to use an `if let`
```swift
if isTopLevelGroup, let relativePath = try? path.relativePath(from: projectDirectory ?? project.basePath).string {
    groupPath = relativePath
} else {
    groupPath = path.lastComponent
}
```

yonaskolb: There might be other places using `project.basePath` in the file. Perhaps we can have an instance property called `projectBasePath` that gets set in the initializer, which everything can read

giginet: @yonaskolb I introduce `resolveGroupPath` instead. It may help us to get a good group path from projectBasePath relatively.

https://github.com/yonaskolb/XcodeGen/pull/681/commits/e63fe6be6f795109fd8de63911242c3b04211061

Sources/XcodeGenKit/ProjectGenerator.swift
```diff
@@ -13,10 +13,11 @@ public class ProjectGenerator {
         self.project = project
     }
 
-    public func generateXcodeProject() throws -> XcodeProj {
+    public func generateXcodeProject(to projectDirectory: Path? = nil) throws -> XcodeProj {
```
kateinoigakukun: Is this default argument necessary? It seems there is no use of the default value.

yonaskolb: Keep in mind that XcodeGenKit is used as a library in other third party tools

kateinoigakukun: Oh I overlooked `public` modifier.

yonaskolb: minor nitpick if using naming like that:
```suggestion
    public func generateXcodeProject(in projectDirectory: Path? = nil) throws -> XcodeProj {
```

yonaskolb:
Looks good apart from a minor param name comment. Would be good to to know if this fixes this issue https://github.com/yonaskolb/XcodeGen/issues/260

giginet:
@yonaskolb Merging is blocked because of minor changes after approval....
https://github.com/yonaskolb/XcodeGen/pull/681/commits/80808d69c8ff9d3d2e970f93ed83605c090cc9d1

Could you approve this again?

PR #685, opened 2019-10-14
Add base localisation by default

### LINKED ISSUE #680: Enable Base localization
image

how can I fix this warning? Tried to set a

```
knownRegions: ["en", "de", "Base"]
```

element on top level, but didn't help...

yonaskolb:
This adds a base localisation even if there are no Base localized files
Fixes #680

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -69,7 +69,7 @@
         "repositoryURL": "https://github.com/tuist/xcodeproj.git",
         "state": {
           "branch": null,
-          "revision": "0f563e2d7d604499e7b57a28c78ff23d5c545acd",
+          "revision": "aefcbf59cea5b0831837ed2f385e6dfd80d82cc9",
           "version": "7.1.0"
         }
       },


bryansum:
👋 After patching this version in I'm still getting the same warning. I made a change here: https://github.com/bryansum/XcodeGen/commit/f0d8eaab862b323e995fc5d63d22bd654cc510f8 which fixes it for me, although I'm not sure whether this is the right solution.

yonaskolb:
If your fix works in your case, then I would think the warning is legitimate and would show up in a regular project. Xcode 11 wants you to have things localized to base

bryansum:
Makes sense. In my case I have an `en.lproj` (and other languages) but no `Base.lproj`, which leads to a derived `knownRegions` value of only `en` (thus the warning). I suppose I could move those strings to `Base` but it wasn't clear to whether that's the "correct" solution. Thanks!

yonaskolb:
Does the warning offer a fixit? See what it does

bryansum:
yeah it adds "Base" to `knownRegions` as the fix-it.

yonaskolb:
It doesn’t move your en files to base or copy them? If that’s all it does, then maybe we can indeed apply your fix

bryansum:
No, it doesn't move them or copy them, it just adds that small change to add `Base` to `knownRegions` in the `pbxproj` file. Added a PR: https://github.com/yonaskolb/XcodeGen/pull/694 in case.

PR #686, opened 2019-10-15
Force checking SwiftFormat before merging

giginet:
# Motivation and Context

This project integrates SwiftFormat, However, some sources are invalid.

# Description

- Applied changes for all sources.
- I added CI step to lint projects. This setting helps us avoiding to commit a source with wrong formats.


giginet:
I merged the current master.

@yonaskolb I think it's better to check `swiftformat . --lint` on every CIs.
How do you think about?

yonaskolb:
I’m wary of applying checks on CI with SwiftLint or SwiftFormat without locking the versions of those down. I’d also rather have a system that post warnings in the PR rather than simply failing CI

giginet:
We have to do followings before:

- Pin SwiftFormat version (maybe using Mint)
- Notify all warnings to authors (using danger)
- Notify how to execute formatter (on issue templates or CONTRIBUTE.md)

I'll close PR for now. Thanks.

PR #688, opened 2019-10-20
Support Static Frameworks for Carthage dependencies

giginet:
## Motivation and Context

Carthage officially supports Static Frameworks.
https://github.com/Carthage/Carthage#build-static-frameworks-to-speed-up-your-apps-launch-times

These frameworks are built to different directory with dynamic ones. So we can't integrate static frameworks automatically by `carthage` dependency.

## Description

This PR makes it enable to integrate Carthage static framework easily.

```yaml
targets:
  MyTarget:
    dependencies:
      - carthage: AStaticFramework
        static: true
```

Screen Shot 2019-10-20 at 15 02 10

With enabling this option, XcodeGen behaves like following:

- Finding search path should be`Carthage/Build/${Platform}/Static/${FrameworkName}.framework`
- Static frameworks aren't added to the `copy-framework` build phase


Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -647,17 +657,17 @@ public class PBXProjGenerator {
...
-            var platformPath = Path(carthageResolver.buildPath(for: target.platform))
+            var platformPath = Path(carthageResolver.buildPath(for: target.platform, isStatic: isStaticDependency(for: dependency)))
             var frameworkPath = platformPath + dependency.reference
             if frameworkPath.extension == nil {
                 frameworkPath = Path(frameworkPath.string + ".framework")
             }
             let fileReference = sourceGenerator.getFileReference(path: frameworkPath, inPath: platformPath)
 
-            if embed {
+            if embed && !isStaticDependency(for: dependency) {
```
giginet: Static frameworks should not be embeded.

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -917,8 +932,16 @@ public class PBXProjGenerator {
             // set Carthage search paths
             let configFrameworkBuildPaths: [String]
             if !carthageDependencies.isEmpty {
-                let carthagePlatformBuildPath = "$(PROJECT_DIR)/" + carthageResolver.buildPath(for: target.platform)
-                configFrameworkBuildPaths = [carthagePlatformBuildPath] + frameworkBuildPaths.sorted()
+                var carthagePlatformBuildPaths: [String] = []
+                if carthageDependencies.contains(where: { isStaticDependency(for: $0) }) {
+                    let carthagePlatformBuildPath = "$(PROJECT_DIR)/" + carthageResolver.buildPath(for: target.platform, isStatic: true)
```
giginet: If targets have any static frameworks, the static framework search path will be added.

Tests/XcodeGenKitTests/ProjectGeneratorTests.swift
```diff
@@ -1002,6 +1002,70 @@ class ProjectGeneratorTests: XCTestCase {
...
+                        let target = pbxProject.nativeTargets.first!
+                        let configuration = target.buildConfigurationList!.buildConfigurations.first!
+                        try expect(configuration.buildSettings["FRAMEWORK_SEARCH_PATHS"] as? [String]) == ["$(inherited)", "$(PROJECT_DIR)/Carthage/Build/iOS/Static"]
+                        let frameworkBuildPhase = try target.frameworksBuildPhase()
+                        guard let files = frameworkBuildPhase?.files, let file = files.first else {
+                            return XCTFail("frameworkBuildPhase should have files")
+                        }
+                        try expect(file.file?.nameOrPath) == "MyStaticFramework.framework"
+                        
+                        try expect(target.carthageCopyFrameworkBuildPhase()).beNil()
```
giginet: A project contains only static dependencies, `Carthage` build phase should not be added.

Tests/XcodeGenKitTests/ProjectGeneratorTests.swift
```diff
@@ -1002,6 +1002,70 @@ class ProjectGeneratorTests: XCTestCase {
...
+                            return XCTFail("Framework Build Phase should have Dynamic Framework")
+                        }
+                        guard let _ = files.first(where: { $0.file?.nameOrPath == "MyStaticFramework.framework" }) else {
+                            return XCTFail("Framework Build Phase should have Static Framework")
+                        }
+                        
+                        guard let copyCarthagePhase = target.carthageCopyFrameworkBuildPhase() else {
+                            return XCTFail("Carthage Build Phase should be exist")
+                        }
+                        try expect(copyCarthagePhase.inputPaths) == [dynamicFramework.file?.fullPath(sourceRoot: Path("$(SRCROOT)"))?.string]
```
giginet: Static Framework should not be copied.

yonaskolb:
@brentleyjones were you looking at this work earlier?

brentleyjones:
>were you looking at this work earlier?

@yonaskolb I was, but have since long abandoned that branch, and currently don't have strong opinions about static linking. I'll give this a once over though.

Docs/ProjectSpec.md
```diff
@@ -415,6 +421,7 @@ targets:
       - framework: path/to/framework.framework
       - carthage: Result
         findFrameworks: true
+        static: false
```
brentleyjones: Since this defaults to `false`, do we want to instead show the customization? Same for `findFrameworks`.

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -459,6 +459,15 @@ public class PBXProjGenerator {
             }
             return !linkingAttributes.isEmpty ? ["ATTRIBUTES": linkingAttributes] : nil
         }
+        
+        func isStaticDependency(for carthageDependency: Dependency) -> Bool {
```
brentleyjones: Could this instead be an extension on `Dependency`?

Sources/XcodeGenKit/PBXProjGenerator.swift
brentleyjones: We could flip this and check for the `if isStaticDependency(for: dependency) {` case first, then the else would be `else if embed`.

Sources/ProjectSpec/Dependency.swift
```diff
@@ -36,7 +36,7 @@ public struct Dependency: Equatable {
     public enum DependencyType: Equatable {
         case target
         case framework
-        case carthage(findFrameworks: Bool?)
+        case carthage(findFrameworks: Bool?, static: Bool)
```
brentleyjones: @yonaskolb If we required Swift 5.1 to compile we could use a default value here instead of having to have `static: false` added everywhere.

Docs/ProjectSpec.md
giginet: This is an example, so we can change to `static: true` for intelligible.

Sources/XcodeGenKit/PBXProjGenerator.swift
giginet: This is a local function inside `generateTarget` function. So it is a narrower scope than `fileprivate extension`. This is why I use function instead of extension.

Sources/ProjectSpec/Dependency.swift
giginet: CI seems to be still run on Xcode 10. https://github.com/yonaskolb/XcodeGen/blob/03b74b37c0b7659b7c6de8ad4d6e69e4491391a1/.github/workflows/ci.yml#L11

So I maintained backward compatibility.

giginet:
@brentleyjones Thanks for your review!

I addressed your points. Could you review this again?

Sources/XcodeGenKit/PBXProjGenerator.swift
brentleyjones: Yep!

giginet:
@yonaskolb We need your eyes before merging. Could you check this in your spare time?

Sources/ProjectSpec/Dependency.swift
yonaskolb: 1. How do you feel about making this a `linkType: LinkType` enum with `static` and `dynamic` cases? I don't feel too strongly either way. I think we @brentleyjones was building this he mentioned a similar approach.
2. Are there other dependency types that we might want use this on? If so we could move this property into the Dependency itself, as opposed to the enum

Sources/XcodeGenKit/PBXProjGenerator.swift
yonaskolb: I agree about moving this to an extension. If we move static/linkType to Dependency itself like in the other comment we don't even need this function. 

Otherwise if this crashes when used with other types it should be called `carthageDependencyIsStatic`, otherwise it should probably just return false for other cases

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -463,6 +463,15 @@ public class PBXProjGenerator {
...
```
yonaskolb: I agree about moving this to an extension. If we move static/linkType to Dependency itself like in the other comment we don't even need this function. 

Otherwise if this crashes when used with other types it should be called `carthageDependencyIsStatic`, otherwise it should probably just return false for other cases

Sources/ProjectSpec/Dependency.swift
giginet: > 1
I feel good `linkType` than the current implementation 👍 

> 2
> Are there other dependency types that we might want use this on?

No. I think it's easier to read `Dependency` has linkType property than using associated value.
But using associated value is more right to express structures.
That's difficult choices...

giginet:
@yonaskolb @brentleyjones I modified spec interface from your comments. Could you review again?
https://github.com/yonaskolb/XcodeGen/pull/688/commits/705e3512a3b6b3998c8a03bd379664a36f0a5f11

Sources/ProjectSpec/Dependency.swift
giginet: CI environments for legacy Swift versions are dropped on #709.

So I used default associated value. 
https://github.com/yonaskolb/XcodeGen/pull/688/commits/75905237dee8aea274fbdecb2a52b77c7689055d

Sources/ProjectSpec/Dependency.swift
```diff
@@ -32,11 +32,18 @@ public struct Dependency: Equatable {
...
+        case `static`
+        
+        public static let `default` = dynamic
+    }
 
     public enum DependencyType: Equatable {
         case target
         case framework
-        case carthage(findFrameworks: Bool?)
+        case carthage(findFrameworks: Bool?, linkType: CarthageLinkType = .default)
```
brentleyjones: I'm not sure if we are dropping lower Swift compiler versions here yet, so we shouldn't use a default associated value.

giginet: reverted. https://github.com/yonaskolb/XcodeGen/pull/688/commits/14a249d7c306db3618333c34f6a7441916c24242

PR #689, opened 2019-10-22
Don't generate CFBundleExecutable for targets of type bundle

FranzBusch:
When declaring a target like this

```
  BundleTarget:
    platform: iOS
    type: bundle
    info:
      path: some/path/Info.plist
    sources: [some/path/sources]
```

The generated Info.plist file contains a key `CFBundleExecutable` by default. This key indicates that the bundle has a executable. When uploading an application with such a bundle that contains the Info.plist AppStore connect will decline the binary since a bundle by default doesn't contain any executable but rather only resources. Therefore, I suggest we do not generate the `CFBundleExecutable` by default for targets of type `bundle`.

### DIFF
--- a/Sources/XcodeGenKit/InfoPlistGenerator.swift
+++ b/Sources/XcodeGenKit/InfoPlistGenerator.swift
@@ -8,7 +8,7 @@ public class InfoPlistGenerator {
      Default info plist attributes taken from:
      /Applications/Xcode.app/Contents/Developer/Library/Xcode/Templates/Project Templates/Base/Base_DefinitionsInfoPlist.xctemplate/TemplateInfo.plist
      */
-    private func generateDefaultInfoPlist(target: Target) -> [String: Any] {
+    private func generateDefaultInfoPlist(for target: Target) -> [String: Any] {
         var dictionary: [String: Any] = [:]
         dictionary["CFBundleIdentifier"] = "$(PRODUCT_BUNDLE_IDENTIFIER)"
         dictionary["CFBundleInfoDictionaryVersion"] = "6.0"
@@ -26,8 +26,8 @@ public class InfoPlistGenerator {
         return dictionary
     }
 
-    public func generateProperties(target: Target) -> [String: Any] {
-        var targetInfoPlist = generateDefaultInfoPlist(target: target)
+    public func generateProperties(for target: Target) -> [String: Any] {
+        var targetInfoPlist = generateDefaultInfoPlist(for: target)
         switch target.type {
         case .uiTestBundle,
              .unitTestBundle:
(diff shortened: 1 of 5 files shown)


Sources/XcodeGenKit/InfoPlistGenerator.swift
```diff
@@ -8,20 +8,26 @@ public class InfoPlistGenerator {
      Default info plist attributes taken from:
      /Applications/Xcode.app/Contents/Developer/Library/Xcode/Templates/Project Templates/Base/Base_DefinitionsInfoPlist.xctemplate/TemplateInfo.plist
      */
-    var defaultInfoPlist: [String: Any] = {
+    private func generateDefaultInfoPlist(target: Target) -> [String: Any] {
```
giginet: According to [the naming guideline](https://swift.org/documentation/api-design-guidelines/#parameter-names), I prefer to be more grammatical name like `generateDefaultInfoPlist(for:)`

Tests/XcodeGenKitTests/ProjectGeneratorTests.swift
```diff
@@ -1002,6 +1002,35 @@ class ProjectGeneratorTests: XCTestCase {
...
+                let data: Data = try infoPlistFile.read()
+                let infoPlist = try PropertyListSerialization.propertyList(from: data, options: [], format: nil) as! [String: Any]
+                var expectedInfoPlist: [String: Any] = [:]
+                expectedInfoPlist["CFBundleIdentifier"] = "$(PRODUCT_BUNDLE_IDENTIFIER)"
+                expectedInfoPlist["CFBundleInfoDictionaryVersion"] = "6.0"
+                expectedInfoPlist["CFBundleName"] = "$(PRODUCT_NAME)"
+                expectedInfoPlist["CFBundleDevelopmentRegion"] = "$(DEVELOPMENT_LANGUAGE)"
+                expectedInfoPlist["CFBundleShortVersionString"] = "1.0"
+                expectedInfoPlist["CFBundleVersion"] = "1"
+                expectedInfoPlist["CFBundlePackageType"] = "BNDL"
```
giginet: You can make it be immutable variable.

```swift
let expectedInfoPlist: [String: Any] = [
    "CFBundleIdentifier": "$(PRODUCT_BUNDLE_IDENTIFIER)",
]
```

Sources/XcodeGenKit/InfoPlistGenerator.swift
```diff
@@ -8,20 +8,26 @@ public class InfoPlistGenerator {
...
         dictionary["CFBundleIdentifier"] = "$(PRODUCT_BUNDLE_IDENTIFIER)"
         dictionary["CFBundleInfoDictionaryVersion"] = "6.0"
-        dictionary["CFBundleExecutable"] = "$(EXECUTABLE_NAME)"
+
         dictionary["CFBundleName"] = "$(PRODUCT_NAME)"
         dictionary["CFBundleDevelopmentRegion"] = "$(DEVELOPMENT_LANGUAGE)"
         dictionary["CFBundleShortVersionString"] = "1.0"
         dictionary["CFBundleVersion"] = "1"
+
+        // Bundles should not contain any CFBundleExecutable otherwise they will be rejected when uploading.
```
giginet: Could you update ProjectSpec documentation too?
https://github.com/yonaskolb/XcodeGen/blob/master/Docs/ProjectSpec.md#target

> The following properties are generated automatically

FranzBusch:
@giginet Thanks for your review, I addressed your comments.

PR #692, opened 2019-10-22
Updated Default Value

### LINKED ISSUE #691: Default CODE_SIGN_IDENTITY
As a Developer I want to ensure that my build scripts work from a clean installation of the project, Xcodegen is a critical component of that process.

With the release of Mac Catalyst, the default signing identity certificates have been renamed from "iPhone Developer" to "Apple Development" so that a single certificate now covers both iOS and Mac signing.  

Not updating the CODE_SIGN_IDENTITY build setting can result in failed builds.

The `application_iOS.yml` should be updated to reflect this change the default signing.

dpfannenstiel:
Updated default CODE_SIGN_IDENTITY value to reflect changes to support Mac Catalyst distribution.

Fixes #691


SettingPresets/Product_Platform/application_iOS.yml
```diff
@@ -1,2 +1,2 @@
-CODE_SIGN_IDENTITY: iPhone Developer
+CODE_SIGN_IDENTITY: Apple Development
```
yonaskolb: If this is the case should it be changed in the other presets as well?

yonaskolb:
To get CI to pass you’ll need to run the tests and commit the updated fixtures

SettingPresets/Product_Platform/application_iOS.yml
giginet: Does this change still support working for Xcode 10?
I think XcodeGen should switch presets by `xcodeVersion`.

dpfannenstiel: You're right about Xcode 10.  I'll take another pass at a more sophisticated implementation.

clausjoergensen:
Why does this default even exist? It causes Xcode to ignore `CODE_SIGN_IDENTITY` from xcconfig files...

PR #694, opened 2019-10-22
Add Base to knownRegions even if one doesn't exist

bryansum:
Addendum to https://github.com/yonaskolb/XcodeGen/pull/685 in the case there's   
no `Base.lproj` folder but one for the development language -- in this case      
XcodeGen generates a `knownRegions` result of only `en`, which leads to warnings 
in Xcode 11. Adding `Base` by default in this case silences Xcode.

### DIFF
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -224,8 +224,8 @@ public class PBXProjGenerator {
         let projectAttributes: [String: Any] = ["LastUpgradeCheck": project.xcodeVersion]
             .merged(project.attributes)
 
-        let knownRegions = sourceGenerator.knownRegions.sorted()
-        pbxProject.knownRegions = knownRegions.isEmpty ? ["en"] : knownRegions
+        let knownRegions = sourceGenerator.knownRegions
+        pbxProject.knownRegions = (knownRegions.isEmpty ? ["en"] : knownRegions).union(["Base"]).sorted()
         pbxProject.packages = packageReferences.sorted { $0.key < $1.key }.map { $1 }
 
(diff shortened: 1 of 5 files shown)


Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -222,8 +222,7 @@ public class PBXProjGenerator {
         let projectAttributes: [String: Any] = ["LastUpgradeCheck": project.xcodeVersion]
             .merged(project.attributes)
 
-        let knownRegions = sourceGenerator.knownRegions.sorted()
-        pbxProject.knownRegions = knownRegions.isEmpty ? ["en"] : knownRegions
+        pbxProject.knownRegions = sourceGenerator.knownRegions.union(["Base"]).sorted()
```
yonaskolb: It looks like if a project has no localized files it will now default to having a `Base` instead of `en`. I guess that's ok. I can't imagine it breaking anything?

yonaskolb: It looks like if a project has no localized files it will now default to having a `Base` instead of `en`. I guess that's ok. I can't imagine it breaking anything?

yonaskolb:
Could you also add a changelog entry. And to get tests to pass you'll need to run the tests and then check in the updated fixtures

Sources/XcodeGenKit/PBXProjGenerator.swift
bryansum: Yeah on consideration, when opening up a new Xcode project `knownRegions` is set to `en, Base` (for me) so we should probably keep your existing logic for `isEmpty` in addition to the union.

raptorxcz:
Impatiently waiting for this 🙂 🙏 Release this soon.

yonaskolb:
Yep, will do a release once https://github.com/yonaskolb/XcodeGen/pull/700 is merged 👍

PR #696, opened 2019-10-24
Fix macOS unit test target TEST_HOST 

### LINKED ISSUE #695: No Test Host set on macOS
I opened #678 a while ago and happy to report that the last release fixed that regression 👍 

The next problem I've run in to, is that while my test target is generated, the Host Application for the test target is blank.

Example project yaml:

```name: Project
settings:
  DEVELOPMENT_TEAM: 
options:
    bundleIdPrefix:  
targets:
    DirectTests:
      platform: macOS
      type: bundle.unit-test
      sources:
        - Tests/
      dependencies:
        - target: Direct
      gatherCoverageData: true
      settings:
        configs:
          debug:
            SWIFT_ACTIVE_COMPILATION_CONDITIONS: MACOSTESTING_DIRECT
    Direct:
      type: application
      platform: macOS
      deploymentTarget: "10.14"
      sources:
        - /Common
        - /macOS
      settings:
        PRODUCT_BUNDLE_IDENTIFIER: 
      scheme:
        testTargets:
          - DirectTests
```
I've looked through some previous issues and can see tangentially related things, but nothing that seems obvious (to me). What am I doing wrong?

If I do the same thing with an iOS target it works fine.

mjarvis:
Resolves #695

Previously, macOS unit test targets would appear to have no test host.

By manually setting the test host in Xcode on the fixture project, I discovered that macOS test hosts require an additional `Contents/MacOS/` path component before the product name.

We can manually detect the platform when setting the test host and as such generate the host correctly for macOS unit test targets.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Don't generate CFBundleExecutable in default generated Info.plist for `bundle` target types [#689](https://github.com/yonaskolb/XcodeGen/pull/689) @FranzBusch
 - Fixed resolving relative paths with custom project destination [#681](https://github.com/yonaskolb/XcodeGen/pull/681) @giginet
 - Fixed resolving relative paths for Info.plist [#683](https://github.com/yonaskolb/XcodeGen/pull/683)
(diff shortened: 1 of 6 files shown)


BalintZoo:
Was this fixed for iOS as well? I'm experiencing the same problems there.

Issue #609, opened 2019-07-18
buildPhase=headers does not work when type:framework.static


samchang8826:
```
targets:
   xxxx:
    type: framework.static
    platform: iOS
    deploymentTarget: "12.0"
    sources:
      - path: xxx
        excludes:
          - "*[!h]"
        buildPhase: headers
```

the headers section on buildPhase is empty.

PR #705, opened 2019-11-06
Fix header phase for static frameworks

wag-miles:
Corrects a check that was removing all headers from static frameworks. https://github.com/yonaskolb/XcodeGen/issues/609

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 #### Fixed
 - Improved variable expansion runtime [#704](https://github.com/yonaskolb/XcodeGen/pull/704) @rcari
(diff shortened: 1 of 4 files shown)


wag-miles:
Not averse to adding a test for this but would need some guidance

wag-miles:
Added a changelog entry. Not sure why the change is causing CI to fail

brentleyjones:
> Added a changelog entry. Not sure why the change is causing CI to fail

CI is failing because Xcode 10.3 is no longer supported it seems. #707 will resolve it for the time being.

brentleyjones:
@wag-miles please rebase on master to get the updated CI

brentleyjones:
Thanks @wag-miles!

PR #712, opened 2019-11-08
Support optional include

giginet:
# Motivation

I want to add targets dynamically. including subspecs help me in the situation.
The included spec is automatically generated and if not exists ignore adding targets.

# Description

I added `optional` option to `Include`.
This value indicates whether ignoring failure loading (e.g. the path does not exist).

```yaml
include:
  - path: not_exist.yml
    optional: true
name: OptionalInclude
targets:
  IncludedTarget:
    type: application
    platform: iOS
    sources:
      - template
```

### DIFF
--- a/Sources/ProjectSpec/SpecFile.swift
+++ b/Sources/ProjectSpec/SpecFile.swift
@@ -13,17 +13,21 @@ public struct SpecFile {
     fileprivate struct Include {
         let path: Path
         let relativePaths: Bool
+        let optional: Bool
 
         static let defaultRelativePaths = true
+        static let defaultOptional = false
 
         init?(any: Any) {
             if let string = any as? String {
                 path = Path(string)
                 relativePaths = Include.defaultRelativePaths
+                optional = Self.defaultOptional
             } else if let dictionary = any as? JSONDictionary,
                 let path = dictionary["path"] as? String {
                 self.path = Path(path)
                 relativePaths = dictionary["relativePaths"] as? Bool ?? Include.defaultRelativePaths
+                optional = dictionary["optional"] as? Bool ?? Self.defaultOptional
             } else {
                 return nil
             }
@@ -64,8 +68,16 @@ public struct SpecFile {
         let jsonDictionary = try SpecFile.loadDictionary(path: path)
 
         let includes = Include.parse(json: jsonDictionary["include"])
-        let subSpecs: [SpecFile] = try includes.map { include in
-            try SpecFile(include: include, basePath: basePath, relativePath: relativePath)
+        let subSpecs: [SpecFile] = try includes.compactMap { include in
+            do {
+                return try SpecFile(include: include, basePath: basePath, relativePath: relativePath)
+            } catch {
+                if include.optional {
+                    return nil
+                } else {
+                    throw error
+                }
+            }
         }
 
         self.init(filename: filename, jsonDictionary: jsonDictionary, basePath: basePath, relativePath: relativePath, subSpecs: subSpecs)
--- a/Tests/ProjectSpecTests/SpecLoadingTests.swift
+++ b/Tests/ProjectSpecTests/SpecLoadingTests.swift
@@ -41,6 +41,25 @@ class SpecLoadingTests: XCTestCase {
                     Target(name: "NewTarget", type: .application, platform: .iOS, sources: ["template", "target"]),
                 ]
             }
+            
+            $0.context("with optional include for non exist subspecs") {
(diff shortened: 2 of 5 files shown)


Sources/ProjectSpec/SpecFile.swift
```diff
@@ -13,17 +13,21 @@ public struct SpecFile {
...
+        let optional: Bool
 
         static let defaultRelativePaths = true
+        static let defaultOptional = false
 
         init?(any: Any) {
             if let string = any as? String {
                 path = Path(string)
                 relativePaths = Include.defaultRelativePaths
+                optional = Self.defaultOptional
```
giginet: `Self.` will work only on Swift 5.1 or above.

Tests/ProjectSpecTests/SpecLoadingTests.swift
```diff
@@ -41,6 +41,28 @@ class SpecLoadingTests: XCTestCase {
                     Target(name: "NewTarget", type: .application, platform: .iOS, sources: ["template", "target"]),
                 ]
             }
+            
+            $0.context("with optional include for non exist subspecs") {
+                $0.it("ignore subspecs") {
+                    let path = fixturePath + "optional_include/optional_include.yml"
+                    let project = try XCTUnwrap(loadSpec(path: path))
```
giginet: `XCTUnwrap` will work only on Swift 5.1 or above.

yonaskolb: Can you use the new `unwrap` function from recently merged https://github.com/yonaskolb/XcodeGen/pull/714

Sources/ProjectSpec/SpecFile.swift
yonaskolb: I've updated the rest of the codebase to Swift 5.1 in https://github.com/yonaskolb/XcodeGen/pull/714

Sources/ProjectSpec/SpecFile.swift
```diff
@@ -64,8 +68,16 @@ public struct SpecFile {
...
 
         let includes = Include.parse(json: jsonDictionary["include"])
-        let subSpecs: [SpecFile] = try includes.map { include in
-            try SpecFile(include: include, basePath: basePath, relativePath: relativePath)
+        let subSpecs: [SpecFile] = try includes.compactMap { include in
+            do {
+                return try SpecFile(include: include, basePath: basePath, relativePath: relativePath)
+            } catch {
+                if include.optional {
+                    return nil
```
yonaskolb: This would make an optional includes silently fail if there was an different error than simply a missing file, for example if one of the includes own includes was missing. Is that the intention? Maybe it's better to only return nil if the file is actually missing

giginet: > Maybe it's better to only return nil if the file is actually missing

Do you mean `optional` should ignore throwing from `try path.read()` ?
https://github.com/yonaskolb/XcodeGen/blob/15a4f6abe601cbe5750276f93b572bd723e45912/Sources/ProjectSpec/SpecFile.swift#L89

Sources/ProjectSpec/SpecFile.swift
giginet: I'll review #714

giginet:
@yonaskolb Following your comments I implemented a new strategy to cause errors.
https://github.com/yonaskolb/XcodeGen/pull/712/commits/037d44b5c8987a3125d6498f0a64b819fd07d7cf

However, prior implementation seems to be simpler...
How do you feel about?

Tests/ProjectSpecTests/SpecLoadingTests.swift
giginet: fixed

Sources/ProjectSpec/SpecFile.swift
```diff
@@ -60,40 +60,41 @@ public struct SpecFile {
...
         self.init(filename: filename, jsonDictionary: jsonDictionary, basePath: basePath, relativePath: relativePath, subSpecs: subSpecs)
     }
 
-    static func loadDictionary(path: Path) throws -> JSONDictionary {
+    static func loadDictionary(path: Path, shouldIgnoreNotFound: Bool) throws -> JSONDictionary {
+        let data: Data
+        do {
+            data = try path.read()
+        } catch {
+            if shouldIgnoreNotFound {
```
yonaskolb: I was thinking throwing a `GenerationError.includeNotFound`, which you can then catch at the top level, and it it's not optional then rethrow

PR #719, opened 2019-11-17
Plist.path as Path

RomanPodymov:
Hello.
Thank you for XcodeGen.
In this pull request I modified `Plist.path` type. Now it is `Path`, not `String`.

### DIFF
--- a/Sources/ProjectSpec/Plist.swift
+++ b/Sources/ProjectSpec/Plist.swift
@@ -1,12 +1,13 @@
 import Foundation
 import JSONUtilities
+import PathKit
 
 public struct Plist: Equatable {
 
-    public let path: String
+    public let path: Path
     public let properties: [String: Any]
 
-    public init(path: String, attributes: [String: Any] = [:]) {
+    public init(path: Path, attributes: [String: Any] = [:]) {
         self.path = path
         properties = attributes
     }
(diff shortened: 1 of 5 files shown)


Sources/ProjectSpec/Plist.swift
```diff
@@ -1,12 +1,13 @@
...
+import PathKit
 
 public struct Plist: Equatable {
 
-    public let path: String
+    public let path: Path
     public let properties: [String: Any]
 
-    public init(path: String, attributes: [String: Any] = [:]) {
+    public init(path: Path, attributes: [String: Any] = [:]) {
```
giginet: Public interfaces are modified

RomanPodymov:
> Could you tell us the motivation for the changes?
> 
> This PR brakes backward-compatibility. So I want what's the motivation of this change.

Hello @giginet 
Originally I had the issue with `INFOPLIST_FILE`. If the name of this file is not **Info.plist** XcodeGen copies it to the resources that leads to a warning in Xcode. However XcodeGen does not log it.
I planned to fix this issue and as the first step I opened this pull request. With `path` declared as `Path` it is very easy to get file name and other properties.
I know that this pull request changes public API. As a possible workaround I can add a new property `let filePath: Path` and mark `let path: String` as deprecated.

yonaskolb:
Hi @RomanPodymov, as @giginet says I’d rather not change the public API for now. We’ve kept path out of all the decoded properties that are actually paths. At some point we could perhaps upgrade them all.

If you’re using XcodeGen as a library you can add any extensions you need in your own code in the meantime. 
If you’re using this as a first step to change behaviour within XcodeGen itself, you can always create a path at the usage sight. If you created a PR with that, we’d get a better sense of your needs as well. If it led to a huge amount of code we can revisit this change.

I’ll close this for now, but thanks for the contribution! 😃👍

PR #701, opened 2019-10-30
Support external target dependencies via subprojects

evandcoleman:
Resolves #570 

This builds upon the work done in #655 by allowing target dependencies to reference targets in other Xcode projects.

### Motivation

At my company, we have several internal dependencies that we include via submodules. We'd like the ability to include these dependencies as subprojects and reference their targets in our main Xcode project.

### Changes

As introduced in #655, subprojects are declared at the top level.

```yaml
projectReferences:
  FooLib:
    path: path/to/FooLib.xcodeproj
targets:
  MyApp:
    type: application
    platform: iOS
    deploymentTarget: "11.0"
    sources: [Sources]
    dependencies:
      - target: FooLib/MyTarget
```

PR #723, opened 2019-11-20
Add bundle dependency type

kateinoigakukun:
This PR cherry-picks https://github.com/yonaskolb/XcodeGen/pull/616 changes related with bundle


Tests/XcodeGenKitTests/ProjectGeneratorTests.swift
```diff
@@ -622,17 +631,24 @@ class ProjectGeneratorTests: XCTestCase {
...
+                    }
+
+                    if let expectedBundlesFiles = expectedBundlesFiles[target.name],
+                        target.type != .staticLibrary && target.type != .dynamicLibrary {
+                        expectedCopyFilesPhasesCount += 1
+                        let copyBundles = (copyBundlesPhase?.files ?? [])
+                            .compactMap { $0.file?.nameOrPath }
+                        try expect(Set(copyBundles)) == expectedBundlesFiles
                     }
+                    print(target.name)
```
yonaskolb: Can we remove the print

kateinoigakukun:
I updated to use this feature in TestProject 👍

yonaskolb:
As CI actually builds the fixtures to make sure everything is working they have to be buildable. That means you'll need to add an actual bundle, even if there's nothing in there.
I notice as well that it's failing at build time instead of at generation time. We probably need to add a check in the SpecValidator to make sure the bundle exists. Would anyone ever have an optional bundle? If so, then we could add an `optional: true` property as well

kateinoigakukun:
@yonaskolb 

First, there is two case to embed bundle as dependency.

kateinoigakukun:
@yonaskolb 

First, there is 3 cases to embed bundle as dependency.

1. Reference a bundle target in `dependencies` of `project.yml`

e.g.
```yaml
targets:
  App:
    dependencies:
    - target: BundleX
  BundleX:
    type: bundle
```


2. Copy an existing bundle directory from `sources`

e.g.
```yaml
targets:
  App:
    sources:
      - Sources/BundleY.bundle
```

3. Reference **another project's bundle target** as a dependency

e.g.
```yaml
targets:
  App:
    dependencies:
    - bundle: BundleZ.bundle
```

This PR supports 3rd case new and it always reference the target implicitly.
So we can't validate it while generating xcodeproj.

Docs/ProjectSpec.md
```diff
@@ -374,13 +374,14 @@ targets:
...
 
-A dependency can be one of a 3 types:
+A dependency can be one of a 6 types:
 
 - `target: name` - links to another target
 - `framework: path` - links to a framework
 - `carthage: name` - helper for linking to a Carthage framework
 - `sdk: name` - links to a dependency with the SDK. This can either be a relative path within the sdk root or a single filename that references a framework (.framework) or lib (.tbd)
 - `package: name` - links to a Swift Package. The name must match the name of a package defined in the top level `packages`
+- `bundle: name` - adds the pre-built bundle for the supplied name to the copy resources build phase. This is useful when a dependency exists on a static library target that has an associated bundle target, both existing in a separate project. Only usable in target types which can copy resources. Always use `implicit: true`. This will not cause the bundle to be explicitly built.
```
yonaskolb: Are there cases where you would be `implicit: false`. If it's always implicit we can remove reference to `implicit: true` here

Docs/ProjectSpec.md
```diff
@@ -392,7 +393,7 @@ A dependency can be one of a 3 types:
 
 **Implicit Framework options**:
 
-This only applies to `framework` dependencies. Implicit framework dependencies are useful in Xcode Workspaces which have multiple `.xcodeproj` that are not embedded within each other yet have a dependency on a framework built in an adjacent `.xcodeproj`.  By having `Find Implicit Dependencies` checked within your scheme `Build Options` Xcode can link built frameworks in `BUILT_PRODUCTS_DIR`.
+This only applies to `framework` and `bundle` dependencies. Implicit framework dependencies are useful in Xcode Workspaces which have multiple `.xcodeproj` that are not embedded within each other yet have a dependency on a framework built in an adjacent `.xcodeproj`.  By having `Find Implicit Dependencies` checked within your scheme `Build Options` Xcode can link built frameworks in `BUILT_PRODUCTS_DIR`.
```
yonaskolb: Does it apply to `bundle` dependencies? Can you turn it off?

kateinoigakukun:
@yonaskolb I checked the PR. In most case, it's enough to use implicit reference to build targets of another project and it covers same feature as #701 only about bundle.
The only difference I could see between these is whether the build is triggered implicitly by Xcode.

#701 do a lot of configuration to trigger dependency builds without any Xcode. On the other hand, this PR relies on Xcode build system to build the bundle.

kateinoigakukun:
@yonaskolb Can I merge this?

PR #682, opened 2019-10-08
[WIP] Spec generation

ryohey:
Although it is not a complete PR, it is now operating to some extent. There are some issues and I would like to hear your feedback.

## Implementation status

- [x] Write Project to a yaml file
- [ ] Generate Project from XcodeProj
  - roughly done
  - some project settings are not treated yet
- [ ] Remove default values from yaml
- [ ] Change to proper command names and CLI improvements
- [ ] Fix to find C header files appropriate way

About the last item. As far as I know, C header files are not added to the build phase on the application target, so we have to find them from the group. Currently I am adding all header files in the group as sources, but if the group doesn't match the target it will cause problems. If you know anything about how to deal with this, please let me know.

## output example 

```bash
xcodegen generate-spec --project /path/to/project/Fireside.xcodeproj --spec  /path/to/project/project.yml
```

This produces below

<details>
<summary>project.yml</summary>

```yaml
attributes:
  LastUpgradeCheck: 1020
configs:
  Debug: debug
  Release: release
name: Fireside
options:
  groupSortPosition: bottom
  transitivelyLinkDependencies: false
settingGroups:
  Debug:
    ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES: YES
    ALWAYS_SEARCH_USER_PATHS: NO
    CLANG_ANALYZER_NONNULL: YES
    CLANG_ANALYZER_NUMBER_OBJECT_CONVERSION: YES_AGGRESSIVE
    CLANG_CXX_LANGUAGE_STANDARD: gnu++14
    CLANG_CXX_LIBRARY: libc++
    CLANG_ENABLE_MODULES: YES
    CLANG_ENABLE_OBJC_ARC: YES
    CLANG_WARN_BLOCK_CAPTURE_AUTORELEASING: YES
    CLANG_WARN_BOOL_CONVERSION: YES
    CLANG_WARN_COMMA: YES
    CLANG_WARN_CONSTANT_CONVERSION: YES
    CLANG_WARN_DIRECT_OBJC_ISA_USAGE: YES_ERROR
    CLANG_WARN_DOCUMENTATION_COMMENTS: YES
    CLANG_WARN_EMPTY_BODY: YES
    CLANG_WARN_ENUM_CONVERSION: YES
    CLANG_WARN_INFINITE_RECURSION: YES
    CLANG_WARN_INT_CONVERSION: YES
    CLANG_WARN_NON_LITERAL_NULL_CONVERSION: YES
[...]

PR #735, opened 2019-12-04
Spec generation

ryohey:
moved from https://github.com/yonaskolb/XcodeGen/pull/682

## Implementation status

- [x] Write Project to a yaml file
- [ ] Generate Project from XcodeProj
  - roughly done
  - some project settings are not treated yet
- [ ] Remove default values from yaml
- [ ] Change to proper command names and CLI improvements
- [ ] Fix to find C header files appropriate way

About the last item. As far as I know, C header files are not added to the build phase on the application target, so we have to find them from the group. Currently I am adding all header files in the group as sources, but if the group doesn't match the target it will cause problems.

### DIFF
--- a/Sources/ProjectSpec/CollectionExtensions.swift
+++ b/Sources/ProjectSpec/CollectionExtensions.swift
@@ -1,0 +1,33 @@
+import Foundation
+
+private protocol EmptyRemovable {
+    func removeEmpty() -> Self
+    var isEmpty: Bool { get }
+}
+
+extension Array: EmptyRemovable {
+    public func removeEmpty() -> Array {
+        return compactMap {
+            if let e = ($0 as? EmptyRemovable)?.removeEmpty() {
+                return e.isEmpty ? nil : e as? Element
+            }
+            return $0
+        }
+    }
+}
+
+extension Dictionary: EmptyRemovable {
+    // this is a little trick that defines the generics parameter from optional Value to unwrapped Value
+    private static func removeEmpty<Key, Value>(dict: Dictionary<Key, Value?>) -> Dictionary<Key, Value> {
+        return dict.compactMapValues {
+            if let e = ($0 as? EmptyRemovable)?.removeEmpty() {
+                return e.isEmpty ? nil : e as? Value
+            }
+            return $0
+        }
+    }
+
+    public func removeEmpty() -> Dictionary<Key, Value> {
+        return Dictionary.removeEmpty(dict: self)
+    }
+}
--- a/Sources/ProjectSpec/Settings.swift
+++ b/Sources/ProjectSpec/Settings.swift
@@ -110,11 +110,35 @@ extension Settings: JSONEncodable {
     public func toJSONValue() -> Any {
         if groups.count > 0 || configSettings.count > 0 {
             return [
-                "base": buildSettings,
+                "base": buildSettings.toJSONValue(),
                 "groups": groups,
                 "configs": configSettings.mapValues { $0.toJSONValue() },
             ]
         }
-        return buildSettings
+        return buildSettings.toJSONValue()
+    }
+}
+
+extension BuildSettings: JSONEncodable {
+    private func asJSONValue(_ value: Any) -> Any {
+        switch value {
+        case let value as Path:
+            return value.string
+        case let value as [Any]:
+            return value.map(asJSONValue)
+        case let value as [String: Any]:
+            return value.mapValues(asJSONValue)
+        case is Bool: fallthrough
+        case is Int: fallthrough
+        case is Double: fallthrough
+        case is String:
+            return value
+        default:
+            fatalError("not supported type")
+        }
+    }
+
+    public func toJSONValue() -> Any {
+        return mapValues(asJSONValue)
     }
 }
--- a/Sources/ProjectSpec/TargetSource.swift
+++ b/Sources/ProjectSpec/TargetSource.swift
@@ -213,7 +213,7 @@ extension TargetSource: JSONEncodable {
             dict["optional"] = optional
         }
 
-        if dict.count == 0 {
+        if dict.removeEmpty().isEmpty {
             return path
         }
 
--- a/Sources/XcodeGenCLI/GenerateSpecCommand.swift
+++ b/Sources/XcodeGenCLI/GenerateSpecCommand.swift
@@ -1,0 +1,32 @@
+import Foundation
+import PathKit
+import ProjectSpec
+import SwiftCLI
+import XcodeGenKit
+import XcodeProj
+import Yams
+
+class GenerateSpecCommand: Command {
+    let name: String = "generate-spec"
+
+    let projectFile = Key<Path>("-p", "--project", description: "The path to the project file")
+
+    let spec = Key<Path>(
+        "-s",
+        "--spec",
+        description: "The path to the project spec file should be generated. Defaults to project.yml"
+    )
+
+    func execute() throws {
+        guard let file = projectFile.value else {
+            return
+        }
+        let xcodeProj = try XcodeProj(path: file)
+        let project = try generateSpec(xcodeProj: xcodeProj, projectDirectory: file.parent())
+        let projectDict = project?.toJSONDictionary().removeEmpty()
+        let encodedYAML = try Yams.dump(object: projectDict)
+        let defaultOutPath = file.parent() + "project.yml"
+        let outPath = spec.value ?? defaultOutPath
+        try encodedYAML.write(toFile: outPath.string, atomically: true, encoding: .utf8)
+    }
+}
--- a/Sources/XcodeGenCLI/XcodeGenCLI.swift
+++ b/Sources/XcodeGenCLI/XcodeGenCLI.swift
@@ -7,6 +7,7 @@ public class XcodeGenCLI {
 
     public init(version: Version) {
         let generateCommand = GenerateCommand(version: version)
+        let specGenerationCommand = GenerateSpecCommand()
 
         cli = CLI(
             name: "xcodegen",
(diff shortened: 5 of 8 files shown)


mqln:
just letting y'all know how rad this feature would be!

siempay:
Hi! did this feature make it to production after all ?

2075:
do you need any support on this? imho this is a must have.

siempay:
Yes I wanna know if I can use it on excisting project ?

ryohey:
Hi, there

I've done a basic implementation. This can probably be used for simple project. 
Would anyone like to test it on some project?

usage

```bash
./XcodeGen migrate --project /path/to/ExistingProject.xcodeproj --spec generated.yaml
```

cc @2075 @siempay

4brunu:
Hey, I used this to try to simplify the migration to XcodeGen and it did 80% of the work! 😃 
It still needed some manual work, but it's a great starting point!
Maybe this could be introduced as an experimental functionality.
Thanks, this PR is awesome.

SShayashi:
Thank you for your awesome work, I also used on our project. It works well 80%.
I report some issues that I was faced with!
- Only create “debug”, “release” build configuration. you’d get the following error: 
`Build settings has invalid build configuration “Adhoc”`
   (You can easily fix just add config `Adhoc` to `Project` `configs` because any other targets already have `Adhoc` settings.)
- doesn’t create `Build phases` dependencies properly
- doesn’t consider `Build phases` order. This should happen an issue, such as R.swift scripts.
- doesn’t create scheme settings
- verbose compiler settings
- contains Carthage, Cocoapods settings
- `sources` have each file

I hope this would be helpful.

sfdux:
I am looking forward to this feature.  C header files are not added to the build phase on the application target, this is a problem...I report these issues below:
- Doesn’t create Build phases dependencies properly, such as embed
- delete scheme settings

ryohey:
@SShayashi @sfdux Thank you for your feedback! I am working on issues you reported.

- [ ] Only create “debug”, “release” build configuration
- [x] Doesn’t create Build phases dependencies properly
- [x] Doesn’t consider Build phases order. This should happen an issue, such as R.swift scripts.
- [ ] Doesn’t create scheme settings
- [ ] Creates verbose compiler settings
- [ ] Contains Carthage, Cocoapods settings
- [ ] sources have each file

Satoru-PriChan:
Hello!
 I tried hitting `xcodegen migrate --project MyProject.xcodeproj --spec project.yml` but it showed an error message: "Error: The folder “en.lproj” doesn’t exist". Does anyone know the cause of the error? Thank you.

ryohey:
@Satoru-PriChan It seems that some localized files are missing on the disk. Can you build that project without error?

Satoru-PriChan:
@ryohey Yeah, it compiles fine.

I found the cause of the error myself. I use a library named "[ALCameraViewController](https://github.com/AlexLittlejohn/ALCameraViewController)", which has a .strings file. The file is localized into English and located in en.lproj folder. So compiles no problem. 

スクリーンショット 2021-04-09 10 38 34

But the folder is not located at the root of my project but at the root of ALCameraViewController: <MyProject root> > ALCameraViewController > en.lproj > CameraView.strings. I disabled the English localization of the file, then `xcodegen migrate` started working fine. I guess that is because `xcodegen migrate` can't find en.lproj folder that is not at the root of the whole project.

sfdux:
Hello, I'd like to know the progress of Spec generation? When will it be released...I've been waiting for this feature

4brunu:
I already mention this, but I think this feature should be introduce as experimental, because I think this will be hard to ever do 100% of the work.
This is extremely useful, if wasn't for this feature, I think I wouldn't be using XcodeGen today, because I have a project with a framework and multiple targets, and without experience creating those project.yml files, would be hard.
But this did all the hard work, I just tweaked a bit to work.
@yonaskolb what do you think about adding this as an experimental feature, and marking it as experimental in the docs?
This would also open this PR for external contributions, on the edge cases.

ryohey:
@Satoru-PriChan I cannot reproduce the error with .lproj directory located in a nested directory. Could you give me a minimum project to reproduce it?

ryohey:
There are still some issues such as redundant spec files being generated and, as already reported, errors in some situations, but after seeing it used in various places over the past year or so, I think we have created a feature that is somewhat useful, assuming the user fixes it manually.
I've also added unit tests, although they don't cover much, so I think other contributors will be able to continuously improve this feature.
As @4brunu suggested, I think it might help some people if it is merged as an experimental feature.
Then the only work left to do would be to modify the CLI, e.g. to make the argument a long name like `experimental-migrate`, and to print a warning on stdout that the spec needs to be modified manually.
@yonaskolb What do you think?

yonaskolb:
Yeah I'm open to releasing this in an experimental state.

I've just added the generated specs for our fixtures to the tests, so we can check check progress more easily. At the moment those fail when parsing the generated files back into a Project.swift. I'd like to get at least that fixed. @ryohey seems there are a lot of `null` values generated which I think is causing the issue, if we could resolve that

taglia3:
@yonaskolb I pulled the spec-generation branch, runned the xcodegen schema with launch argument "--migrate" but I encounter an error:

```
Usage: xcodegen [options]

Generate an Xcode project from a spec

Options:
      --cache-path <value>      Where the cache file will be loaded from and save to. Defaults to ~/.xcodegen/cache/{SPEC_PATH_HASH}
  -h, --help                    Show help information
  -n, --no-env                  Disable environment variable expansions
      --only-plists             Generate only plist files
  -p, --project <value>         The path to the directory where the project should be generated. Defaults to the directory the spec is in. The filename is defined in the project spec
  -r, --project-root <value>    The path to the project root directory. Defaults to the directory containing the project spec.
  -q, --quiet                   Suppress all informational and success output
  -s, --spec <value>            The path to the project spec file. Defaults to project.yml
  -c, --use-cache               Use a cache for the xcodegen spec. This will prevent unnecessarily generating the project if nothing has changed

Error: unrecognized option '--migrate'

Program ended with exit code: 1
```

Where am I wrong?

user512:
@taglia3 I think it's `xcodegen migrate --project Project.xcodeproj --spec project.yml`

stackotter:
What still needs to be done to get this to a state where it could be included as an experimental feature? I would be happy to help out with some of the remaining todos depending on their nature.

miku1958:
Hi, when I tried to use migrate to convert a very old project of my company (20 years old, pbxproj has 3.8MB), I found some very unexpected bugs:

1. If a folder has disappeared for historical reasons, but pbxproj still has records about it, it will crash because the folder does not exist
2. Performance issues, after I fixed issue 1, it ran for half an hour and used 6GB of RAM but it wasn't finished, so I added cache, and the effect was that it took less than 5 minutes and about 100MB of RAM to run
3. The conversion does not take into account the relative paths and absolute paths that are missing. 
   My project has a large number of folders without correctly mapped paths for historical reasons, and the files in the folders are directly mapped relative to the root directory (which is the problem I want to solve with xcodegen), so using the converted yml to "xcodegen generate", the generated xcodeproj skips all the folders that are not mapped correctly, and shows all the folders that are mapped correctly in the root directory

ssup1:
Could you explain how can I test this feature on my project?

skofgar:
I think the command [has been changed to](https://github.com/yonaskolb/XcodeGen/pull/735/commits/8c8894cbf6bf75466e0d14cff4671f5d0f116bef#diff-3b872fc3b4a5189112ab71a017987648695a2e8e7e28c17c7b5dafe9b7b5a8e1):
```
xcodegen migrate --spec project.yml Project.xcodeproj 
```

khoogheem:
I am seeing the same issue with not being able to find the `lproj` files.. 

They are nested:
```
App/
   project.xcodeproj
   AppCode/
     Localization/
         xx.lproj/
         zz.lproj/
```

lemon-li:
any progress ?

2jumper3:
You are best, everyone who working on this.

victorwads:
Hey, what happens here? i really need this feature, i dont understand to much of xcode parms, like most ios/macos devs.

i'm tring to improve my project https://github.com/victorwads/UptimeLogger to use xcodegen, but don't even know how to start confgiuring my project.

Anyone?

2jumper3:
> Hey, what happens here? i really need this feature, i dont understand to much of xcode parms, like most ios/macos devs.
> 
> i'm tring to improve my project https://github.com/victorwads/UptimeLogger to use xcodegen, but don't even know how to start confgiuring my project.
> 
> Anyone?

Hi Victor. It works =) u can write me in telegram @okeygo, I'll try to help you =)

victorwads:
> @okeygo

Telegram was blocked on my country by judicial decision TODAY!!!

I spend like 12hours nonstop converting my project. but there are some thing i did not solve.
Can u help on another platform? =D

Abay99:
Does anyone try this out on Apple M1 Pro?

victorwads:
> Does anyone try this out on Apple M1 Pro?
@Abay99 

Eh, i tryed. But, never got working.
I used chat GPT to explain me step by step, i share with it some text parts of my .project, and with like some 3 or 4 hours i could finish converting everything.

https://github.com/victorwads/UptimeLogger/blob/main/project.yml

Abay99:
> 

It worked fine with Apple M1 Pro, but I guess not fully converted it.
Will try to use chatgpt to finish it on 100%.

aemreb:
I'm having issues with the installation, where it can't seem to find SettingPresets:

`cp -R /SettingPresets /usr/local/share/xcodegen/SettingPresets
cp: /SettingPresets: No such file or directory
make: *** [install] Error 1
`

What could be the reason for this?

4brunu:
You need to build this branch from source and use that binary

murad1981:
> You need to build this branch from source and use that binary

Thanks, but the `Spec Generation` branch is outdated, with the last commit being four years ago. I believe versions of Xcode later than the commit on this branch have included many new settings entries in the `.xcodeproj`file.

PR #737, opened 2019-12-16
Files sorted by type

yukkobay:
## Motivation
I want to sort files sort by type as we can with Xcode.

## Description
I added code to sort by type before sorting by name.
We should consider adding as optional. Do you have any ideas?

### DIFF
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -415,7 +415,9 @@ public class PBXProjGenerator {
                 if sortOrder1 != sortOrder2 {
                     return sortOrder1 < sortOrder2
                 } else {
-                    if child1.nameOrPath != child2.nameOrPath {
+                    if child1.extension != child2.extension {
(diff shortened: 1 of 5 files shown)


toshi0383:
fyi you can use other script to do it.
https://github.com/toshi0383/scripts/blob/master/xcode/sort-Xcode-project-file

Maybe we can consider plugin feature to apply whatever modification to the generated output by XcodeGen, before actually write to the disk.(maybe in the future)

toshi0383:
BTW I’m thumb-up to this change. Nice work @yukkobay 👍

yonaskolb:
Thanks for the PR @yukkobay. I'm not comfortable making this the default, but we could add a `SpecOptions.SortType` enum with `name` and `type` cases with the default being `name`

yonaskolb:
@toshi0383 and yes, agree that having some way to modify the XcodeProj in Swift, before it is generated would be great. If you have any good ideas around how to do this, let me know

yukkobay:
@yonaskolb Thank you for reviewing!
I agree with you. I'll try it! 💪

PR #739, opened 2019-12-17
Add option to generate only plist files

namolnad:
At Instacart we use XcodeGen in conjunction with Bazel. For one of our CI testing steps, we need only to generate the Plist files, and not the project in its entirety. This adds the option to exit early from the project generation command, after only generating the plist files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 #### Added
 - Support for language and region settings on a target basis [#728](https://github.com/yonaskolb/XcodeGen/pull/728) @FranzBusch
+- Added option to generate only Info.plist files with `--only-plists` [#739](https://github.com/yonaskolb/XcodeGen/pull/739) @namolnad
 
 #### Fixed
 - Fixed resolving a relative path for `projectReference.path` [#740](https://github.com/yonaskolb/XcodeGen/pull/740) @kateinoigakukun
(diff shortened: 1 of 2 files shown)


yonaskolb:
Interesting. Is not having this a performance issue for you? Theoretically you could generate a project with a name like `Temp.xcodeproj` and then run the following in your script:
```
xcodegen generate
rm -r Temp.xcodeproj
```

I can see the advantage of having this option though for simplicity

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 #### Added
 - Support for language and region settings on a target basis [#728](https://github.com/yonaskolb/XcodeGen/pull/728) @FranzBusch
+- Added option to generate only Info.plist files [#739](https://github.com/yonaskolb/XcodeGen/pull/739) @namolnad
```
```suggestion
- Added option to generate only Info.plist files with `--only-plists` [#739](https://github.com/yonaskolb/XcodeGen/pull/739) @namolnad
```

namolnad:
@yonaskolb, it would be completely a convenience/simplicity thing for us. I just found myself thinking it would be a nice feature to have so thought I'd check on what it would take to add. (Turned out to be remarkably simple)

PS XcodeGen has been awesome and saving us so many headaches since we adopted it — so thanks for all your work on it!

yonaskolb:
Cool, happy to add it. One thing though is that the SwiftCLI has been upgraded in the meantime so you’ll have to rebase and use the new property wrapper for the Flag

namolnad:
sounds good, thanks for the heads up @yonaskolb! (After a quick bad rebase,) I think we should be g2g now. Thanks again for taking a look at this!

Issue #729, opened 2019-11-26
projectReferences path is not resolved correctly


PycKamil:
Looks like there is some kind of issue in resolution of path in `projectReferences`. When I would like to reference a project is upper directory like `path: ../../Test/Test.xcodeproj` it will cause error message for project not found. Changing it to incorrect path `path: Test/Test.xcodeproj` will work but that create a wrong reference. Quick workaround for this issue is to create symbolic link to actual project in current folder but it would be nice to fix it properly.

yonaskolb:
Are you running XcodeGen with any arguments, like --spec or --project, and are you using any includes?

PycKamil:
Yes I am using `--spec` argument

yonaskolb:
Yeah, though so. I don’t think the project reference paths are being set relative to the spec base path.
As a workaround for now you should be able to run XcodeGen in the same directory as your spec

yonaskolb:
@kateinoigakukun

kateinoigakukun:
@PycKamil Sorry for late reply, could you try this branch https://github.com/yonaskolb/XcodeGen/pull/740?

PR #740, opened 2019-12-18
Replace projectReference.path as relativePath

kateinoigakukun:
Fix https://github.com/yonaskolb/XcodeGen/issues/729


yonaskolb:
There is a test somewhere that validates that all the relevant relative paths are updated. Could you add this case to it as well?

Issue #746, opened 2019-12-27
Add Organization Name in spec file


Christian-Seiler:
I would like to have the organization name added in my projects. (as for copyright notice)

yonaskolb:
@Christian-Seiler I left a comment in your PR. This is possible by using project attributes

```yml
attributes:
  ORGANIZATIONNAME: My Org
```

yonaskolb:
Closing this for now. Let me know if you have any issues

PR #747, opened 2019-12-27
#746 Add Organization Name in spec file

Christian-Seiler:


yonaskolb:
Thanks for the PR @Christian-Seiler. This is already possible to do though by setting the project attributes
```yml
attributes:
  ORGANIZATIONNAME: My Org
```
I'd rather not add specific properties for every single attribute we want to support, unless it interacts with other features.

Let me know if that works out for you, and then I think we could close this PR.

PR #751, opened 2020-01-07
Use full path for XcodeProj

PycKamil:
This solves issue with relative project path reported with #729. It was partially fixed with #740 project reference is loaded correctly now, but when projects target is for example used in scheme, `XcodeProj` fails to read it, since the `../Project.xcodeproject` path is not recognised. This PR fix it by adding a base project path before passing it to it `XcodeProj`.


yonaskolb:
Thank you @PycKamil! Could you add a changelog entry as well

PycKamil:
Sure thing @yonaskolb, entry added 👍

PR #752, opened 2020-01-15
Prefer default Debug and Release configs

john-flanagan:
The default schemes generated for a target is using the `suitableConfig` helper function to prefer configs named "Debug" or "Release" over the alphabetically first scheme of the matching type:

https://github.com/yonaskolb/XcodeGen/blob/480477a7e8863e1585596c417cd6ce34513b9ef2/Sources/XcodeGenKit/SchemeGenerator.swift#L57-L58

but `generateScheme(_: Scheme)` was using the computed `default<ConfigType>Config` properties which only took into account the sorting order for top level schemes.

https://github.com/yonaskolb/XcodeGen/blob/480477a7e8863e1585596c417cd6ce34513b9ef2/Sources/XcodeGenKit/SchemeGenerator.swift#L244-L250

This change uses the same `suitableConfig` function to back the `default<ConfigType>Config` properties so that top level schemes and target schemes will have the same default config behavior.


yonaskolb:
Great, thanks @john-flanagan! Could you please add a changelog entry as well

PR #759, opened 2020-01-23
Add pre and post-gen commands

### LINKED ISSUE #479: Add postGen hook
This would add a hook for a script to be run after generation. Critically this won't run if caching is on and the project isn't generated. 
This solves the issue of running `pod install` only if the project has changed as discussed here https://github.com/yonaskolb/XcodeGen/pull/412.

Example:
```yml
options:
  postGenScript: pod install
targets:
  ...
```

yonaskolb:
Resolves #479 
These are useful in combination with `--use-cache` and will then only be run if the project is actually generated.
This is especially useful for running `pod install`


giginet:
Awesome feature ✨ 

I concern about there are use cases to execute external shell scripts like Post Build Script phase.

```yaml
options:
  preGenCommand:
    path: "./my-complex-bootstrap-script.sh"
  postGenCommand:
    script: "pod install"
```

yonaskolb:
@giginet you can run scripts using the existing syntax 👍
```
options:
  preGenCommand: ./my-complex-bootstrap-script.sh
```

PR #772, opened 2020-02-04
ignore vim swp file

toshi0383:
`swp` file was unexpectedly added to pbxproj while editing some files.

Does this change make sense?
Otherwise it's okay to me to just exclude them in project spec.


toshi0383:
> Otherwise it's okay to me to just exclude them in project spec

`"**/*.swp"` in excludes sometimes didn't match swp files..🙄

toshi0383:
Closing due to inactivity.👋🏻

PR #787, opened 2020-02-25
Use defaultConfigurationName for every target in a project

ken0nek:
Currently `defaultConfigurationName` of a target is empty even if we set `defaultConfigurationName` in project settings. This is the same behavior as when we change a default configuration in Xcode.

## As is

```
Project A -  defaultConfigurationName: "Release"
Target 1 -  defaultConfigurationName: ""
Target 2 -  defaultConfigurationName: ""
```

## To be

```
Project A -  defaultConfigurationName: "Release"
Target 1 -  defaultConfigurationName: "Release"
Target 2 -  defaultConfigurationName: "Release"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Next Version
 
+#### Fixed
+- Set `defaultConfigurationName` for every target which is defined in a project. [#787](https://github.com/yonaskolb/XcodeGen/pull/787)
+
 ## 2.13.1
 
 #### Fixed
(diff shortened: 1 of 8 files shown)


kateinoigakukun:
Note: This `defaultConfigurationName` is used when building targets from `xcodebuild` without specific config.

yonaskolb:
Does this need to be on every target, or could it just be set once in the project build settings?

ken0nek:
> Does this need to be on every target, or could it just be set once in the project build settings?

In Xcode, a config name defined in `Use [CONFIG_NAME] for command-line builds` is used for every target in the project. I followed this behavior. Should I fix this change? 🙏 

Screen Shot 2020-02-24 at 5 14 02 PM

Screen Shot 2020-02-24 at 5 14 07 PM

yonaskolb:
Yep, all good 👍

PR #788, opened 2020-02-25
Fix #784 Swift Package Manager (Embed & Sign) for dynamic packages.

### LINKED ISSUE #784: Dependencies: SPM dynamic dependency embed & sign
Similar to #271 but for Swift Package Manager dynamic dependencies declared like:

```swift
// swift-tools-version:5.1
import PackageDescription

let package = Package(
    name: "Components",
    platforms: [
        .iOS(.v12),
    ],
    products: [
        .library(name: "Components", type: .dynamic, targets: ["Components"]),
    ],
    targets: [
        .target(name: "Components", dependencies: []),
    ],
    swiftLanguageVersions: [.v5]
)
```

My project.yml is like:
```yaml
    type: application
    platform: iOS
    sources: Sources
    dependencies:
      - package: Components
        embed: true
```

I want my dynamic SPM dependency as Embed & Sign for application target:

Captura de pantalla 2020-02-19 a las 16 35 06

To be included here:

Captura de pantalla 2020-02-19 a las 16 35 34

@yonaskolb can you help me with config? Thanks! 😃

alexruperez:
Swift Package Manager dynamic dependencies declared like:

```swift
.library(name: "MyLib", type: .dynamic, targets: ["MyLib"])
```

Are embedded (**Embed & Sign** or **Embed Without Signing** if `codeSign: false`) when you declare `embed: true` in package dependency definition:

Captura de pantalla 2020-02-19 a las 16 35 06

And included here:

Captura de pantalla 2020-02-19 a las 16 35 34

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Next Version
 
+#### Added
+- Support Swift Package Manager (Embed & Sign/Embed Without Signing) for dynamic packages. [#788](https://github.com/yonaskolb/XcodeGen/pull/788)
+
 #### Fixed
 - Set `defaultConfigurationName` for every target which is defined in a project. [#787](https://github.com/yonaskolb/XcodeGen/pull/787)
 
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -821,7 +821,7 @@ public class PBXProjGenerator {
                     dependencies.append(targetDependency)
                 }
                 
-                if embed {
+                if dependency.embed == true {
                     let embedFile = addObject(
                         PBXBuildFile(product: packageDependency,
                                      settings: getEmbedSettings(dependency: dependency, codeSign: dependency.codeSign ?? true))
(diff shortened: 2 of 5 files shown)


JesusMartinAlonso:
It would be great if this PR could be approved soon! Thanks @alexruperez for the contribution! :)

brentleyjones:
@alexruperez Thanks for the fix! Can you update the tests and include the updated test fixtures as well? Also, can you add a Changelog entry?

TotaX:
great work @alexruperez :) thanks for taking time to fix it..

yonaskolb:
Thanks for the PR @alexruperez!
Currently this PR embeds all Swift packages by default which is not I think what we want. You can see this in the test failure. We either need a way to tell if a package library is static or dynamic, or only embed if it’s been defined explicitly in the spec. The current embed property defaults to yes for application targets as can be seen above your change. We can create our own property for swift package dependencies that defaults to false
```swift
 let embed = dependency.embed ?? false
```
Why do you think? I more thorough solution would be to check the framework itself for a good default.

In addition could you please:
- add a changelog entry
- add a package with an embedded dynamic library to the SPM fixture https://github.com/yonaskolb/XcodeGen/blob/master/Tests/Fixtures/SPM/project.yml
- run the tests and check in the updated generated fixtures
- ideally we’d have a unit test for this too if you have the time

Thanks!

alexruperez:
Thanks for the feedback @yonaskolb. Added changelog entry and fixtures updates.

Also SPM dependencies are embedded only when `embed: true` is declared in the `project.yml`

Does this fit for you?

yonaskolb:
Released in 2.14.0

PR #792, opened 2020-02-26
Set `TEST_TARGET_NAME` only when a project has UITest bundle

ken0nek:
`TEST_TARGET_NAME` is only for UITest and I think we don't have to set this if we just have UnitTest.

ref: https://github.com/yonaskolb/XcodeGen/pull/452


brentleyjones:
Thank you @ken0nek :smile:.

Issue #721, opened 2019-11-20
Can't Add Local Swift Package as Dependency


jsorge:
Using Xcodegen v2.10.1 and adding a local Swift package to my project, and I want it to be a dependency for one of my targets. The following is not working for me:

```
name: MyApp
localPackages:
  - Modules/Utilities
packages:
  Pathos:
    from: 0.2.0
    url: https://github.com/dduan/Pathos.git
targets:
  MyApp:
    type: application
    platform: iOS
    sources:
      - MyApp/Sources
      - MyApp/Resources
    dependencies:
      - package: Utilities
    configFiles:
      Debug: xcconfigs/BindleWriter-Debug.xcconfig
      Release: xcconfigs/BindleWriter-Release.xcconfig
```

And the error I get is `Spec validation error: Target "MyApp" has an invalid package dependency "Utilities"`. If I alter the dependency on Utilities to be on Pathos instead (which is a remote package) then the project generates just fine.

Am I holding it wrong?

yonaskolb:
Hi @jsorge, currently local packages are only supported when they mirror remote packages. There is mention of that buried here https://github.com/yonaskolb/XcodeGen/blob/master/Docs/Usage.md

Additional work is required to support completely local packages

jsorge:
Ah that makes sense. I hadn't seen that note before. Thanks for letting me know.

AV8R-:
Hi @yonaskolb, what kind of additional work is required to make this feature work?
Actually I tried to remove any package validation form SpecValidation.swift (that was on line 168) and removed `guard` statement in PBXProjGenerator.swift on line 630.

And it works for me. Except I don't have any remote package validation for now.
What should I do additionally to make a pull request for this feature?

WilliamIzzo83:
Hi everybody

I'm kinda interested too. Are there any news about this?

freddi-kit:
I made a pull request and it would help this issue => https://github.com/yonaskolb/XcodeGen/pull/796

WilliamIzzo83:
Great work guys. Thank you so much! Waiting for a new release :D

freddi-kit:
Hello. v2.15.0 is released and we can use local Swift Package from now by following format.
Thank you all!

```yml
name: GennedProj
...
packages:
  MyLibrary: 
    path: Packages/MyLibrary
targets:
  GennedProj:
    ...
    dependencies:
      - package: MyLibrary
```

ThePredators:
Hello,

[...]

PR #796, opened 2020-03-03
Support Local Swift Package as Dependency

freddi-kit:
# Summary
Add Support Local Swift Package as Dependency by `localPackages`
Fix: https://github.com/yonaskolb/XcodeGen/issues/721

# What Changed?
- Add Support `localPackages:` for Local Swift Package as Dependency (not only mirror remote packages)
 - I added new format of `localPackages:` to make package name clear and we can set it as `package` in `dependencies`
- Fixed resolving path issue of `localPackages:`

So we can use local Swift Package by ...

```yaml
name: GennedProj
...
localPackages:
  MyLibrary: 
    path: Packages/MyLibrary
targets:
  GennedProj:
    ...
    dependencies:
      - package: MyLibrary
```

note) we can also use old format of `localPackages:` for `[String]` like following. But it is only supporting mirror one.

```yaml
name: GennedProj
...
localPackages:
  - Packages/MyLibrary
targets:
  GennedProj:
    ...
```

# Motivation
Currently, `localPackages` is supporting Dependency only the packages which are mirror remote packages. 
However, as you can see, there are some voices to desire supporting local one in https://github.com/yonaskolb/XcodeGen/issues/721, I also want it.

So I propose this PR to support Local Swift Package as Dependency.

## New format of `localPackages`
As I mentioned, I create new format of `localPackages`.
At first, I thought it is okay to keep old format with supporting local package.

```yaml
name: GennedProj
...
localPackages:
  - Packages/MyLibrary
targets:
  GennedProj:
    ...
    dependencies:
      - package: Packages/MyLibrary
```

However, this has a problem after generating because .xcodeproj has a problem when package name has "/".
And I think it should be better to specify package name to use in `dependencies`, like `packages`. 

So I made format similar with `packages`, like following yaml code.

```yaml
localPackages:
  MyLibrary1: 
    path: Packages/MyLibrary1
  MyLibrary2: 
    path: Packages/MyLibrary2
```

And we can specify them in `dependencies` without problem 🎉 and they are linked as framework automatically.

```yaml
targets:
  GennedProj:
    dependencies:
      - package: MyLibrary1
      - package: MyLibrary2
```

# For Testing
You can test this Pull Request by it -> https://github.com/freddi-kit/TestRepoForXcodeGenPR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Next Version
 
 #### Added
-- Support local Swift pacakge by `localPackages`. [#712](https://github.com/yonaskolb/XcodeGen/pull/796) @freddi-kit
+- Support local Swift pacakge by `localPackages`. [#796](https://github.com/yonaskolb/XcodeGen/pull/796) @freddi-kit
 
 ## 2.14.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Next Version
 
+#### Added
+- Support Swift Package Manager (Embed & Sign/Embed Without Signing) for dynamic packages. [#788](https://github.com/yonaskolb/XcodeGen/pull/788)
+
 #### Fixed
 - Revert "Add Base to known regions even if one doesn't exist" [#791](https://github.com/yonaskolb/XcodeGen/pull/792) @bryansum
 - Set `defaultConfigurationName` for every target which is defined in a project. [#787](https://github.com/yonaskolb/XcodeGen/pull/787) @ken0nek
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,14 +2,18 @@
 
 ## Next Version
 
+## 2.14.0
+
 #### Added
-- Support Swift Package Manager (Embed & Sign/Embed Without Signing) for dynamic packages. [#788](https://github.com/yonaskolb/XcodeGen/pull/788)
+- Add ability to embed and code sign Swift package dependencies with dynamic products. [#788](https://github.com/yonaskolb/XcodeGen/pull/788) @alexruperez
 
 #### Fixed
 - Revert "Add Base to known regions even if one doesn't exist" [#791](https://github.com/yonaskolb/XcodeGen/pull/792) @bryansum
 - Set `defaultConfigurationName` for every target which is defined in a project. [#787](https://github.com/yonaskolb/XcodeGen/pull/787) @ken0nek
 - Set `TEST_TARGET_NAME` only when a project has UITest bundle. [#792](https://github.com/yonaskolb/XcodeGen/pull/792) @ken0nek
 - Set xcodeproj path in project.xcworkspace/contents.xcworkspacedata [#793](https://github.com/yonaskolb/XcodeGen/pull/793) @ken0nek
+
+[Commits](https://github.com/yonaskolb/XcodeGen/compare/2.13.1...2.14.0)
 
 ## 2.13.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 #### Added
 - Support local Swift pacakge by `localPackages`. [#796](https://github.com/yonaskolb/XcodeGen/pull/796) @freddi-kit
 
+#### Fixed
+- Fixed resolving path to local Swift pacakge by `localPackages`. [#796](https://github.com/yonaskolb/XcodeGen/pull/796) @freddi-kit
+
 ## 2.14.0
 
 #### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,10 +3,10 @@
 ## Next Version
 
 #### Added
-- Support local Swift pacakge by `localPackages`. [#796](https://github.com/yonaskolb/XcodeGen/pull/796) @freddi-kit
+- Add support non-mirrored local Swift Packages in `localPackages`. [#796](https://github.com/yonaskolb/XcodeGen/pull/796) @freddi-kit
 
 #### Fixed
-- Fixed resolving path to local Swift pacakge by `localPackages`. [#796](https://github.com/yonaskolb/XcodeGen/pull/796) @freddi-kit
+- Fixed resolving path to local Swift Packages [#796](https://github.com/yonaskolb/XcodeGen/pull/796) @freddi-kit
 
 ## 2.14.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 #### Fixed
 - Fixed resolving path to local Swift Packages [#796](https://github.com/yonaskolb/XcodeGen/pull/796) @freddi-kit
+- Added ability to stop on every main thread checker issue on Run schemes and TargetSchemes [#799](https://github.com/yonaskolb/XcodeGen/pull/799) @ionutivan
 
 ## 2.14.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Next Version
 
+#### Added
+- Support local Swift pacakge by `localPackages`. [#712](https://github.com/yonaskolb/XcodeGen/pull/796) @freddi-kit
+
 ## 2.14.0
 
 #### Added
--- a/Docs/ProjectSpec.md
+++ b/Docs/ProjectSpec.md
@@ -50,7 +50,7 @@ You can also use environment variables in your configuration file, by using `${S
 - [ ] **schemes**: **[Scheme](#scheme)** - A list of schemes by name. This allows more control over what is found in [Target Scheme](#target-scheme)
 - [ ] **targetTemplates**: **[String: [Target Template](#target-template)]** - a list of targets that can be used as templates for actual targets which reference them via a `template` property. They can be used to extract common target settings. Works great in combination with `include`.
 - [ ] **packages**: **[String: [Swift Package](#swift-package)]** - a map of Swift packages by name
-- [ ] **localPackages**: **[String: [Local Swift Package](#local-swift-package)]** -  a map of local Swift packages by name. The paths must be directories with a `Package.swift` file in them. If same name Swift package is listed in `packages`, this override the same name package in  `packages` with a local version for development purposes. We can also use by path list ( `[String]` ). However, local packages that don't mirror remote packages aren't able to be linked to, by `[String]` format.
+- [ ] **localPackages**: **[String: [Local Swift Package](#local-swift-package)]** -  a map of local Swift packages by name. The paths must be directories with a `Package.swift` file in them. If same name remote repo is listed in `packages`, remote repo will be overridden to a local version in `localPackages` for development purposes. We can also use by path list ( `[String]` ). However, local packages that don't mirror remote packages aren't able to be linked to, by `[String]` format.
 - [ ] **projectReferences**: **[String: [Project Reference](#project-reference)]** - a map of project references by name
 
 ### Include
--- a/Docs/ProjectSpec.md
+++ b/Docs/ProjectSpec.md
@@ -598,6 +598,7 @@ This is a convenience used to automatically generate schemes for a target based 
 - [ ] **testTargets**: **[[Test Target](#test-target)]** - a list of test targets that should be included in the scheme. These will be added to the build targets and the test entries. Each entry can either be a simple string, or a [Test Target](#test-target)
 - [ ] **gatherCoverageData**: **Bool** - a boolean that indicates if this scheme should gather coverage data. This defaults to false
 - [ ] **disableMainThreadChecker**: **Bool** - a boolean that indicates if this scheme should disable disable the Main Thread Checker. This defaults to false
+- [ ] **stopOnEveryMainThreadCheckerIssue**: **Bool** - a boolean that indicates if this scheme should stop at every Main Thread Checker issue. This defaults to false
 - [ ] **language**: **String** - a String that indicates the language used for running and testing. This defaults to nil
 - [ ] **region**: **String** - a String that indicates the region used for running and testing. This defaults to nil
 - [ ] **commandLineArguments**: **[String:Bool]** - a dictionary from the argument name (`String`) to if it is enabled (`Bool`). These arguments will be added to the Test, Profile and Run scheme actions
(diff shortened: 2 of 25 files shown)


freddi-kit:
I also updated the document, ready for review

Sources/ProjectSpec/Project.swift
```diff
@@ -188,7 +188,17 @@ extension Project {
         } else {
             packages = [:]
         }
-        localPackages = jsonDictionary.json(atKeyPath: "localPackages") ?? []
+        if let localPackages: [String: LocalSwiftPackage] = jsonDictionary.json(atKeyPath: "localPackages") {
+            self.localPackages = localPackages
+        } else if let localPackages: [String] = jsonDictionary.json(atKeyPath: "localPackages") {
+            self.localPackages = localPackages.reduce([String: LocalSwiftPackage]()) { result, path in
```
giginet: I prefer to use `reduce(into:)` 👍 

```swift
localPackages.reduce(into: [String: LocalSwiftPackage]()) { $0[$1] = LocalSwiftPackage(path: $1) }
```

Sources/ProjectSpec/SpecValidation.swift
```diff
@@ -53,9 +53,9 @@ extension Project {
             }
         }
 
-        for package in localPackages {
-            if !(basePath + Path(package).normalize()).exists {
-                errors.append(.invalidLocalPackage(package))
+        for (name, package) in localPackages {
+            if !(basePath + Path(package.path).normalize()).exists {
```
giginet: Maybe you can use `relativePath` extension instead.

https://github.com/yonaskolb/XcodeGen/blob/e9239f83bd3725c462d42219ba45761a88ffe1e1/Sources/Core/PathExtensions.swift#L22

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -56,9 +56,8 @@ public class PBXProjGenerator {
             try sourceGenerator.getFileGroups(path: group)
         }
 
-        let localPackages = Set(project.localPackages)
-        for package in localPackages {
-            let path = project.basePath + Path(package).normalize()
+        for (_, package) in project.localPackages {
+            let path = project.basePath + Path(package.path).normalize()
```
giginet: relativePath (same above)

Tests/XcodeGenKitTests/ProjectGeneratorTests.swift
```diff
@@ -991,7 +991,7 @@ class ProjectGeneratorTests: XCTestCase {
                 let project = Project(name: "test", targets: [app], packages: [
                     "XcodeGen": SwiftPackage(url: "http://github.com/yonaskolb/XcodeGen", versionRequirement: .branch("master")),
                     "Codability": SwiftPackage(url: "http://github.com/yonaskolb/Codability", versionRequirement: .exact("1.0.0")),
-                ], localPackages: ["../XcodeGen"], options: .init(localPackagesGroup: "MyPackages"))
+                ], localPackages: ["../XcodeGen" : LocalSwiftPackage(path: "../XcodeGen")], options: .init(localPackagesGroup: "MyPackages"))
```
giginet: Can we use absolute paths?

freddi-kit: No, I found that we will meet not linking local package issue to project when using absolute paths.
This looks old bug and I fixed in latest commit, so please check it after pushed

Thank you for pointing out, I did not notice it! 👍

freddi-kit:
@giginet Thank you for reviewing. 

> For example, we have to care about -p option of generate command.

Yeah, I tried `-p` and there is a not linking package to project issue. It looks `localPackages` already have this issue in 2.14.0.
So I fixed in my latest issue by using some logic in `resolveGroupPath`.

I will push fix after some checking. Thank you for sharing  👍

Sources/ProjectSpec/SpecValidation.swift
freddi-kit: I think it is intended to no use `relativePath` here. 
It is because this part is validating the specified file is existing or not by using an absolute path, so other code of this function is also using absolute path.

yonaskolb:
We could potentially use a short form of package definitions that is just a map of name to path. Do we envisage setting other properties on a local package other than path? Perhaps it's probably best to keep the full form in case we do, as you've done already.
```
localPackages:
  MyLibrary1: 
    path: Packages/MyLibrary1
```
```
localPackages:
  MyLibrary1: Packages/MyLibrary1
```
Thoughts?

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -56,10 +56,8 @@ public class PBXProjGenerator {
             try sourceGenerator.getFileGroups(path: group)
         }
 
-        let localPackages = Set(project.localPackages)
-        for package in localPackages {
-            let path = project.basePath + Path(package).normalize()
-            try sourceGenerator.createLocalPackage(path: path)
+        for (_, package) in project.localPackages {
+            try sourceGenerator.createLocalPackage(path: Path(package.path))
```
yonaskolb: We should sort the local packages by name here before generating them, so the output is deterministic

freddi-kit:
@yonaskolb Thank you for the suggestion

I'm not envisaging setting other properties, but I believe this style has extensibility for the feature in the future. 
For example, `pakcages:` has properties (e.g. `from:`) but `localPackages:` can also have and develop easily if we use  full form.

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## Next Version
 
+#### Added
+- Support local Swift pacakge by `localPackages`. [#796](https://github.com/yonaskolb/XcodeGen/pull/796) @freddi-kit
```
```suggestion
- Add support non-mirrored local Swift Packages in `localPackages`. [#796](https://github.com/yonaskolb/XcodeGen/pull/796) @freddi-kit
```

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
...
+
+#### Fixed
+- Fixed resolving path to local Swift pacakge by `localPackages`. [#796](https://github.com/yonaskolb/XcodeGen/pull/796) @freddi-kit
```
```suggestion
- Fixed resolving path to local Swift Packages [#796](https://github.com/yonaskolb/XcodeGen/pull/796) @freddi-kit
```

yonaskolb:
Thanks the for the great PR @freddi-kit! I’ll review properly tonight 👍

Sources/XcodeGenKit/SourceGenerator.swift
```diff
@@ -60,12 +60,18 @@ class SourceGenerator {
             localPackageGroup = addObject(PBXGroup(sourceTree: .sourceRoot, name: groupName))
             rootGroups.insert(localPackageGroup!)
         }
+        
+        let absolutePath = path.isAbsolute ? path : project.basePath + path.normalize()
```
yonaskolb: I believe PathKit automatically does this automatically already so if you add an absolute path to a relative path, it just evaluates to the absolute path

Tests/Fixtures/SPM/SPM.xcodeproj/project.pbxproj
```diff
@@ -49,7 +49,7 @@
 		4E22B8BCC18A29EFE1DE3BE4 /* Assets.xcassets */ = {isa = PBXFileReference; lastKnownFileType = folder.assetcatalog; path = Assets.xcassets; sourceTree = "<group>"; };
 		61C17B77601A9D1B7895AB42 /* StaticLibrary.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = StaticLibrary.swift; sourceTree = "<group>"; };
 		A9601593D0AD02931266A4E5 /* App.xctestplan */ = {isa = PBXFileReference; path = App.xctestplan; sourceTree = "<group>"; };
-		C1DE9A872F470EAA65B9B0B0 /* XcodeGen */ = {isa = PBXFileReference; lastKnownFileType = folder; name = XcodeGen; path = ../../..; sourceTree = SOURCE_ROOT; };
+		B71D3227C09EFC935723F45B /* .. */ = {isa = PBXFileReference; lastKnownFileType = folder; name = ..; path = ../../..; sourceTree = "<group>"; };
```
yonaskolb: Looks like the path maybe needs a `normalise` to get an actual name?

Sources/XcodeGenKit/SourceGenerator.swift
```diff
@@ -60,12 +60,18 @@ class SourceGenerator {
...
+
+        // Get local pakcage's relative path from generating project's root
+        let fileReferencePath = try? absolutePath.relativePath(from: projectDirectory ?? project.basePath).string
+        
         let fileReference = addObject(
             PBXFileReference(
-                sourceTree: .sourceRoot,
+                sourceTree: .group,
```
yonaskolb: Was this required?

Docs/ProjectSpec.md
```diff
@@ -49,7 +50,7 @@ You can also use environment variables in your configuration file, by using `${S
...
 - [ ] **targetTemplates**: **[String: [Target Template](#target-template)]** - a list of targets that can be used as templates for actual targets which reference them via a `template` property. They can be used to extract common target settings. Works great in combination with `include`.
 - [ ] **packages**: **[String: [Swift Package](#swift-package)]** - a map of Swift packages by name
-- [ ] **localPackages**: **[String]** - A list of paths to local Swift Packages. The paths must be directories with a `Package.swift` file in them. This is used to override `packages` with a local version for development purposes. For now local packages that don't mirror remote packages aren't able to be linked to
+- [ ] **localPackages**: **[String: [Local Swift Package](#local-swift-package)]** -  a map of local Swift packages by name. The paths must be directories with a `Package.swift` file in them. If same name remote repo is listed in `packages`, remote repo will be overridden to a local version in `localPackages` for development purposes. We can also use by path list ( `[String]` ). However, local packages that don't mirror remote packages aren't able to be linked to, by `[String]` format.
```
yonaskolb: We can probably leave out the comment about `[String]` being possible, to not confuse people

Docs/Usage.md
```diff
@@ -178,16 +178,26 @@ If you want to check in the `Package.resolved` file so that everyone is on the s
...
 
-You can also include local Swift Packages by referencing them by paths in `localPackages`. When these have the same name as `packages` they will be used instead of the remote repos. This is useful for local development.
+You can also include local Swift Packages by referencing them by paths in `localPackages` and you can use them as `dependencies` same with `packages`. When these local packages in `localPackages` have the same name as remote repos in `packages`, they will be used instead of the remote repos. This is useful for local development.
 
 ```yml
+packages:
+  Yams:
+    url: https://github.com/jpsim/Yams
+    from: 2.0.0
 localPackages:
```
yonaskolb: What do you think of just having a single `packages` dictionary, and you can specify a `path` on a Package. A Package can then be an enum with either a `.remote(url: String, version: VersionRequirement)` or a `.local(path: String)`. If a definition had a path it would be local, otherwise would parse a remote package. The only downside is that you can't have a remote and a local at the same time for mirroring, but I'm not sure if that's important as you can just use a local package? Thoughts?

Tests/Fixtures/SPM/SPM.xcodeproj/project.pbxproj
freddi-kit: Oh sorry, I thought too so I fixed at my lastest commit

Docs/ProjectSpec.md
freddi-kit: Or how about adding deprecated tag?

Sources/XcodeGenKit/SourceGenerator.swift
freddi-kit: Oh I found this change is not needed, I'll fix 👍

Docs/ProjectSpec.md
freddi-kit:
```suggestion
- [ ] **localPackages**: **[String: [Local Swift Package](#local-swift-package)]** -  a map of local Swift packages by name. The paths must be directories with a `Package.swift` file in them. If same name remote repo is listed in `packages`, remote repo will be overridden to a local version in `localPackages` for development purposes. **Deprecated**: We can also use by path list ( `[String]` ). However, local packages that don't mirror remote packages aren't able to be linked to, by `[String]` format.
```

Docs/Usage.md
freddi-kit: I agree 👍👍👍 . It is great to make the format more consistent if it is okay to only use `packages`. I really want to do so. 

However, IMO it would be better to do it later. I also not sure of the importance of the existing local package feature. However, it would affect backward compatibility so I want to set as deprecated at first before deleting.

freddi-kit: Or I have already prepared an implementation on local so can I make another PR after this merged?

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -799,8 +798,12 @@ public class PBXProjGenerator {
                 }
             // Embedding handled by iterating over `carthageDependencies` below
             case .package(let product):
-                guard let packageReference = packageReferences[dependency.reference] else {
-                    return
```
freddi-kit: I found that this code would have an issue when there is no package reference. This early return will cause the outputting unfinished project file. Currently, This does not occur because current project validation works well, but I think it would be better to make this as `continue` ASAP. => https://github.com/yonaskolb/XcodeGen/pull/796/commits/dd7737354dd00baac77bd19f75b16ff834f11c00

mkkroliks:
@yonaskolb When can we expect this PR to be merged?

Sources/XcodeGenKit/SourceGenerator.swift
```diff
@@ -60,12 +60,18 @@ class SourceGenerator {
             localPackageGroup = addObject(PBXGroup(sourceTree: .sourceRoot, name: groupName))
             rootGroups.insert(localPackageGroup!)
         }
+        
+        let absolutePath = project.basePath + path.normalize()
+
+        // Get local pakcage's relative path from generating project's root
```
yonaskolb:
```suggestion
        // Get the local package's relative path from the project root
```

Tests/XcodeGenKitTests/ProjectGeneratorTests.swift
yonaskolb: The name shouldn't be a path
```suggestion
                ], localPackages: ["XcodeGen" : LocalSwiftPackage(path: "../XcodeGen")], options: .init(localPackagesGroup: "MyPackages"))
```

Docs/ProjectSpec.md
freddi-kit: Okay, let me remove old one 👍

Docs/Usage.md
```diff
@@ -178,16 +178,26 @@ If you want to check in the `Package.resolved` file so that everyone is on the s
...
+    path: ../Yams
+targets:
+  App:
+    dependencies:
+      - package: MyPackage
+      - package: Yams # Yams in local will be used instead of Yams from remote repos
 
-> For now local packages that don't mirror remote packages aren't able to be linked to
+```
+These local packages get put into a `Packages` group in the root of the project by default. This can be changed with `options.localPackagesGroup`. However you can not set it as `dependencies`, and they are only used for overriding `packages`.
```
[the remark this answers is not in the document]
freddi-kit: Ahhh, I'm sorry. I misunderstood how works options.localPackagesGroup. I think I can delete this sentence ... my added sentence is not true.  

At first, I misunderstood as localPackagesGroup will register local package in once by specifying the folder. But it looks localPackagesGroup is for only changing local package directly name.

Docs/Usage.md
freddi-kit: Thank you! I'll make PR ASAP after it merged 👍

Docs/Usage.md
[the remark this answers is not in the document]
freddi-kit: Removed, Thank you!

freddi-kit:
Thank you a lot!! @yonaskolb

mateuszjablonski-slabs:
@yonaskolb are you going to release it immediately, or waiting for some more PRs?

yonaskolb:
Will release once #808 is merged

PR #797, opened 2020-03-03
Allow external dependencies in aggregate targets

evandcoleman:
Building upon #701, this allows referencing external targets in aggregate targets.

### DIFF
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -322,7 +322,17 @@ public class PBXProjGenerator {
             return addObject(buildConfig)
         }
 
-        let dependencies = target.targets.map { generateTargetDependency(from: target.name, to: $0) }
+        let dependencies = try target.targets.map { dependencyReference -> PBXTargetDependency in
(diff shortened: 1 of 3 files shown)


Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -419,30 +429,34 @@ public class PBXProjGenerator {
             )
         )
 
-        guard let productType = targetObject.productType,
-            let buildConfigurations = targetObject.buildConfigurationList?.buildConfigurations,
-            let defaultConfigurationName = targetObject.buildConfigurationList?.defaultConfigurationName,
-            let defaultConfiguration = buildConfigurations.first(where: { $0.name == defaultConfigurationName }) ?? buildConfigurations.first else {
+        if let platform = platform {
```
kateinoigakukun: I think it's better to split up this method into "generate targetDependency" and others instead of allowing nil platform.
I expect that it will make it unnecessary to [unwrap `Optional` forcibly](https://github.com/yonaskolb/XcodeGen/pull/797/files#diff-2f3c621620623f55e7ff42cc32198a10R706).

PR #798, opened 2020-03-04
Use remote-runnable launch actions for Apple Watch schemes

### LINKED ISSUE #736: Unable to install watchOS app
Using the current [TestProject](https://github.com/yonaskolb/XcodeGen/tree/master/Tests/Fixtures/TestProject) I am unable to install and run the watchOS app scheme.

The error message is: "Could not install at this time. No WatchKit 1.0 app bundle found for app com.project.app.watch.extension".

XCode - Version 11.2.1 (11B500)
Latest version of XcodeGen

daltonclaybrook:
Resolves #736 

This PR fixes an issue that prevents a Watch app from being launched using the generated scheme. The `BuildableProductRunnable` launch action has been replaced with a `RemoteRunnable` action, and the host app has been added as a `BuildActionEntry`. This more closely matches the scheme that Xcode generates when creating a new Apple Watch app target and seems to work correctly when building/running the target from Xcode.

### DIFF
--- a/Tests/XcodeGenKitTests/SchemeGeneratorTests.swift
+++ b/Tests/XcodeGenKitTests/SchemeGeneratorTests.swift
@@ -336,6 +336,48 @@ class SchemeGeneratorTests: XCTestCase {
                 try expect(buildableReference?.blueprintName) == "ExternalTarget"
                 try expect(buildableReference?.referencedContainer) == "container:\(externalProject.string)"
             }
+
+            $0.it("generates scheme with remote runnable for watch") {
+                let xcscheme = try self.makeWatchScheme(appType: .watch2App, extensionType: .watch2Extension)
+                try expect(xcscheme.launchAction?.runnable).beOfType(XCScheme.RemoteRunnable.self)
+            }
+
+            $0.it("generates scheme with host target build action for watch") {
+                let xcscheme = try self.makeWatchScheme(appType: .watch2App, extensionType: .watch2Extension)
+                let buildEntries = xcscheme.buildAction?.buildActionEntries ?? []
+                try expect(buildEntries.count) == 2
+                try expect(buildEntries.first?.buildableReference.blueprintName) == "WatchApp"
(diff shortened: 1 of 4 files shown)


daltonclaybrook:
@yonaskolb Let me know if there are any changes you think need to be made, stylistic or otherwise.

daltonclaybrook:
@yonaskolb Is there any interest in getting this bug fixed? My team at Peloton has recently started using XcodeGen and we're loving it ❤️, but we currently have a hack-fix in place for this issue in particular. If this is not the best approach, please let me know.

Tests/XcodeGenKitTests/SchemeGeneratorTests.swift
```diff
@@ -336,6 +336,48 @@ class SchemeGeneratorTests: XCTestCase {
                 try expect(buildableReference?.blueprintName) == "ExternalTarget"
                 try expect(buildableReference?.referencedContainer) == "container:\(externalProject.string)"
             }
+
+            $0.it("generates scheme with remote runnable for watch") {
+                let xcscheme = try self.makeWatchScheme(appType: .watch2App, extensionType: .watch2Extension)
+                try expect(xcscheme.launchAction?.runnable).beOfType(XCScheme.RemoteRunnable.self)
```
brentleyjones: Can you add the opposite check elsewhere? To ensure we only make it a `RemoteRunnable` in this case, and a `BuildableProductRunnable` elsewhere?

tschob:
@daltonclaybrook Did you encounter this issue with a standalone watch app?

When debugging my schemes, I also noticed the difference between Xcode and XcodeGen regarding `RemoteRunnable`. But even with your fix, I am not able to create a scheme for a companion app (builds app + watch target). I assume that my (basic) setup is currently not supported with XcodeGen, but I also might simply miss the trick which let XcodeGen understand that it should pick the watch app as executable instead of the app target.

daltonclaybrook:
@tschob I tested this fix with a companion watch app. I will try to test again soon now that it's merged.

PR #799, opened 2020-03-04
Add stopOnEveryMainThreadCheckerIssue

ionutivan:
- I added a new property to help with issue 325, specifically stopOnEveryMainThreadCheckerIssue. 
- I also added unit tests & integration test. 
- I added documentation.
- Thanks for letting me know if this is ok as this is my first contribution to the project

### DIFF
--- a/Docs/ProjectSpec.md
+++ b/Docs/ProjectSpec.md
@@ -597,6 +597,7 @@ This is a convenience used to automatically generate schemes for a target based 
 - [ ] **testTargets**: **[[Test Target](#test-target)]** - a list of test targets that should be included in the scheme. These will be added to the build targets and the test entries. Each entry can either be a simple string, or a [Test Target](#test-target)
 - [ ] **gatherCoverageData**: **Bool** - a boolean that indicates if this scheme should gather coverage data. This defaults to false
 - [ ] **disableMainThreadChecker**: **Bool** - a boolean that indicates if this scheme should disable disable the Main Thread Checker. This defaults to false
+- [ ] **stopOnEveryMainThreadCheckerIssue**: **Bool** - a boolean that indicates if this scheme should stop at every Main Thread Checker issue. This defaults to false
 - [ ] **language**: **String** - a String that indicates the language used for running and testing. This defaults to nil
 - [ ] **region**: **String** - a String that indicates the region used for running and testing. This defaults to nil
 - [ ] **commandLineArguments**: **[String:Bool]** - a dictionary from the argument name (`String`) to if it is enabled (`Bool`). These arguments will be added to the Test, Profile and Run scheme actions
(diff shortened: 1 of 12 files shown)


brentleyjones:
Thank you @ionutivan! Can you please add a CHANGELOG entry as well?

Docs/ProjectSpec.md
```diff
@@ -597,6 +597,7 @@ This is a convenience used to automatically generate schemes for a target based
 - [ ] **testTargets**: **[[Test Target](#test-target)]** - a list of test targets that should be included in the scheme. These will be added to the build targets and the test entries. Each entry can either be a simple string, or a [Test Target](#test-target)
 - [ ] **gatherCoverageData**: **Bool** - a boolean that indicates if this scheme should gather coverage data. This defaults to false
 - [ ] **disableMainThreadChecker**: **Bool** - a boolean that indicates if this scheme should disable disable the Main Thread Checker. This defaults to false
+- [ ] **stopOnEveryMainThreadCheckerIssue**: **Bool** - a boolean that indicates if this scheme should stop at every Main Thread Checker issue. This defaults to false
```
yonaskolb: Could you add docs to down below in `Schemes` as well https://github.com/yonaskolb/XcodeGen/blob/master/Docs/ProjectSpec.md#scheme

ionutivan: Of course! Done

Tests/Fixtures/TestProject/project.yml
```diff
@@ -116,6 +116,7 @@ targets:
         - App_iOS_UITests
       gatherCoverageData: true
       disableMainThreadChecker: true
+      stopOnEveryMainThreadCheckerIssue: false
```
yonaskolb: Can we set this to true to generate the diff in the checked in fixture 👍

Tests/ProjectSpecTests/SpecLoadingTests.swift
```diff
@@ -712,6 +712,7 @@ class SpecLoadingTests: XCTestCase {
                     "language": "en",
                     "region": "US",
                     "disableMainThreadChecker": true,
+                    "stopOnEveryMainThreadCheckerIssue": false,
```
yonaskolb: Can we set this to true, otherwise we're not testing the parsing as false is the default

Tests/Fixtures/TestProject/project.yml
ionutivan: Done

Tests/ProjectSpecTests/SpecLoadingTests.swift
ionutivan: My understanding is that the default should be false because it shouldn't stop on every main thread checker issue. Am I right?

Tests/Fixtures/TestProject/project.yml
ionutivan: My understanding is that the default should be false because it shouldn't stop on every main thread checker issue. Am I right?

Tests/ProjectSpecTests/SpecLoadingTests.swift
ionutivan: @yonaskolb / @brentleyjones Can I please get a bit of help on this?

brentleyjones: @ionutivan you are correct about the default, what @yonaskolb was asking for is that we set it to `true` in a test, so we can test the parsing.

ionutivan: @brentleyjones Thank you for the help!

PR #803, opened 2020-03-11
Optimize bottlenecks

michaeleisel:
These are some bottlenecks that were causing significant performance overhead for me. The changes are rough, but if they look good overall I can clean them up

### DIFF
--- a/Sources/ProjectSpec/Decoding.swift
+++ b/Sources/ProjectSpec/Decoding.swift
@@ -4,18 +4,36 @@ import PathKit
 import Yams
 
 extension Dictionary where Key: JSONKey {
-
-    public func json<T: NamedJSONDictionaryConvertible>(atKeyPath keyPath: JSONUtilities.KeyPath, invalidItemBehaviour: InvalidItemBehaviour<T> = .remove) throws -> [T] {
+    public func json<T: NamedJSONDictionaryConvertible>(atKeyPath keyPath: JSONUtilities.KeyPath, invalidItemBehaviour: InvalidItemBehaviour<T> = .remove, parallel: Bool = false) throws -> [T] {
         guard let dictionary = json(atKeyPath: keyPath) as JSONDictionary? else {
             return []
         }
-        var items: [T] = []
-        for (key, _) in dictionary {
-            let jsonDictionary: JSONDictionary = try dictionary.json(atKeyPath: .key(key))
-            let item = try T(name: key, jsonDictionary: jsonDictionary)
-            items.append(item)
+        if parallel {
+            var itemOptionals: [T?] = Array(repeating: nil, count: dictionary.count)
+            var ops: [BlockOperation] = []
+            var idx: Int = 0
+            for (key, _) in dictionary {
+                ops.append(BlockOperation {[idx] in
+                    let jsonDictionary: JSONDictionary = try! dictionary.json(atKeyPath: .key(key))
+                    let item = try! T(name: key, jsonDictionary: jsonDictionary)
+                    itemOptionals[idx] = item
+                })
+                idx += 1
+            }
+            let queue = OperationQueue()
+            queue.qualityOfService = .userInteractive
+            queue.maxConcurrentOperationCount = 8
+            queue.addOperations(ops, waitUntilFinished: true)
+            return itemOptionals.compactMap { $0 }
+        } else {
+            var items: [T] = []
+            for (key, _) in dictionary {
+                let jsonDictionary: JSONDictionary = try dictionary.json(atKeyPath: .key(key))
+                let item = try T(name: key, jsonDictionary: jsonDictionary)
+                items.append(item)
+            }
+            return items
         }
-        return items
     }
 
     public func json<T: NamedJSONConvertible>(atKeyPath keyPath: JSONUtilities.KeyPath, invalidItemBehaviour: InvalidItemBehaviour<T> = .remove) throws -> [T] {
--- a/Sources/ProjectSpec/Decoding.swift
+++ b/Sources/ProjectSpec/Decoding.swift
@@ -9,14 +9,19 @@ extension Dictionary where Key: JSONKey {
             return []
         }
         if parallel {
-            var itemOptionals: [T?] = Array(repeating: nil, count: dictionary.count)
+            let defaultError = NSError(domain: "Unspecified error", code: 0, userInfo: nil)
+            var itemResults: [Result<T, Error>] = Array(repeating: .failure(defaultError), count: dictionary.count)
             var ops: [BlockOperation] = []
             var idx: Int = 0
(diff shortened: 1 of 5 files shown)


Sources/ProjectSpec/Decoding.swift
```diff
@@ -4,18 +4,36 @@ import PathKit
 import Yams
 
 extension Dictionary where Key: JSONKey {
-
-    public func json<T: NamedJSONDictionaryConvertible>(atKeyPath keyPath: JSONUtilities.KeyPath, invalidItemBehaviour: InvalidItemBehaviour<T> = .remove) throws -> [T] {
+    public func json<T: NamedJSONDictionaryConvertible>(atKeyPath keyPath: JSONUtilities.KeyPath, invalidItemBehaviour: InvalidItemBehaviour<T> = .remove, parallel: Bool = false) throws -> [T] {
```
michaeleisel: parallelizing the json parsing

Sources/XcodeGenKit/CarthageDependencyResolver.swift
```diff
@@ -10,11 +10,14 @@ import ProjectSpec
 import PathKit
 
 public class CarthageDependencyResolver {
```
michaeleisel: parallelizing dependency resolution

Sources/ProjectSpec/Decoding.swift
```diff
@@ -4,18 +4,36 @@ import PathKit
...
-            let jsonDictionary: JSONDictionary = try dictionary.json(atKeyPath: .key(key))
-            let item = try T(name: key, jsonDictionary: jsonDictionary)
-            items.append(item)
+        if parallel {
+            var itemOptionals: [T?] = Array(repeating: nil, count: dictionary.count)
+            var ops: [BlockOperation] = []
+            var idx: Int = 0
+            for (key, _) in dictionary {
+                ops.append(BlockOperation {[idx] in
+                    let jsonDictionary: JSONDictionary = try! dictionary.json(atKeyPath: .key(key))
```
mateuszjablonski-slabs: no need to use exclamation mark on `try`

Sources/ProjectSpec/Decoding.swift
```diff
@@ -4,18 +4,36 @@ import PathKit
...
+                    let item = try! T(name: key, jsonDictionary: jsonDictionary)
```
mateuszjablonski-slabs: no need to use exclamation mark on `try`

michaeleisel: agreed. any other concerns?

mateuszjablonski-slabs: nope. but bear in mind i am not a real reviewer 🙈

michaeleisel:
Is this project still maintained?

Sources/ProjectSpec/Decoding.swift
freddi-kit: According to another paring function, it would be better to set `try`, not `try!`

https://github.com/yonaskolb/XcodeGen/blob/a0e144b847c420ec6faa26fd5d029e5b94c0df95/Sources/ProjectSpec/Decoding.swift#L39

yonaskolb:
Thanks for your work here @michaeleisel. Would you be able to post some timings before and after this PR (after a rebase) for a large project, and also for the performance tests. Thanks!

Sources/XcodeGenKit/CarthageVersionLoader.swift
```diff
@@ -9,24 +9,52 @@ import Foundation
...
+        closure(&value)
+        semaphore.signal()
+    }
+}
+
 class CarthageVersionLoader {
 
     private let buildPath: Path
-    private var cachedFiles: [String: CarthageVersionFile] = [:]
+    private var cachedFilesMutex: Mutex<[String: CarthageVersionFile]> = Mutex([:])
```
michaeleisel: note that we'll need to be careful in the future not to add any non-thread-safe state to CarthageVersionLoader. for now it's just `cachedFiles` that needs this, but something to keep in mind

michaeleisel:
For carthage dependencies, on my 8 (physical) core machine it reduces the time from 600ms to 200ms for a large project. For the JSON parsing, it goes from 1,500 ms to 300ms. for performance testing, i can take a look. but first, are we good with these at a high level? (i.e., excluding stuff like try vs. try!, which can easily be fixed)

brentleyjones:
I can't speak for @yonaskolb, but I'm interested in this. From a cursory glance I'm good with it at a high level.

Sources/ProjectSpec/Decoding.swift
michaeleisel: fyi if we used zippyjson for the json parsing, even single-threaded it would be a lot (maybe 10x?) faster. but parallelizing is a simple win for this commit

Sources/ProjectSpec/Decoding.swift
michaeleisel: looking at this further, i think it would be simpler to use `try!` instead of `try`. the issue is that these blocks are being run asynchronously, so they don't just bubble up to this function's caller. there are other strategies, like storing a result type in the list and then if any of the results are failures, throw the first failure as the error out of this function. we can do that if people want though

michaeleisel: eh, i'll just do it with `try` actually

michaeleisel:
ok i'm happy with it now personally as well. here are the perf test results: https://gist.github.com/michaeleisel/3d1a5316d233571ca5e2ec9531351c82

michaeleisel:
are we good to merge?

Sources/XcodeGenKit/CarthageVersionLoader.swift
brentleyjones: Let's add a comment to `CarthageVersionLoader` to warn future us.

brentleyjones:
More details on our performance bottleneck:

image

Seems it's all in project generation and writing. In particular these two functions are hot spots: https://github.com/yonaskolb/XcodeGen/blob/01c6959f487fd67f57d05f1fe82d1263771cda25/Sources/XcodeGenKit/SourceGenerator.swift#L80-L82
(or just  https://github.com/yonaskolb/XcodeGen/blob/master/Sources/XcodeGenKit/PBXProjGenerator.swift#L222)

and

https://github.com/tuist/XcodeProj/blob/master/Sources/XcodeProj/Utils/ReferenceGenerator.swift#L27

michaeleisel:
Yeah, I've only ensured it improves my own use case. I spent some time with that `getAllSourceFiles`. Without a deep understanding, it seems somewhat difficult to optimize. Maybe we could open another issue about it.

michaeleisel:
@brentleyjones ptal

PR #812, opened 2020-03-21
Bugfix: settings override on occasion (v. 2.14.0)

tralf:
**Problem**
On our project we have to use build configurations with prefixed names (e. g. **Test** and **ABTest**), and it leads to a situation, when **ABTest** build settings are overridden with **Test** ones.

**Solution**
Updated `getBuildSettings` function of `Project` to take this edge case into consideration.

**Unit testing note**
As `getBuildSettings` result depends on `settings.configSettings` enumeration order, which is undefined, sometimes this function works correctly even without a provided fix. So, added unit test case may not fail on old implementation from time to time.

### DIFF
--- a/Sources/XcodeGenKit/SettingsBuilder.swift
+++ b/Sources/XcodeGenKit/SettingsBuilder.swift
@@ -81,7 +81,7 @@ extension Project {
         buildSettings += settings.buildSettings
 
         for (configVariant, settings) in settings.configSettings {
-            let isPartialMatch = config.name.lowercased().contains(configVariant.lowercased())
+            let isPartialMatch = config.name.lowercased().split(separator: " ").contains(Substring(configVariant.lowercased()))
             if isPartialMatch {
                 let exactConfig = getConfig(configVariant)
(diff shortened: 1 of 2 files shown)


Sources/XcodeGenKit/SettingsBuilder.swift
```diff
@@ -81,7 +81,7 @@ extension Project {
         buildSettings += settings.buildSettings
 
         for (configVariant, settings) in settings.configSettings {
-            let isPartialMatch = config.name.lowercased().contains(configVariant.lowercased())
+            let isPartialMatch = config.name.lowercased().split(separator: " ").contains(Substring(configVariant.lowercased()))
```
yonaskolb: As far as I can see this would be a breaking change, as putting `stag` for `Staging Release` would now fail, as this is now matching whole words instead of any substring.
We could change this so it finds a config in the following order:
- an exact match
- matching a word (space delimited)
- any substring

tralf: Got your point, thank you! I'll update the fix and add extra tests to cover this scenario as soon as have time.

PR #815, opened 2020-03-26
[WIP] Invalid resourcetags source to be throw error.

noppefoxwolf:
The `resourceTags` cannot be used in BuildPhase other than resource.
A incorrect spec should fail before generation.

https://github.com/yonaskolb/XcodeGen/blob/59ab2239227811f2607a4981c8d304219ca5f47d/Sources/XcodeGenKit/SourceGenerator.swift#L134

### DIFF
--- a/Sources/ProjectSpec/SpecValidation.swift
+++ b/Sources/ProjectSpec/SpecValidation.swift
@@ -185,6 +185,9 @@ extension Project {
                 if !source.optional && !sourcePath.exists {
                     errors.append(.invalidTargetSource(target: target.name, source: sourcePath.string))
                 }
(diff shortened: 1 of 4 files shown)


noppefoxwolf:
Always set resource flag when nothing to set buildPhase.
I may have misunderstood.

https://github.com/yonaskolb/XcodeGen/blob/59ab2239227811f2607a4981c8d304219ca5f47d/Sources/XcodeGenKit/SourceGenerator.swift#L276

noppefoxwolf:
Currently xcodegen set resource flag automatically.
So I'm not suggest this PR.

PR #823, opened 2020-04-03
Add scheme actions to build step for TargetScheme

brentleyjones:
We say in the documentation that the preActions and postActions of a TargetScheme are applied to _all_ steps. They weren't being applied to the build step though.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@
 - Prefix static library target filenames with 'lib' to match Xcode. [#831](https://github.com/yonaskolb/XcodeGen/pull/831) @ileitch
 - Fixed duplicate addition of carthage static frameworks. [#829](https://github.com/yonaskolb/XcodeGen/pull/829) @funzin
 
+#### Fixed
+- Set `preActions` and `postActions` on the `build` action of a TargetScheme instead of the other actions. [#823](https://github.com/yonaskolb/XcodeGen/pull/823) @brentleyjones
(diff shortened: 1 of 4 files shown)


yonaskolb:
@brentleyjones this is interesting. Wouldn't this mean an action would run twice? Once on build and then once in the action someone is launching. If so one could argue this should happen on everything except build, or just build

brentleyjones:
I agree, I found this bug because I wanted it to run on build, but not the others. I submitted the fix based on the description. Ideally it would be build only.

yonaskolb:
Out of interest what breaks when the script is run on the active step and not build?
I'd be open to changing this PR to just happen on build. Can you think of that breaking anything?

Also as a note, the documentation is referring to all build phases instead of all launch types
> Scripts that are run before all actions

brentleyjones:
When you build a static library, it doesn't have a "run" action, you have to use "build". We are trying to collect build statistics that require scripts to be run before and after a build, and it doesn't work for static libraries right now.

>Can you think of that breaking anything?

If people were doing something non-build related. I can't think of what one would want to do here, but: https://xkcd.com/1172/

yonaskolb:
Haha, I was actually thinking about that XKCD comic!
I think it will be fine to move everything over to build, which would be the correct behaviour. Could you do that in this PR and also update the documentation for it

brentleyjones:
@yonaskolb Done!

PR #824, opened 2020-04-06
Add improved support for simple iOS sticker packs

scelis:
In the process of updating some of our iOS sticker packs to use XcodeGen, I ran into two issues:

1. By default, messages applications for simple iOS sticker packs created in Xcode do not include a compile sources phase. While XcodeGen supports this for the sticker pack extension, it does not for the messages application itself. The first commit in this PR addresses this issue and allows the target to skip the compile sources phase if the application has no sources.

2. In order for the run scheme to function properly, [`launchAutomaticallySubstyle`](https://rubydoc.info/gems/xcodeproj/Xcodeproj/XCScheme/LaunchAction#launch_automatically_substyle-instance_method) must be set to 2. Doing so will change the executable dropdown from the sticker app to `Ask on Launch`. Without this change, running the scheme will cause the simulator to open but hang instead of launching Messages.

Screen Shot 2020-04-06 at 8 08 44 AM

The second commit in this PR addresses that issue and updates the TestProject fixture which needed to be changed.

Since none of this is obvious, I added an example project for a simple sticker pack to `Examples.md`.


brentleyjones:
Thanks @scelis! Can you add ~a changelog entry and~ some tests for this?

scelis:
@brentleyjones Done!

PR #826, opened 2020-04-08
Bugfix: Consider folders with dot in it when generating projects

### LINKED ISSUE #825: Bug when generating projects for paths which have a dot in it
I am facing this issue when running xcodegen on my work project, which has a `.` in some of the paths which have swift files.

The issue is that these files get added to `Copy Bundle Resources` when they should be in `Compile sources` phase

Here's a [demo project](https://github.com/asifmohd/demo-xcodegen-app) hosted on github which recreates the issue.

asifmohd:
Fixes #825 

Folders which would have a dot in it, would be added to Copy Bundle Resources even though it contained swift files

- This PR, updates the if checks for directories and file path in `SourceGenerator.swift`.
- I have added tests for the same
- Have updated the TestProject fixture as well

### DIFF
--- a/Sources/XcodeGenKit/SourceGenerator.swift
+++ b/Sources/XcodeGenKit/SourceGenerator.swift
@@ -413,12 +413,14 @@ class SourceGenerator {
         let children = try getSourceChildren(targetSource: targetSource, dirPath: path, excludePaths: excludePaths, includePaths: includePaths)
 
         let createIntermediateGroups = targetSource.createIntermediateGroups ?? project.options.createIntermediateGroups
+        let directoryExtensionsToSkip = ["lproj", "bundle", "xcassets", "xcdatamodeld"]
 
         let directories = children
-            .filter { $0.isDirectory && $0.extension != "lproj" && $0.extension != "xcassets" && $0.extension != "bundle" && $0.extension != "xcdatamodeld" }
+            .filter { $0.isDirectory && !directoryExtensionsToSkip.contains($0.extension ?? "") }
 
+        let whitelistedDirectoryExtensionsForFilePath = ["xcdatamodeld", "xcdatamodel" , "xcassets" , "intentdefinition"]
         let filePaths = children
-            .filter { $0.isFile || $0.extension != nil && $0.extension != "lproj" && ($0.extension == "xcdatamodeld" || $0.extension == "xcdatamodel" || $0.extension == "xcassets" || $0.extension == "intentdefinition" || !$0.isDirectory) }
+            .filter { $0.isFile || $0.extension != nil && $0.extension != "lproj" && (whitelistedDirectoryExtensionsForFilePath.contains($0.extension ?? "") || !$0.isDirectory) }
 
         let localisedDirectories = children
             .filter { $0.extension == "lproj" }
--- a/Sources/XcodeGenKit/SourceGenerator.swift
+++ b/Sources/XcodeGenKit/SourceGenerator.swift
@@ -415,10 +415,10 @@ class SourceGenerator {
         let createIntermediateGroups = targetSource.createIntermediateGroups ?? project.options.createIntermediateGroups
 
         let directories = children
-            .filter { $0.isDirectory && $0.extension != "lproj" && $0.extension != "xcassets" && $0.extension != "bundle" }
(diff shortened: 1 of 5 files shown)


Sources/XcodeGenKit/SourceGenerator.swift
```diff
@@ -415,10 +415,10 @@ class SourceGenerator {
         let createIntermediateGroups = targetSource.createIntermediateGroups ?? project.options.createIntermediateGroups
 
         let directories = children
-            .filter { $0.isDirectory && $0.extension == nil && $0.extension != "lproj" }
+            .filter { $0.isDirectory && $0.extension != "lproj" && $0.extension != "xcassets" && $0.extension != "bundle" }
 
         let filePaths = children
-            .filter { $0.isFile || $0.extension != nil && $0.extension != "lproj" }
+            .filter { $0.isFile || $0.extension != nil && $0.extension != "lproj" && !$0.isDirectory }
```
asifmohd, line 421: This is to make sure that the directory doesn't get added as a file

If we don't have this if check, then the generated project has two references of the folder, one in Compile Sources and one in Copy Bundle Resources phase

Sources/XcodeGenKit/SourceGenerator.swift
```diff
@@ -415,10 +415,10 @@ class SourceGenerator {
         let createIntermediateGroups = targetSource.createIntermediateGroups ?? project.options.createIntermediateGroups
 
         let directories = children
-            .filter { $0.isDirectory && $0.extension == nil && $0.extension != "lproj" }
+            .filter { $0.isDirectory && $0.extension != "lproj" && $0.extension != "xcassets" && $0.extension != "bundle" }
 
         let filePaths = children
-            .filter { $0.isFile || $0.extension != nil && $0.extension != "lproj" }
+            .filter { $0.isFile || $0.extension != nil && $0.extension != "lproj" && ($0.extension == "xcdatamodeld" || !$0.isDirectory) }
```
asifmohd, line 421: This is to make sure that the directory doesn't get added as a file

If we don't have this if check, then the generated project has two references of the folder, one in Compile Sources and one in Copy Bundle Resources phase

If we don't have the check for `xcdatamodeld` a unit test related to versionGroup checks starts failing, probably because it won't be able to find `.xccurrentversion` if the directory is not considered in the file path.

Sources/XcodeGenKit/SourceGenerator.swift
```diff
@@ -415,10 +415,10 @@ class SourceGenerator {
         let createIntermediateGroups = targetSource.createIntermediateGroups ?? project.options.createIntermediateGroups
 
         let directories = children
-            .filter { $0.isDirectory && $0.extension == nil && $0.extension != "lproj" }
+            .filter { $0.isDirectory && $0.extension != "lproj" && $0.extension != "xcassets" && $0.extension != "bundle" && $0.extension != "xcdatamodeld" }
```
asifmohd, line 418: The extension == nil, would have always prevented the next && from being evaluated.

So I have removed that and explicitly whitelisted the required folders. Most of these extensions are something I have discovered after trying to fix broken unit tests

Tests/XcodeGenKitTests/SourceGeneratorTests.swift
```diff
@@ -400,12 +403,15 @@ class SourceGeneratorTests: XCTestCase {
                     - B:
                       - b.swift
                       - c.jpg
+                    - D2.0:
```
brentleyjones, line 406: Can you also include all of the `directoryExtensionsToSkip` in a test somewhere, to ensure we didn't break those?

asifmohd: - There is already a separate test for `xcdatamodel` [here](https://github.com/yonaskolb/XcodeGen/blob/b1edbda5831ae8f0a996ef7aec5112bc1195be51/Tests/XcodeGenKitTests/SourceGeneratorTests.swift#L101)
- Existing tests for `lproj` [here](https://github.com/yonaskolb/XcodeGen/blob/b1edbda5831ae8f0a996ef7aec5112bc1195be51/Tests/XcodeGenKitTests/SourceGeneratorTests.swift#L166) and [here](https://github.com/yonaskolb/XcodeGen/blob/b1edbda5831ae8f0a996ef7aec5112bc1195be51/Tests/XcodeGenKitTests/SourceGeneratorTests.swift#L124)
- tests for `xcassets` were already covered in the same test linked to this comment
- I have updated the test to include an assert for `bundle` though

Issue #708, opened 2019-11-07
SwiftPM dependencies don't link their required dependencies


JonasVautherin:
Adding `transitivelyLinkDependencies: true` to my options or target has no effect on the resulting xcode project.

I am depending on SwiftPM packages using the `package` dependency type. For instance, I depend on `grpc-swift`, which depends on `swift-nio`, which itself builds `CNIOAtomics`. I can build grpc-swift with xcodegen ([project.yml here](https://github.com/JonasVautherin/grpc-swift/blob/nio/project.yml)), and in Xcode I see that it does build CNIOAtomics. Then from Carthage, I try to depend on that grpc-swift (the point of my using xcodegen being Carthage), and there, `CNIOAtomics` is missing. So `cartage update` works, but building the end project fails with "Missing required module 'CNIOAtomics'".

Now, if I add `swift-nio` to the `packages` of my end project, then it works, suggesting an error with transitive dependencies. I thought that maybe that was because I should set `transitivelyLinkDependencies: true`, but this has no effect on the resulting `project.pbxproj`.

Am I missing something? Is it wrong to use `transitivelyLinkDependencies` with SwiftPM packages?

brentleyjones:
Ahh, `transitivelyLinkDependencies` wasn't updated to support SwiftPM packages when they were added. I'll see if I can get around to this in the next couple days.

JonasVautherin:
That would be awesome! I'm trying to bring Carthage support to grpc-swift in a nice way, and I find xcodegen is a super clean way to deal with that.

I am not super advanced in Swift and I don't know the xcodegen codebase, but I could have a look if you pointed me to a place where to start :blush:. Otherwise I will gladly test and review you changes!

brentleyjones:
@JonasVautherin I have a feeling it would be somewhere around https://github.com/yonaskolb/XcodeGen/blob/master/Sources/XcodeGenKit/PBXProjGenerator.swift#L1020-L1023

Though, thinking more about this, a similar thing happens when defining a dependency on a Carthage dependency that itself has another dependency. That isn't resolved with `transitivelyLinkDependencies`, which only links transitive declare dependencies in parent targets. I think this is currently handled by the `findCarthageFrameworks` for Carthage https://github.com/yonaskolb/XcodeGen/blob/master/Sources/XcodeGenKit/PBXProjGenerator.swift#L605-L608 (though I don't use it, so maybe it's not for that).

[...]

PR #830, opened 2020-04-12
Fix linking of multiple products from the same Swift Package

toshi0383:
## Project Spec

https://github.com/toshi0383/abceed/blob/bc1bad08ea861c737c712bb95612f51bd1aad55b/project.yml

```yml
...
  AbceedCore:
    platform: iOS
    type: framework.static
    sources:
      - AbceedCore/
    dependencies:
      - package: RxSwift
        product: RxSwift
      - package: RxSwift
        product: RxCocoa
      - package: RxSwift
        product: RxRelay
...
```

## Current Behavior

スクリーンショット 2020-04-12 20 17 25

## Expected Behavior

スクリーンショット 2020-04-12 20 16 34

## Solution

- [x] Fixed not to ignore products in `getAllDependenciesPlusTransitiveNeedingEmbedding` function.
- [x] write test for this


brentleyjones:
Nice! I think this will fix https://github.com/yonaskolb/XcodeGen/issues/708.

toshi0383:
Test fixed and added CHANGELOG entry.
Ready for review!🙌

yonaskolb:
Thanks @toshi0383! As far as I understand this doesn't so much seem to fix transitive dependencies but rather fixing linking multiple products from the same package right?

toshi0383:
@yonaskolb Yes! Fixed the title. Thanks for fixing changelog entry!

PR #656, opened 2019-09-15
Add option to gather code coverage for specific targets

kateinoigakukun:
To avoid conflict, this branch is based on https://github.com/yonaskolb/XcodeGen/pull/655 and https://github.com/tuist/XcodeProj/pull/473.

**Please merge https://github.com/yonaskolb/XcodeGen/pull/655 first :pray:**

I added `enableCoverage` option for each test target to gather code coverage for specific targets.

PR #832, opened 2020-04-13
Add ability to specify "full" scheme actions with `TargetScheme`

brentleyjones:
FYI: I'll break this in a couple PRs (code cleanup, adding `parallelizeBuild`, adding some test fixtures, and finally adding the full scheme support to `target`) if everything looks good.

---

This PR adds support for specifying full scheme configuration to `target: Target: scheme:`. If any of the actions are specified (e.g. `build`, `test`, `profile`) then that block will be used the same as a top level `scheme` declaration. For any steps not specified they will use the existing helper/defaulting logic.

This allows one to do something like this:

```yml
targets:
  App:
    scheme:
      build:
        targets:
          App: all
      run:
        commandLineArguments:
          "-enable-perf-logging": false
        environmentVariables:
          - variable: TSAN_OPTIONS
            value: suppressions=$(SRCROOT)/tsan-suppressions.txt
            isEnabled: true
      test:
        targets:
         - Apps Tests
        gatherCoverageData: true
      profile:
        config: Profile
      archive:
        config: Release
```

This should help address issue like #656.

### DIFF
--- a/Tests/Fixtures/TestProject/Project.xcodeproj/xcshareddata/xcschemes/App_iOS Production.xcscheme
+++ b/Tests/Fixtures/TestProject/Project.xcodeproj/xcshareddata/xcschemes/App_iOS Production.xcscheme
@@ -3,8 +3,8 @@
    LastUpgradeVersion = "1020"
    version = "1.3">
    <BuildAction
-      parallelizeBuildables = "YES"
-      buildImplicitDependencies = "YES">
+      parallelizeBuildables = "NO"
+      buildImplicitDependencies = "NO">
       <BuildActionEntries>
          <BuildActionEntry
             buildForTesting = "YES"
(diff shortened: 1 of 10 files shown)


Tests/Fixtures/TestProject/Project.xcodeproj/xcshareddata/xcschemes/App_iOS Production.xcscheme
```diff
@@ -3,8 +3,8 @@
    LastUpgradeVersion = "1020"
    version = "1.3">
    <BuildAction
-      parallelizeBuildables = "YES"
-      buildImplicitDependencies = "YES">
+      parallelizeBuildables = "NO"
+      buildImplicitDependencies = "NO">
```
yonaskolb, line 7: Looks like the default changed

yonaskolb: Never-mind looks like you set it explicitly down below

yonaskolb:
@brentleyjones love this! 👏

brentleyjones:
Some things came up, but I'll get to this soon.

brentleyjones:
If this is still desired, someone else will have to re-submit it. I no longer use XcodeGen.

PR #835, opened 2020-04-17
Prevent tests from being set as launch actions

brentleyjones:
When a scheme only contains a test target, and you want it to build when "Building for Run", you don't want the test target to be set as the launch action. This fixes that.


Sources/XcodeGenKit/SchemeGenerator.swift
```diff
@@ -377,3 +377,10 @@ extension PBXProductType {
         }
     }
 }
+
+extension Target {
+    var shouldExecuteOnLaunch: Bool {
```
yonaskolb, line 382: What do you think about moving this into https://github.com/yonaskolb/XcodeGen/blob/master/Sources/ProjectSpec/XCProjExtensions.swift where all the other similar helpers are?

PR #840, opened 2020-04-22
Fixes dumping project references in Project.swift

paciej00:
# What
When dumping to YAML a manifest that contains project references one can get the following `fatalError`:

```
Error: Failed to represent ("MyProject", ["path": "../MyProject/MyProject.xcodeproj"])
```

It is caused by wrong representation of project references (`[(String, Any)]` instead of a `Dictionary`)


paciej00:
It helps a lot, thanks for implementing it and implementing **XcodeGen** itselft :)! I use **XcodeGenKit** manifest dumping to modify an existing project manifest with a newly generated framework module that would reside in an external project. I hope the short diagram below will explain a bit what I use it for:

```
1. MainProject -> Project A, Project B, Project C
2. Generate Project D containing framework D
3. Read, update & dump MainProject/project.yml
4. MainProject -> Project A, Project B, Project C, Project D 
```

BTW, I noticed that the dumping puts to a yaml stuff that wasn't previously there. I have a solution for that. I will create a separate PR so that you could have a look on it and decide if it's worth including in the tool itself. I need it because I work in a complicated project with 30+ internal dependencies coming from other projects and digging through such a huge yaml is challenging ;)

PR #716, opened 2019-11-10
Add Test Plans

yonaskolb:
Adds test plan support
Resolves #684 

- [x] reference test plans in schemes
- [x] generate test plan references in schemes
- [x] parsing tests
- [x] generation tests
- [x] updates scheme versions
- [x] remove xctestplan from resources
- [x] validation errors
- [x] validation tests
- [ ] ability to define test plans in spec (probably not required) OR:
- [ ] way to update UUID references in test plan files
- [x] docs

Possible changes:
- make top level `testPlans` which are then referenced in schemes, similar to the new `projectReferences`?
- ability to choose a default test plan? At the moment the first test plan in a scheme will be the default
- ability to define test plans in the spec itself? If not we probably need a way to update the UUIDs in existing files in case XcodeGen/XcodeProj ever changes them

PR #846, opened 2020-04-24
Test plans continued

omares:
This supersedes #716 

### Additional changes

- [x] ability to choose a default test plan
- [x] validation errors
- [x] validation tests
- [x] docs
- [ ] way to update UUID references in test plan files

### Checklist from the previous PR

- [x]  reference test plans in schemes
- [x]  generate test plan references in schemes
- [x]  parsing tests
- [x]  generation tests
- [x]  updates scheme versions
- [x]  remove xctestplan from resources

### DIFF
--- a/Docs/ProjectSpec.md
+++ b/Docs/ProjectSpec.md
@@ -868,7 +868,7 @@ schemes:
     test:
       testPlans:
         - path: app.xctestplan
-          default: true
+          defaultPlan: true
 ```
 
 ### Scheme Template
--- a/Docs/ProjectSpec.md
+++ b/Docs/ProjectSpec.md
@@ -620,6 +620,7 @@ This is a convenience used to automatically generate schemes for a target based 
 - [ ] **region**: **String** - a String that indicates the region used for running and testing. This defaults to nil
 - [ ] **commandLineArguments**: **[String:Bool]** - a dictionary from the argument name (`String`) to if it is enabled (`Bool`). These arguments will be added to the Test, Profile and Run scheme actions
 - [ ] **environmentVariables**: **[[Environment Variable](#environment-variable)]** or **[String:String]** - environment variables for Run, Test and Profile scheme actions. When passing a dictionary, every key-value entry maps to a corresponding variable that is enabled.
+- [ ] **testPlans**: **[String]** - List of test plan locations that will be referenced in the scheme.
 - [ ] **preActions**: **[[Execution Action](#execution-action)]** - Scripts that are run *before* the build action
 - [ ] **postActions**: **[[Execution Action](#execution-action)]** - Scripts that are run *after* the build action
 
--- a/Docs/ProjectSpec.md
+++ b/Docs/ProjectSpec.md
@@ -620,7 +620,7 @@ This is a convenience used to automatically generate schemes for a target based 
 - [ ] **region**: **String** - a String that indicates the region used for running and testing. This defaults to nil
 - [ ] **commandLineArguments**: **[String:Bool]** - a dictionary from the argument name (`String`) to if it is enabled (`Bool`). These arguments will be added to the Test, Profile and Run scheme actions
 - [ ] **environmentVariables**: **[[Environment Variable](#environment-variable)]** or **[String:String]** - environment variables for Run, Test and Profile scheme actions. When passing a dictionary, every key-value entry maps to a corresponding variable that is enabled.
-- [ ] **testPlans**: **[String]** - List of test plan locations that will be referenced in the scheme.
+- [ ] **testPlans**:  **[[Test Plan](#test-plan)]** - List of test plan locations that will be referenced in the scheme.
 - [ ] **preActions**: **[[Execution Action](#execution-action)]** - Scripts that are run *before* the build action
 - [ ] **postActions**: **[[Execution Action](#execution-action)]** - Scripts that are run *after* the build action
 
@@ -856,6 +856,19 @@ schemes:
       config: prod-release
       customArchiveName: MyTarget
       revealArchiveInOrganizer: false
+```
+### Test Plan
+
+- [x] **path**: **String** - path that provides the `xctestplan` location.
+- [ ] **defaultPlan**: **Bool** - a bool that defines if given plan is the default one. Defaults to false.
+
+```yaml
+schemes:
+  TestTarget:
+    test:
+      testPlans:
+        - path: app.xctestplan
+          default: true
 ```
 
 ### Scheme Template
--- a/Sources/ProjectSpec/SpecValidation.swift
+++ b/Sources/ProjectSpec/SpecValidation.swift
@@ -201,6 +201,10 @@ extension Project {
             if let action = scheme.run, let config = action.config, getConfig(config) == nil {
                 errors.append(.invalidSchemeConfig(scheme: scheme.name, config: config))
             }
+
+            let invalidTestPlans: [TestPlan] = scheme.test?.testPlans.filter { !(basePath + $0.path).exists } ?? []
+            errors.append(contentsOf: invalidTestPlans.map{ .invalidTestPlan($0) })
+
             if let action = scheme.test, let config = action.config, getConfig(config) == nil {
                 errors.append(.invalidSchemeConfig(scheme: scheme.name, config: config))
             }
--- a/Sources/ProjectSpec/SpecValidation.swift
+++ b/Sources/ProjectSpec/SpecValidation.swift
@@ -205,6 +205,11 @@ extension Project {
             let invalidTestPlans: [TestPlan] = scheme.test?.testPlans.filter { !(basePath + $0.path).exists } ?? []
             errors.append(contentsOf: invalidTestPlans.map{ .invalidTestPlan($0) })
 
+            let defaultPlanCount = scheme.test?.testPlans.filter { $0.defaultPlan }.count ?? 0
+            if (defaultPlanCount > 1) {
+                errors.append(.multipleDefaultTestPlans)
+            }
+
             if let action = scheme.test, let config = action.config, getConfig(config) == nil {
                 errors.append(.invalidSchemeConfig(scheme: scheme.name, config: config))
             }
--- a/Sources/ProjectSpec/SpecValidationError.swift
+++ b/Sources/ProjectSpec/SpecValidationError.swift
@@ -33,6 +33,7 @@ public struct SpecValidationError: Error, CustomStringConvertible {
         case invalidProjectReference(scheme: String, reference: String)
         case invalidProjectReferencePath(ProjectReference)
         case invalidTestPlan(TestPlan)
+        case multipleDefaultTestPlans
 
         public var description: String {
             switch self {
@@ -82,6 +83,8 @@ public struct SpecValidationError: Error, CustomStringConvertible {
                 return "Project reference \(reference.name) has a project file path that doesn't exist \"\(reference.path)\""
             case let .invalidTestPlan(testPlan):
                 return "Test path \(testPlan.path) doesn't exist"
+            case let .multipleDefaultTestPlans:
+                return "Your test plan definition contains more than one default test plan"
             }
         }
     }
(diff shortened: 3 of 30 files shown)


Docs/ProjectSpec.md
```diff
@@ -838,6 +839,19 @@ schemes:
       customArchiveName: MyTarget
       revealArchiveInOrganizer: false
 ```
+### Test Plan
+
+- [x] **path**: **String** - path that provides the `xctestplan` location.
+- [ ] **defaultPlan**: **Bool** - a bool that defines if given plan is the default one. Defaults to false.
```
armintelker, line 845: here we call it `defaultPlan`. In  [Line 853](https://github.com/yonaskolb/XcodeGen/pull/846/files#diff-938b5cf412212f1137a53f072512963aR853) we name it  `default`

Sources/ProjectSpec/SpecValidation.swift
```diff
@@ -201,6 +201,10 @@ extension Project {
             if let action = scheme.run, let config = action.config, getConfig(config) == nil {
                 errors.append(.invalidSchemeConfig(scheme: scheme.name, config: config))
             }
+
+            let invalidTestPlans: [TestPlan] = scheme.test?.testPlans.filter { !(basePath + $0.path).exists } ?? []
```
armintelker, line 205: Idea: 
You could save a `map` loop when you combine the logic of `filter` and `map` in a `compactMap`

**META CODE**
```swift 
if let testPlans = scheme.test?.testPlans { // You can also do `??` if you like it more. but since the lines above and below use `if let...` we should go with it too. 
    let invalidTestPlans [ValidationError] = testPlans.compactMap { 
        !(basePath + $0.path).exists ? .invalidTestPlan($0) : nil 
    }
    errors.append(contentsOf: invalidTestPlans)
}
```
or a bit shorter
```swift 
if let testPlans = scheme.test?.testPlans {
     errors.append(contentsOf: testPlans.compactMap { 
        !(basePath + $0.path).exists ? .invalidTestPlan($0) : nil // Not sure if compiler will find `ValidationError` enum
    })
}
```

Sources/ProjectSpec/TestPlan.swift
```diff
@@ -0,0 +1,44 @@
...
+        path = try jsonDictionary.json(atKeyPath: "path")
+        defaultPlan = jsonDictionary.json(atKeyPath: "defaultPlan") ?? false
+    }
+}
+
+extension TestPlan: JSONEncodable {
+    public func toJSONValue() -> Any {
+        [
+            "path": path,
+            "defaultPlan": defaultPlan,
```
armintelker, line 31: Maybe we could go with `default` because apple is using the term `default` it in the `xscheme`
But on the other hand `reference` is also a worse name than `path`. 
Let's discuss it. :)
```xml
<TestPlans>
    <TestPlanReference
        reference = "container:first.xctestplan"
        default = "YES">
    </TestPlanReference>
    <TestPlanReference
        reference = "container:second.xctestplan">
    </TestPlanReference>
</TestPlans>
```

Sources/XcodeGenKit/SchemeGenerator.swift
```diff
@@ -197,10 +197,18 @@ public class SchemeGenerator {
         let launchVariables = scheme.run.flatMap { $0.environmentVariables.isEmpty ? nil : $0.environmentVariables }
         let profileVariables = scheme.profile.flatMap { $0.environmentVariables.isEmpty ? nil : $0.environmentVariables }
 
+        let testPlans = scheme.test?.testPlans.map {
+            testPlan in XCScheme.TestPlanReference(
+                reference: "container:\(testPlan.path)",
```
armintelker, line 202: Small thing but:
The code should stick to one type of string operation. on some places it uses concatenation and then interpolation. At some places we use `a + b` on this here `"xy\()"`

Here: 
```swift
"container:\(testPlan.path)"
```

`Tests/XcodeGenKitTests/SchemeGeneratorTests.swift`
```swift
 let testPlanPath = fixturePath.string + "/TestProject/App_iOS/App_iOS.xctestplan"
```

Sources/ProjectSpec/TestPlan.swift
omares: `default` is a reserved word and should not be used as a variable name or similar. Even if its possible to use it, via escaping, i am strongly against that as it can cause more trouble than doing good.

Sources/ProjectSpec/SpecValidationError.swift
```diff
@@ -79,6 +80,8 @@ public struct SpecValidationError: Error, CustomStringConvertible {
                 return "Scheme \(scheme.quoted) has invalid project reference \(project.quoted)"
             case let .invalidProjectReferencePath(reference):
                 return "Project reference \(reference.name) has a project file path that doesn't exist \"\(reference.path)\""
+            case let .invalidTestPlan(testPlan):
+                return "Test path \(testPlan.path) doesn't exist"
             }
```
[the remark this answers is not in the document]
omares: I don't want to outsmart the creator of the scheme file, maybe there is no default test plan on purpose. Would not do this for now.

Sources/XcodeGenKit/SchemeGenerator.swift
omares: Changed in 642ba14

Docs/ProjectSpec.md
omares: Changed in 642ba14

Sources/ProjectSpec/SpecValidationError.swift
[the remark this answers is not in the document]
omares: Changed in 642ba14

Docs/ProjectSpec.md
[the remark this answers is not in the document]
omares: I don't wont to outsmart the creator of the scheme file. Maybe there is no default test plan on purpose. Would not do this for now.

Sources/ProjectSpec/SpecValidation.swift
omares: I prefer my version of the code as it moves downwards instead of to the right (nesting), thus I find it more readable.

Loop performance should not be an issue here as this code is executed on macs with modern hardware and the lists will be very small.

omares:
@yonaskolb & @giginet I am eager to get your feedback on this PR.

Also, I am stuck with the "way to update UUID references in test plan files" as i am not sure what's meant with that.

yonaskolb:
Thanks for your work here @omares! Would you be able to target my branch with this PR so the changes are more easily reviewable? Maybe we can merge that into there.

In terms of `way to update UUID references in test plan files`, that is the big sticking point in this whole addition. The problem is that the test plans require ids of the targets. You could create a test plan in Xcode after the project is generated, but the UUID's of the targets may change later, so then the test plan wouldn't work anymore. Some solutions could be:
- define and generate the whole test plan file with xcodegen. Would prefer not to lump every little setting into here.
- have some way of identifying and replacing ids dynamically. Would be useful for schemes too.
- some other solution?

How did you imagine this working?

selin194:
Hey @omares! Thanks for this PR for adding `test plan` support to XcodeGen. When this feature will be released? Thank you!

omares:
Hey @selin194 and @yonaskolb. Sorry for not getting back earlier, somehow I did not notice the activity in this PR, probably due to the long silence.

> Thanks for your work here @omares! Would you be able to target my branch with this PR so the changes are more easily reviewable? Maybe we can merge that into there.

Not 100% sure if this is easily doable as I updated my branch with changes from the master. I will try and see what happens.

> In terms of `way to update UUID references in test plan files`, that is the big sticking point in this whole addition. The problem is that the test plans require ids of the targets. You could create a test plan in Xcode after the project is generated, but the UUID's of the targets may change later, so then the test plan wouldn't work anymore. Some solutions could be:
> 
>     * define and generate the whole test plan file with xcodegen. Would prefer not to lump every little setting into here.
>     * have some way of identifying and replacing ids dynamically. Would be useful for schemes too.
>     * some other solution?
> 
> 
> How did you imagine this working?

I really have no preference/solution, as I lack the experience when it comes to XcodeGen and the inserts of the apple config files. Is there a native way to parse IDs out of the files? If we have to do that manually, this will be a lot of pain :)

Really open and grateful for suggestions on how to tackle this!

omares:
I investigated a bit.


@yonaskolb Please provide your input and correct me where I am wrong, without a sparing partner this will be hard for me.

- A project can have multiple `xcscheme` files that contain references to the identifiers inside `xctestplan`
- The identifiers inside `xcscheme` are changing "magically" and thus we have to update the testplan files.
- Parsing the `xcscheme` and `xctestplan` files should be doable. As one is XML the second one Json. Probably some Xpath magic has to be involved looking for `<BuildableReference>` elements that have a `BuildableName` attribute that ends with `xctest`
- When should we trigger the process?

dpfannenstiel:
@omares 
> The identifiers inside xcscheme are changing "magically" and thus we have to update the testplan files.

I don't think the point of the testplan isn't to have the testplan file; it's to separate out the test plans from the xcodeproj files.
  
If there are UUIDs in the test plans that get modified based on the `xcscheme` it sounds like the right path, in the context of xcodegen, is to define a type `TestPlan` that can be defined in the project YAML.  The `TestPlan` can then be assigned to the currently existing `Scheme` objects.  Any given `TestPlan` will have a base configuration.  Additional configurations may be defined.  These additional configurations can override any value in the base configuration.  When the project is created the test plan files should then be generated and added to the project.  But as with any project setting the source of truth will always be the YAML.

jsorge:
Any updates on this? I'm hoping to start using XcodeGen for our project at work and we have test plans, so this is a must in order to hop aboard the XcodeGen train 🙂

liamnichols:
> way to update UUID references in test plan files

If it's any use.. I've found that the references actually seem to be meaningless. As a work around for now official test plan support, my strategy has been to add the following to my spec file:

```yml
options:
  # While TestPlans are not supported by XcodeGen, we manually inject the XML into the scheme.
  # This is enough for Xcode 11 to read in order to use the Test Plan however upon opening the
  # scheme, Xcode will make some additional changes that we don't need to be too concerned about.
  postGenCommand: >-
    FILE='Global.xcodeproj/xcshareddata/xcschemes/Screenshot-Tests.xcscheme';
    echo "Manually patching '${FILE}' to inject Test Plan while waiting for https://github.com/yonaskolb/XcodeGen/pull/716.";
    sed -i '' -E 's/<\/Testables>/<\/Testables><TestPlans><TestPlanReference reference="container:GlobalUITests\/Tests\/Snapshots\/TestPlan.xctestplan" default="YES"><\/TestPlanReference><\/TestPlans>/g' $FILE;
    UI_TEST_SCHEME='Global.xcodeproj/xcshareddata/xcschemes/UI-Tests.xcscheme';
    sed -i '' -E 's/<\/Testables>/<\/Testables><TestPlans><TestPlanReference reference="container:GlobalUITests\/Tests\/EndToEnd\/TestPlans\/UI-TestPlan.xctestplan" default="YES"><\/TestPlanReference><\/TestPlans>/g' $UI_TEST_SCHEME;
```

This little hack as actually been enough to have Test Plans working like normal despite the fact that that the .xctestplan file is actually referencing an identifier that doesn't exist. 

I'm not sure how or why this works, but given that it does work (and continues to work on Xcode 12), does it help in any way knowing that?

mat1th:
What is the status of this PR? I would love to use the Test plans support in my project.

albertodebortoli:
I would love this too very, very much. Not having support for test plans sadly blocks us from using XcodeGen.

namolnad:
> I would love this too very, very much. Not having support for test plans sadly blocks us from using XcodeGen.

@albertodebortoli not to say this can't be improved, but fwiw we utilize test plans with xcodegen. Because the ids generated are stable we are able to keep the test plans in source control and just add test targets manually. Not perfect, but if it unblocks you from starting to use xcodegen in the first place, seems like a win.

foxware00:
> > I would love this too very, very much. Not having support for test plans sadly blocks us from using XcodeGen.
> 
> @albertodebortoli not to say this can't be improved, but fwiw we utilize test plans with xcodegen. Because the ids generated are stable we are able to keep the test plans in source control and just add test targets manually. Not perfect, but if it unblocks you from starting to use xcodegen in the first place, seems like a win.

@namolnad are you able to shed some light on how you've acheive this?

hakkurishian:
@yonaskolb  @omares -->  are there any blockers that you could use help with pushing this forward? I think this is one of the major open issues of XcodeGen and especially if project files aren´t checked in the viable alternative is not scalable for big projects such as ours.

yonaskolb:
@omares have pulled your changes into the original PR, closing this one

PR #847, opened 2020-04-26
Add staticBinary linkType for Carthage dependency

d-date:
When we try to use binary only framework with Carthage with Cartfile as below,  XcodeGen generate only `"$(PROJECT_DIR)/Carthage/Build/iOS/Static"` in `FRAMEWORK_SEARCH_PATH`.

Binary only framework in Carthage will download binary at `"$(PROJECT_DIR)/Carthage/Build/iOS"` so we need to distinguish these.

Cartfile
```
binary "https://dl.google.com/dl/firebase/ios/carthage/FirebaseAnalyticsBinary.json"
binary "https://dl.google.com/dl/firebase/ios/carthage/FirebaseAuthBinary.json"
binary "https://dl.google.com/dl/firebase/ios/carthage/FirebaseCrashlyticsBinary.json"
binary "https://dl.google.com/dl/firebase/ios/carthage/FirebaseDynamicLinksBinary.json"
binary "https://dl.google.com/dl/firebase/ios/carthage/FirebaseInAppMessagingBinary.json"
binary "https://dl.google.com/dl/firebase/ios/carthage/FirebaseMessagingBinary.json"
binary "https://dl.google.com/dl/firebase/ios/carthage/FirebasePerformanceBinary.json"
binary "https://dl.google.com/dl/firebase/ios/carthage/FirebaseRemoteConfigBinary.json"
```

project.yml (Before)
```yaml
    dependencies:
      - carthage: FirebaseABTesting
        linkType: static
      - carthage: FirebaseAnalytics
        linkType: static
      - carthage: FirebaseAuth
        linkType: static
      - carthage: FirebaseCore
        linkType: static
      ...
```

project.yml (After)
```yaml
    dependencies:
      - carthage: FirebaseABTesting
        linkType: staticBinary
      - carthage: FirebaseAnalytics
        linkType: staticBinary
      - carthage: FirebaseAuth
        linkType: staticBinary
      - carthage: FirebaseCore
        linkType: staticBinary
      ...
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,8 @@
 - Improve speed of metadata parsing and dependency resolution. [#803](https://github.com/yonaskolb/XcodeGen/pull/803) @michaeleisel
 - Improve support for iOS sticker packs and add support for `launchAutomaticallySubstyle` to run schemes. [#824](https://github.com/yonaskolb/XcodeGen/pull/824) @scelis
 - Add --project-root option to generate command. [#828](https://github.com/yonaskolb/XcodeGen/pull/828) @ileitch
-- Add an ability to set an order of groups with `options.groupOrdering` [#480](https://github.com/yonaskolb/XcodeGen/pull/613) @Beniamiiin
+- Add an ability to set an order of groups with `options.groupOrdering` [#613](https://github.com/yonaskolb/XcodeGen/pull/613) @Beniamiiin
+- Add `staticBinary` linkType for Carthage dependency [#847](https://github.com/yonaskolb/XcodeGen/pull/847) @d-date
 
 #### Fixed
 - Fixed issue when linking and embeding static frameworks: they should be linked and NOT embed. [#820](https://github.com/yonaskolb/XcodeGen/pull/820) @acecilia
--- a/Docs/ProjectSpec.md
+++ b/Docs/ProjectSpec.md
@@ -423,7 +423,7 @@ This only applies to `framework` dependencies. Implicit framework dependencies a
 **Carthage Dependency**
 
 - [ ] **findFrameworks**: **Bool** - Whether to find Carthage frameworks automatically. Defaults to `true` .
-- [ ] **linkType**: **String** - Dependency link type. This value should be `dynamic` or `static`. Default value is `dynamic` .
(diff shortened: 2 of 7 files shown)


d-date:
@brentleyjones Thank you for reviewed my PR and I've modified change log also.

brentleyjones:
@d-date There is a test failure: https://github.com/yonaskolb/XcodeGen/pull/847/checks?check_run_id=637043045#step:6:80

d-date:
@brentleyjones I fixed test cases and CI has passed now.

brentleyjones:
Thanks again @d-date!

yonaskolb:
Hi @d-date. What was the issue you were running into exactly? When using such dependencies, you can set `embed` to false. Does that solve your issue?
```
dependencies:
      - carthage: FirebaseABTesting
         embed: false
```
I'm hesitant to overload the `linkType` term with the idea of a dependency being a binary. If anything we could maybe add a `binary: true`, which XcodeGen could potentially determine for you automatically? Thoughts? @brentleyjones?

brentleyjones:
I don’t like having to manually edit `embed`. Before we had explicit static support that worked, but ideally we have something else that sets the correct embed state.

I’m not married to this change, and `binary: true` does seem like a better orthogonal way to express this, especially since Carthage could choose to download them to a different location based on of it was a binary (they don’t currently).

d-date:
@yonaskolb We have problem that  `FRAMEWORK_SEARCH_PATH` is wrong on linking statically.

In console you can see:
```
ld: framework not found
```

We have a workaround for this problem.

```
  framework: ./Carthage/iOS/Build/XXX.framework
  embed: false
```

This is originally solution before introduced `linkType:` on Carthage dependency.
However, this solution asks users to understand what framework should be embedded.

`binary: true` is better than `linkType: staticBinary`, and default value should be false and only worked in `linkType: static`.

brentleyjones:
>only worked in linkType: static.
I think it should just not care about the link type then. Since regardless of link type a binary is always downloaded to the same location (same end result as what you are saying though).

@d-date Do you mind opening a PR to make this change?

yonaskolb:
@d-date, I'd like to make a release, but would like to get this fixed first. Do you know if you have time to make this update?

d-date:
@yonaskolb Sorry for late but I'll take times to fix on this week. Thanks.

yonaskolb:
I'm reverting this change, so I can do a release. Feel free to open another PR @d-date! 😄👍

d-date:
Thanks!

Issue #646, opened 2019-09-04
Expose a command that allows users to print out target dependency graph without generating project


Gricha:
I am curious what are your thoughts on that and how feasible it is with the way that XcodeGen is implemented.

I'd like to build some light-weight script-based tooling that would help me analyze some assumptions on our dependency graph. It is somewhat possible with reading .yml/.json files and walking the graph, but there are nuances that make it difficult (like targets that can use templates that I'd have to unwind myself and duplicate the already existing logic).

Would it be feasible to have a command that takes a project but instead of passing it to `generate()`, it would JSONify it and print it out?

yonaskolb:
Hi @Gricha. This is definitely feasible, and wouldn't be very hard. It would just need a new command like 
```
xcodegen dump --type json/yaml/graphviz
```
by default I guess it would output to stdout which you can pipe to something, but could also take a `--file file.json` argument

PR #852, opened 2020-04-29
Adding graphviz support in dump

jeffctown:
Hey there! 👋

## Motivation

I came here looking for the same thing mentioned in #646: a way to output a dependency graph from `xcodegen` project files

## Whats in this PR

I added support, like specified in the linked issue.  

The following will output a Graphviz to stdout:

`xcodegen dump --type graphviz`

And this will output to a file:

`xcodegen dump --type graphviz --file ~/Desktop/output.viz`

When the output is pasted into an [online graphviz tool](https://dreampuf.github.io/GraphvizOnline/) it can produce a target dependency graph based on xcodegen project files.

This was created from the Test Fixture project in `Tests/Fixtures/TestProject/`:

TestProjectDependencyGraph

## What Should We Discuss?

Well, I'm adding a dependency on a [graphviz package](https://github.com/SwiftDocOrg/GraphViz).  It is actively maintained.

Is this worth adding a dependency? 🤷‍♂️

### DIFF
--- a/Sources/XcodeGenKit/GraphVizGenerator.swift
+++ b/Sources/XcodeGenKit/GraphVizGenerator.swift
@@ -1,0 +1,27 @@
+import DOT
+import Foundation
+import GraphViz
+import ProjectSpec
+
+public class GraphVizGenerator {
+    
+    public init() {}
+    
+    public func generateModuleGraphViz(targets: [Target]) -> String {
+        return DOTEncoder().encode(generateGraph(targets: targets))
+    }
+    
+    func generateGraph(targets: [Target]) -> Graph {
+        var graph = Graph(directed: true)
+        
+        targets.forEach { target in
+            target.dependencies.forEach { dependency in
+                let from = Node(target.name)
+                let to = Node(dependency.reference)
+                let edge = Edge(from: from, to: to)
+                graph.append(edge)
+            }
+        }
+        return graph
+    }
+}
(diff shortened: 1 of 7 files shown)


jeffctown:
@yonaskolb - What a great question!  This got me thinking, and I decided to update the style to be as close to a UML dependency diagram as possible.  My hope is that by using a standard, the diagrams will be understandable by the broadest audience.

#### Changes Since Last Review:

1. Updated all Nodes' shape to be a box
2. Updated Edges to use dashed lines since they are dependencies
3. Updated dependency labels to include `<<external>>` for any dependency targets who's type is not `.target` (external to the project). 

#### Updated TestProject Diagram:

Graph

Sources/XcodeGenKit/GraphVizGenerator.swift
```diff
@@ -0,0 +1,53 @@
+import DOT
+import Foundation
+import GraphViz
+import ProjectSpec
+
+extension Dependency {
+    var graphVizName: String {
+        switch self.type {
+        case .bundle, .package, .sdk, .framework, .carthage:
+            return "<<external>>\\n\(reference)"
```
yonaskolb, line 10: How about including the name of the dependency type instead of external? Same as enum cases but strings
```
[carthage]
Alamofire
```

jeffctown: Updated [here](https://github.com/yonaskolb/XcodeGen/pull/852/commits/e32078dae8486069d4f3c30f0f932da52dc5e30e)

PR #853, opened 2020-04-30
Don't link transitive static carthage libraries

akkyie:
The current version of XcodeGen tends to link all static carthage libraries transitively.
That means it links static dependencies of an App target to its test target. 
https://github.com/yonaskolb/XcodeGen/blob/aaae84f883c31dc0a4050f56942ee99c1b109135/Sources/XcodeGenKit/PBXProjGenerator.swift#L910-L925

Images below illustrate this behavior, where SwiftyJSON is a static library linked to the app target, and swift-tagged is another one linked directly to the test target.

|App|Test|
|---|----|
|Screen Shot 2020-04-30 at 13 34 48|Screen Shot 2020-04-30 at 13 35 00|

I believe this is not how we expect it to work because SwiftyJSON binary is embedded into both app and test executable, it causes symbol collision when you run test.

I changed PBXProjGenerator to link static libraries only what the top-level target depends, but not transitive ones. The count of results of `CarthageDependencyResolver.dependencies(for:)` won't change, as I'd rather want to keep its completeness.

Dependencies of the test target after this change:
Screen Shot 2020-04-30 at 13 35 51

I just added SwiftyJSON and swift-tagged statically to the fixture project in 7e65bd7 first, to see the actual diff for the code changes in 1da2133 (and have no opinion about the choice of libraries; just lightweight ones.)


Tests/Fixtures/TestProject/Cartfile
```diff
@@ -1 +1,3 @@
-github "antitypical/Result"
\ No newline at end of file
+github "antitypical/Result"
+github "SwiftyJSON/SwiftyJSON"
+github "pointfreeco/swift-nonempty"
```
brentleyjones, line 3: Nit: Can you add a trailing newline?

akkyie:
@brentleyjones Thanks! Added newline and merged master. Found fixture generation becoming indeterministic, also fixed.

> Do we have similar issues with transitive static library Target or SwiftPM dependencies?

I now tried both cases and they doesn't look to have problems. The app in the fixture project already has a static dependency but it's test is fine, and checked the SPM case on my local.
Should I also add it to the fixture?

brentleyjones:
If you could strengthen the fixture that would be great, thanks.

akkyie:
@brentleyjones Added SPM fixture changes and test fixes 👍  79621eb 2dde630

brentleyjones:
Looks like we have some test failures.

akkyie:
Oops, it succeeded with the empty source directory only on my local. Added a source.

akkyie:
Sorry again, fixing a typo in the spec.

akkyie:
Looks fine now?

PR #860, opened 2020-05-16
Preserve ordering of scheme targets as defined in the yml

mrabiciu:
Scheme targets are currently being sorted alphabetically.
Target order matters because the [first entry is grabbed when creating a buildableReference](https://github.com/yonaskolb/XcodeGen/blob/master/Sources/XcodeGenKit/SchemeGenerator.swift#L167)

I ran into this issue when trying to create a scheme for a watch app
```
schemes:
  Watch:
    build:
      targets:
        WatchApp: all
        MainApp: all
```
xcodegen ends up defaulting to my MainApp as the Executable instead of my WatchApp because MainApp comes first alphabetically.


EricRabil:
Why was this closed?

PR #862, opened 2020-05-16
Bugfix: Ensure folder and wrapper directories are correctly set on generated project

KhaosT:
#826 introduced a regression that resulted the generated Xcode project not to include certain types of folder containers due to the hardcoded extension list.

This PR addressed the issue by removing the hardcoded extension list and replaced it with a UTI based check.


KhaosT:
Added the entry :)

brentleyjones:
Thanks again @KhaosT 😄

Issue #589, opened 2019-05-20
Scheme auto selects first build target as an executable to run.


mbanasiewicz:
👋 First of all, thank you for this amazing library! 
We are in the process of migrating our big ( ~30 targets ) project into `XcodeGen` 🎉 and we are loving it! 
But we've stumbled upon an issue 🙀 

In our old project, we had a scheme that would aggregate all unit tests, which when build - would build all the apps, and when tested - would run all the tests. We've managed to express it using `xcodegen` in the following way:

```
schemes:
  AllTests:
    build:
      parallelizeBuild: true
      buildImplicitDependencies: true
      targets:
        - Target 1
        - Target 2
        - Target 3
    test:
      targets:
        -  Target 1 tests
        -  Target 2 tests
        - ...
```

However, we've noticed that `AllTests` target has run executable set ( this is the issue ), and we do not want that.
We've dug into xcodegen a bit deeper and we've found this bit of logic that selects the first build target as `run` if it's executable.
https://github.com/yonaskolb/XcodeGen/blob/846bae8b9ff45bbf1ac527a95ca1cf890748ba9c/Sources/XcodeGenKit/SchemeGenerator.swift#L117

## Questions!
Question 1: Is there a way to express in the configuration file that we do not want `AllTests` to be runnable? 
Like that 👇 
<img width="416" alt="Screenshot 2019-05-20 at 23 18 02" src="https://user-images.githubusercontent.com/1093407/58052619-9b553500-7b55-11e9-9e7e-3fbfa1434154.png">
Question 2: If it's impossible and we were to make a PR with the fix, what is/was the reason for https://github.com/yonaskolb/XcodeGen/blob/846bae8b9ff45bbf1ac527a95ca1cf890748ba9c/Sources/XcodeGenKit/SchemeGenerator.swift#L117 this line of code? 
Could we introduce a flag on a scheme `runnable: true/false` that would drive this behaviour?

Once again, thanks for all the great work!

tomquist:
We have exactly the same problem and it would be great to have a solution for this.

edulpn:
`executable` could be a parameter on a Scheme Run Action. It could default to the first target, but you could also explicitly say that you want it to be a specific application target, to ask on launch or that it does not have an associated executable (None), which covers the `runnable: true/false` flag.

edulpn:
@yonaskolb If it's okay, I could look into it and sending a PR

brentleyjones:
[...]

PR #866, opened 2020-05-19
Added ability to name the executable target in schemes.

Elland:
Attempting to address #589. I only managed to gather a small understanding of the code base, so please let me know if there's a simpler/better way of doing this. For example, I don't like that I'm using a value from the `build` section to configure something in the `run` section.

It works by adding the `executable` key to the `build` subsection of a scheme:

```
schemes:
  MyScheme:
    build:
      executable: MyExecutable
      targets:
      ...
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,8 @@
 - Add an ability to set an order of groups with `options.groupOrdering` [#613](https://github.com/yonaskolb/XcodeGen/pull/613) @Beniamiiin
 - Add `staticBinary` linkType for Carthage dependency [#847](https://github.com/yonaskolb/XcodeGen/pull/847) @d-date
 - Add the ability to output a dependency graph in graphviz format [#852](https://github.com/yonaskolb/XcodeGen/pull/852) @jeffctown
-- Adds uncluttering the project manifest dumped to YAML from empty values [#858](https://github.com/yonaskolb/XcodeGen/pull/858) @paciej00 
-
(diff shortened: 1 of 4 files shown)


brentleyjones:
Can you add a changelog entry? Thanks.

Elland:
@brentleyjones ✅

brentleyjones:
Thanks @Elland!

yonaskolb:
Hi @Elland, thanks for the PR! I think this got merged a bit prematurely though. As you say this property makes more sense on `run` instead of `build`. Could you please make a new PR for this, and also update the docs at https://github.com/yonaskolb/XcodeGen/blob/master/Docs/ProjectSpec.md#scheme

brentleyjones:
My bad for not catching that 😬. Thanks @yonaskolb 😄.

Elland:
@yonaskolb yes, but I couldn't really figure out how to do it via `run` yet. I can try again when I get back to work next week. (It's a 4 day weekend here in Germany)

yonaskolb:
@Elland great thanks! Was there a specific issue you were hitting when adding it to Run? If I'm not mistaken the property should be able to be stored, decoded and read just the same from `Build's` sibling struct `Run` instead. But let me know if there's an issue with that 👍

Elland:
@yonaskolb, I think when I tried that I lacked a bit more understanding of the code, I think I'll manage the second time around :)

yonaskolb:
Thanks @Elland. I've also just added the new property to the docs https://github.com/yonaskolb/XcodeGen/commit/fd7a6d374ba1932d5b75bee5da9dd101435bd20b

PR #869, opened 2020-05-25
Added ability to name executable target in schemes `Run` configuration.

Elland:


pinda:
Does this also provide the option to set the executable to "Ask on Launch"?

Elland:
@pinda unfortunately it does not. Sadly I do not have the bandwidth to look into this this week.

pinda:
No worries, created my own separate PR

Sources/ProjectSpec/Scheme.swift
```diff
@@ -114,9 +114,11 @@ public struct Scheme: Equatable {
         public var launchAutomaticallySubstyle: String?
         public var debugEnabled: Bool
         public var simulateLocation: SimulateLocation?
+        public var executableName: String?
```
brentleyjones, line 117: I think we should remove the property from the `Build` struct as well.

Elland: ✅

brentleyjones: It looks like it's still there up on line 76?

Elland: How about now?

brentleyjones: Perfect

yonaskolb:
Thanks @Elland. I've also just added the new property to the docs https://github.com/yonaskolb/XcodeGen/commit/fd7a6d374ba1932d5b75bee5da9dd101435bd20b

Sources/ProjectSpec/Scheme.swift
paciej00: @brentleyjones  @Elland  When running tests on the CI I get the following error when I used `xcodegen` build from te latest `master` branch.

```
xcodebuild: error: The scheme 'MyTests' is not configured for Running.
--
08-Jun-2020 16:53:55 | The scheme 'MyTests' has nothing configured to build for Running and has no executable specified to Run. Edit the scheme to configure the Run action.
```

I believe the issue is caused by this change, because the previous release of xcodegen (0.15.1) works well for me.

I don't have an idea how to fix that. Do you now how to have the previous functionality working and changes that @Elland introduced?

PR #871, opened 2020-05-26
Feature: Ability to set executable to Ask to Launch

pinda:
For Share Extensions it allows the possibility to set the executable to Ask to Launch.

### DIFF
--- a/Sources/ProjectSpec/Scheme.swift
+++ b/Sources/ProjectSpec/Scheme.swift
@@ -111,6 +111,7 @@ public struct Scheme: Equatable {
         public var stopOnEveryMainThreadCheckerIssue: Bool
         public var language: String?
         public var region: String?
+        public var askForAppToLaunch: Bool?
         public var launchAutomaticallySubstyle: String?
         public var debugEnabled: Bool
         public var simulateLocation: SimulateLocation?
@@ -125,6 +126,7 @@ public struct Scheme: Equatable {
             stopOnEveryMainThreadCheckerIssue: Bool = stopOnEveryMainThreadCheckerIssueDefault,
             language: String? = nil,
             region: String? = nil,
+            askForAppToLaunch: Bool? = nil,
             launchAutomaticallySubstyle: String? = nil,
             debugEnabled: Bool = debugEnabledDefault,
             simulateLocation: SimulateLocation? = nil
@@ -138,6 +140,7 @@ public struct Scheme: Equatable {
             self.stopOnEveryMainThreadCheckerIssue = stopOnEveryMainThreadCheckerIssue
             self.language = language
             self.region = region
+            self.askForAppToLaunch = askForAppToLaunch
             self.launchAutomaticallySubstyle = launchAutomaticallySubstyle
             self.debugEnabled = debugEnabled
             self.simulateLocation = simulateLocation
@@ -364,6 +367,10 @@ extension Scheme.Run: JSONObjectConvertible {
         } else if let string: String = jsonDictionary.json(atKeyPath: "launchAutomaticallySubstyle") {
             launchAutomaticallySubstyle = string
         }
+        
+        if let askLaunch: Bool = jsonDictionary.json(atKeyPath: "askForAppToLaunch") {
+            askForAppToLaunch = askLaunch
+        }
     }
 }
 
@@ -377,6 +384,7 @@ extension Scheme.Run: JSONEncodable {
             "config": config,
             "language": language,
             "region": region,
+            "askForAppToLaunch": askForAppToLaunch,
             "launchAutomaticallySubstyle": launchAutomaticallySubstyle,
         ]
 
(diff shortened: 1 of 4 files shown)


brentleyjones:
>For Share Extensions it allows the possibility to set the executable to Ask to Launch.

For Share Extensions, is it valid for the scheme to be setup either way? If not, should instead default this based on the target type?

pinda:
It is valid to be set up both ways. If you want to run the share extension directly you had to manually change it to Ask to Launch, but its not a requirement to do so. IMO it would be better to make it the default but this way you have the ability without forcing others to default to it.

brentleyjones:
> IMO it would be better to make it the default

What does Xcode default to when creating a similar scheme?

pinda:
> > IMO it would be better to make it the default
> 
> What does Xcode default to when creating a similar scheme?

Defaults to the settings that I created, I can add it as a default as well. Can you point me in the direction where I can find the defaults?

Sources/XcodeGenKit/SchemeGenerator.swift
```diff
@@ -241,6 +241,7 @@ public class SchemeGenerator {
             macroExpansion: shouldExecuteOnLaunch ? nil : buildableReference,
             selectedDebuggerIdentifier: (scheme.run?.debugEnabled ?? Scheme.Run.debugEnabledDefault) ? XCScheme.defaultDebugger : "",
             selectedLauncherIdentifier: (scheme.run?.debugEnabled ?? Scheme.Run.debugEnabledDefault) ? XCScheme.defaultLauncher : "Xcode.IDEFoundation.Launcher.PosixSpawn",
+            askForAppToLaunch: scheme.run?.askForAppToLaunch,
```
brentleyjones, line 244: You would default here, using information from the rest of the target. If you have trouble with that, we can leave the existing behavior as the default for now.

pinda:
@brentleyjones when one of the build targets is an app extension it will default to these settings, but still have the ability to override them in the yaml as people see fit.

Sources/ProjectSpec/Scheme.swift
```diff
@@ -364,6 +367,10 @@ extension Scheme.Run: JSONObjectConvertible {
         } else if let string: String = jsonDictionary.json(atKeyPath: "launchAutomaticallySubstyle") {
             launchAutomaticallySubstyle = string
         }
+        
+        if let askLaunch: Bool = jsonDictionary.json(atKeyPath: "askForAppToLaunch") {
```
Elland, line 371: How about defaulting to `false`?

Sources/XcodeGenKit/SchemeGenerator.swift
```diff
@@ -173,7 +173,7 @@ public class SchemeGenerator {
 
         let buildableReference = buildActionEntries.first(where: { $0.buildableReference.blueprintName == schemeTarget?.name })?.buildableReference ?? buildActionEntries.first!.buildableReference
         let runnables = makeProductRunnables(for: schemeTarget, buildableReference: buildableReference)
-
+
```
Elland, line 176: I personally find it worth it to remove dead whitespace for cleaner diffs.

Sources/ProjectSpec/Scheme.swift
pinda: the key is by default non-existing instead of false

Sources/XcodeGenKit/SchemeGenerator.swift
```diff
@@ -233,6 +233,14 @@ public class SchemeGenerator {
             }
             locationScenarioReference = XCScheme.LocationScenarioReference(identifier: identifier, referenceType: referenceType.rawValue)
         }
+        
+        let appExtensionTarget = scheme.build.targets.compactMap { bt -> Target? in
```
brentleyjones, line 237: Should we instead just use `schemeTarget`?

pinda: in my case the share extension needs both the normal app and the extension set for build targets. When either one of the targets is the app extension I set it to askForAppToLaunch otherwise the key can be nil. Using the schemeTarget doesn't have the same result.

brentleyjones: Ahh, okay. Let's not default it at all then. If its not specified in the `yml` we will behave like we do now.

pinda: Sounds good to me, this way we don't enforce anything upon the user but let them take care of setting the right values in the `yml`.

brentleyjones:
@pinda This looks good. Can you add a changelog entry and a simple unit test as well? Thanks!

brentleyjones:
Thanks again @pinda!

Sources/ProjectSpec/Scheme.swift
yonaskolb: If this value were false, then the property would be nil. Is that what is intended?

yonaskolb:
Thanks for this @pinda. Could you please make a followup PR with an update to the project spec https://github.com/yonaskolb/XcodeGen/blob/master/Docs/ProjectSpec.md

@brentleyjones, it's something we should require when new properties are added 👍

PR #884, opened 2020-06-10
Update ProjectSpec.md

paciej00:
As stated in the PR https://github.com/yonaskolb/XcodeGen/pull/869/files#r438188955 the name of the field should be `executable`

PR #888, opened 2020-06-12
Update JSON Dictionary parser to look for 'executableName' instead of 'executable'

liamnichols:
Related: #869

In the linked PR, the new feature for defining the schemes run action Executable was added. In the ProjectSpec docs, it refers to this property in the yaml as `executableName` but I got stumped because it didn't seem to work.

After digging into the code, it looks like it's referenced everywhere as `executableName` but the JSON decoding was looking for a value under the `executable` key instead. 

I'm not sure what the intention was, but I've updated that key to match the docs in this PR so that it works as described... Or should it have been the other way round? Should I update the docs to mention `executable`? 

Here are the docs - https://github.com/yonaskolb/XcodeGen/blob/f61af8caa9b26ce47814ef6e32402ea99402c208/Docs/ProjectSpec.md#run-action


brentleyjones:
https://github.com/yonaskolb/XcodeGen/pull/884

yonaskolb:
Thanks for this @liamnichols, good catch. Personally I like `executable` better, so closing in favour of #884

liamnichols:
Oh perfect, thanks for that I missed that PR when looking

Issue #776, opened 2020-02-06
[Question] Generating Multiple Projects - Best Practices?


liamnichols:
The work with `projectReferences` is great, especially with the addition to embedding sub projects in #570! With the project I work on, we currently split our app project from our frameworks project so as a result I have two separate **project.yml** files that each are used to generate the project for development. 

Each time I checkout/pull, the regeneration flow is like this:

```
$ mint run yonaskolb/XcodeGen xcodegen --spec Frameworks/project.yml --project Frameworks/
$ mint run yonaskolb/XcodeGen xcodegen --spec App/project.yml --project App/
$ bundle exec pod install --clean-install
```

While this works, one thing that gets to me a little is the fact that I can't leverage the `--use-cache` argument along with `postGenCommand` to avoid regenerating the pods project (and workspace) in the event that there is no change.

I've seen a lot of recent tickets/discussions about people using project references so I'm wondering if there is a way that we could improve this kind of workflow to take advantage of caching capabilities? 

I've seen the old discussion about generating workspaces in #168 however it doesn't look like there has been any movement on that, and maybe its a bit more complex than I thought in my comment at the end of that thread. 

Another thing I was thinking was to potentially extend the `projectReferences` option even further with something like this:

```yml
projectReferences:
  Frameworks:
    path: Frameworks/Frameworks.xcodeproj
    spec: Frameworks/project.yml
```

In the event that a `spec` is declared, XcodeGen could use this as an indicator to generate reference projects also all in the same command so this would allow me to call just the following:

```
$ mint run yonaskolb/XcodeGen xcodegen --spec App/project.yml --project App/ --use-cache
```

And as a result, we could leverage the caching through all referenced specs and trigger the parent  `postGenCommand` if anything changes with either the target project or any of its references.

Maybe this is a bit much, or maybe there are other plans on the horizon however I'm keen to get involved in working on a solution for this kind of thing so I'd appreciate any feedback or ideas/pointers as to how I (and others in the same situation) can achieve such behavior. Thanks!

Issue #168, opened 2017-11-22
[RFC] Workspace generation


sbarow:
I wanted to open the discussion around workspace generation and dependencies - opening a new task for the issue for a more concrete discussion. 

To support making applications as modular as possible we would want a `app/project.yml` file to be able to specify a dependency on another `lib/project.yml` file. Looking at the way [buck](https://buckbuild.com) refers to dependencies in the BUCK file (bazel does something similar [bazel](https://bazel.build) ) they go with the approach of `//<relative path>:<project-name>`. Doing this would require users to generate their projects from the root of their source code and passing in the path to `XcodeGen`. 

As an example lets assume we have a folder structure like this.

```
src/
	apps/
		app_1/
			...
			dependencies:
				- //libs/libs_1
				- //libs/libs_3
			...
		app_2/
			...
			dependencies:
				- //libs/libs_2
			...
	libs/
		lib_1/
			...
			dependencies:
				- //libs/libs_2
			...
		lib_2/
			...
			...
		lib_3/
			...
			...
	Carthage/
		...
```

We could achieve this by introducing a new `project` dependency type which takes in a string prefixed with `//` and generating an `.xcodeproject` per `project.yml` and all of its dependencies. 
Running `XcodeGen //apps/app_1` would generate a workspace for `app_1` at `apps/app_1/App_1.xcworkspace` as well as `.xcproject` in the desired project folder and all dependency folders. The workspace structure would mimic the folder structure defined in the directory. 
* `.xcodeproject` - having nested projects might break things or make it more complex. 
* Framework linking will be done through implicit dependencies in the `.xcworkspace`

Questions:
* Is this something `XcodeGen` should/wants to be able to do?
* Is this approach too opinionated?

rahul-malik:
@sbarow - This is something we want as well and it seems like handling the generation of multiple projects and managing their build dependencies should be in the scope of XcodeGen. We're using Bazel over at Pinterest so I'm a bit biased in favor of your exact implementation suggestion :)

- I imagined a workspace feature would largely be additions into the project specification (introduce a top-level `Workspace` object) that would be able to reference a series of file paths to other yaml files that contain full project descriptions. In addition it would probably contain other metadata around workspace name, settings, etc.
- Dependency targets could use the Buck/Bazel rule format or potentially something minimal like a tuple of `(Path, Dependency)`.
[...]

PR #891, opened 2020-06-14
Idea - Chaining spec generation using projectReferences

liamnichols:
👋 I've been spending some time recently working out a plan for modularising our project and have been looking into recent discussions in the issues about how XcodeGen might be able to further help with this. 

It mostly relates to the RFC/thread for workspace generation (#168) back in 2017, however since the original discussion, XcodeGen has grown quite a bit and now has support for things like subprojects using projectReferences that were introduced in #701. 

From my (somewhat limited) understanding of the project, the projectReferences seem like a potentially good candidate for expansion to allow the automatic generation of a number of Xcode Projects so I tried putting some feelers out there in #776 but didn't get much of a response. Since I had some time this weekend, I wanted to give a POC a go to better understand things.

In this draft PR, I've expanded `SpecLoader` to introduce a new method:

```swift
SpecLoader.loadProjects(path:projectRoot:variables:)
```

This method combined with the new optional `spec` property on a `ProjectReference` allows the loader to resolve all dependant specs in order to compute an array of `Project` objects that can then be passed into the generator to generate dependant projects in the correct order.

----

As an example, lets say that I have the following specs in my workspace:

**./Global/project.yml**
```yml
name: Global
projectReferences:
  Frameworks:
    path: ../Frameworks/Frameworks.xcodeproj
    spec: ../Frameworks/project.yml
  GlobalCore:
    path: ../GlobalCore/GlobalCore.xcodeproj
    spec: ../GlobalCore/project.yml
  Settings:
    path: ../GlobalFeatures/Settings/Settings.xcodeproj
    spec: ../GlobalFeatures/Settings/project.yml
# ...
```

**./Frameworks/project.yml**
```yml
name: Frameworks
# ...
```

**./GlobalCore/project.yml***
```yml
name: GlobalCore
projectReferences:
  Frameworks:
    path: ../Frameworks/Frameworks.xcodeproj
    spec: ../Frameworks/project.yml
# ... 
```

**./GlobalFeatures/Settings/project.yml**
```yml
name: Settings
projectReferences:
  Frameworks:
    path: ../Frameworks/Frameworks.xcodeproj
    spec: ../Frameworks/project.yml
  GlobalCore:
    path: ../GlobalCore/GlobalCore.xcodeproj
    spec: ../GlobalCore/project.yml
# ... 
```

I can now run the generator command on my entry project, I'll get the following output:

```
$ xcodegen generate --project ~./Global --spec ~./Global/project.yml 
Resolved projects in order:
1. ./Frameworks/Frameworks.xcodeproj
2. ./GlobalCore/GlobalCore.xcodeproj
3. ./GlobalFeatures/Settings/Settings.xcodeproj
4. ./Global/Global.xcodeproj
⚙️  Generating plists...
⚙️  Generating project...
⚙️  Writing project...
Created project at ./Frameworks/Frameworks.xcodeproj
⚙️  Generating plists...
⚙️  Generating project...
⚙️  Writing project...
Created project at ./GlobalCore/GlobalCore.xcodeproj
⚙️  Generating plists...
⚙️  Generating project...
⚙️  Writing project...
Created project at ./GlobalFeatures/Settings/Settings.xcodeproj
⚙️  Generating plists...
⚙️  Generating project...
⚙️  Writing project...
Created project at ./Global/Global.xcodeproj
```

As a result, I then end up with all four projects generated as I expect! This helps me to avoid having to maintain a wrapping script for XcodeGen that currently consists of me manually invoking the generator a given number of times since all of the specs are detected based on my `projectReferences` that I already have to define anyway 🚀 

---

So, while all of this works, it's far from being correct. I've identified a number of things that would probably need addressing before hand, but before we look into that, I want to get some thoughts on the concept in general 🙏 

Here are a few things to discuss:

1. `xcodegen` the CLI is very tailored to a specific project/spec since both `GenerateCommand` and `DumpCommand` inherit from `ProjectCommand`. Would this still make sense?
2. This approach works by running the generator a number multiple times to keep changes minimal, but if loaded all projects and generated in one go then we could benefit `dump` by being able to visualise the workspace/repo's dependencies cross-projects
3. Because `projectReferences` is keyed by the `name` that is used for reference in the spec, this has the potential to be different to the name of the actual project and this could get confusing but this can't be solved unless we load the referenced specs before resolving the project itself.
4. SpecLoader might need rewriting, I didn't get time to look at why it stores `project` but maybe that breaks things since there are now multiple. Might be better to investigate loading multiple `SpecFile` objects first and resolving them first? 
5. Having to match up `path` and `spec` in `ProjectReference` could lead to misconfigurations where the same spec points to different project paths, or different project paths use the same spec. We can validate against that but maybe there would also be a way to avoid this? 
6. Is building on top of `projectReferences` the right approach to this? By doing so, it means that you can use them both with existing xcodeproj references and also specs to be generated. I think that's a good thing but I might be overlooking something? 

I'm really keen to help out on a feature like this, but I think the idea still needs fleshing out a bit so any input on this would be great! Thanks!

### DIFF
--- a/Sources/ProjectSpec/ProjectReference.swift
+++ b/Sources/ProjectSpec/ProjectReference.swift
@@ -4,10 +4,12 @@ import JSONUtilities
 public struct ProjectReference: Hashable {
     public var name: String
     public var path: String
+    public var spec: String?
 
-    public init(name: String, path: String) {
+    public init(name: String, path: String, spec: String?) {
         self.name = name
         self.path = path
+        self.spec = spec
     }
 }
 
@@ -17,6 +19,7 @@ extension ProjectReference: PathContainer {
         [
             .dictionary([
                 .string("path"),
+                .string("spec")
             ]),
         ]
     }
@@ -26,6 +29,7 @@ extension ProjectReference: NamedJSONDictionaryConvertible {
     public init(name: String, jsonDictionary: JSONDictionary) throws {
         self.name = name
         self.path = try jsonDictionary.json(atKeyPath: "path")
+        self.spec = jsonDictionary.json(atKeyPath: "spec")
     }
 }
 
@@ -33,6 +37,7 @@ extension ProjectReference: JSONEncodable {
     public func toJSONValue() -> Any {
         [
             "path": path,
+            "spec": spec
         ]
     }
 }
--- a/Sources/ProjectSpec/SpecLoader.swift
+++ b/Sources/ProjectSpec/SpecLoader.swift
@@ -5,6 +5,19 @@ import XcodeProj
 import Yams
 import Version
 
+private extension Project {
+    func referencesSatisfied(with projects: [Project]) -> Bool {
+        // FIXME: Project.projectReferences do not contain enough information to assess this correclty all the time.
+        // This is because we don't know the references 'correct' output path until we load the spec (and read the real `name`).
+        // To overcome this, we should pass some data from `loadedProjects` that helps better resolve a spec path to a loaded project, that way we can be 100% sure...
+        // we'd also need to be a bit careful about cases where differnet specs reference the same spec to different output directories. This is something we could assess at laod time and error about.
+        let availableProjectPaths = projects.map({ $0.basePath.absolute() })
+        return projectReferences
+            .filter { $0.spec != nil }
+            .allSatisfy { availableProjectPaths.contains((basePath + $0.path).parent().absolute()) }
+    }
+}
+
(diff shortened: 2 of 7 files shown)


pavel-trafimuk:
Anew news for this request? cause we have the same issue
Thanks

Wei-Pinkoi:
Hi, 
Are there any updates?

sephoraios:
Any update about the solution ?

PR #892, opened 2020-06-15
Allow creating intermediary groups outside of the project directory

### LINKED ISSUE #889: createIntermediateGroups doesn't work for files outside of the project directory
With the following project yaml:

```
---
name: Foo
options:
    createIntermediateGroups: true
targets:
    Foo-Target:
        type: framework
        platform: iOS
        sources:
            - { path: "../Source/A.h" }
            - { path: "../Source/A.swift" }
            - { path: "../Example/Source/A.swift" }
            - { path: "../Example/Source/B.swift" }
            - { path: "../Example/Source/C.swift" }
```

Two different top-level groups get created name `Source`

Ideally, group nesting would continue to happen for sources outside of the project directory, since Xcode supports that -- having a parent group (say Foo in this case, if that's the name of `..`), or directly nesting `Source` & `Example` as toplevel groups, would help de-clutter the project and avoid the proliferation of thousands of groups named `Sources` (in our actual generated project)

segiddins:
Resolves #889 

Putting this out as a potential solution -- I'm new to the codebase, so I'm not sure if there's maybe a better way of doing this, or if there's a reason this is unacceptable. Just wanted to try it out, since it makes the projects we're generating via bazel in a scratch directory nicer to edit

### DIFF
--- a/Sources/XcodeGenKit/SourceGenerator.swift
+++ b/Sources/XcodeGenKit/SourceGenerator.swift
@@ -294,11 +294,15 @@ class SourceGenerator {
             // lives outside the project base path
             let isOutOfBasePath = !path.absolute().string.contains(project.basePath.absolute().string)
 
+            // whether the given path is a strict parent of the project base path
+            // e.g. foo/bar is a parent of foo/bar/baz, but not foo/baz
+            let isParentOfBasePath = isOutOfBasePath && ((try? path.isParent(of: project.basePath)) == true)
+
             // has no valid parent paths
-            let isRootPath = (isBaseGroup && isOutOfBasePath) || path.parent() == project.basePath
+            let isRootPath = (isBaseGroup && isOutOfBasePath && isParentOfBasePath) || path.parent() == project.basePath
 
             // is a top level group in the project
-            let isTopLevelGroup = !hasCustomParent && ((isBaseGroup && !createIntermediateGroups) || isRootPath)
+            let isTopLevelGroup = !hasCustomParent && ((isBaseGroup && !createIntermediateGroups) || isRootPath || isParentOfBasePath)
 
             let groupName = name ?? path.lastComponent
 
@@ -683,12 +687,26 @@ class SourceGenerator {
     private func createIntermediaGroups(for fileElement: PBXFileElement, at path: Path) {
 
         let parentPath = path.parent()
-        guard parentPath != project.basePath && path.string.contains(project.basePath.string) else {
-            // we've reached the top or are out of the root directory
+        guard parentPath != project.basePath else {
+            // we've reached the top
             return
         }
 
         let hasParentGroup = groupsByPath[parentPath] != nil
+        if !hasParentGroup {
+            do {
+                // if the path is a parent of the project base path (or if calculating that fails)
+                // do not create a parent group
+                // e.g. for project path foo/bar/baz
+                //  - create foo/baz
+                //  - create baz/
+                //  - do not create foo
+                let pathIsParentOfProject = try path.isParent(of: project.basePath)
+                if pathIsParentOfProject { return }
+            } catch {
+                return
+            }
+        }
         let parentGroup = getGroup(
             path: parentPath,
             mergingChildren: [fileElement],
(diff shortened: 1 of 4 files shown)


Sources/XcodeGenKit/SourceGenerator.swift
```diff
@@ -679,12 +680,20 @@ class SourceGenerator {
...
-            // we've reached the top or are out of the root directory
+        guard parentPath != project.basePath else {
+            // we've reached the top
             return
         }
 
         let hasParentGroup = groupsByPath[parentPath] != nil
+        if !hasParentGroup {
+            do {
+                let pathIsParentOfProject = try path.isParent(of: project.basePath)
```
yonaskolb, line 691: Could you document what this section is doing

Sources/XcodeGenKit/SourceGenerator.swift
```diff
@@ -301,12 +301,13 @@ class SourceGenerator {
 
             // lives outside the project base path
             let isOutOfBasePath = !path.absolute().string.contains(project.basePath.absolute().string)
+            let isParentOfBasePath = isOutOfBasePath && ((try? path.isParent(of: project.basePath)) == true)
```
yonaskolb, line 304: Could you document this property

Tests/XcodeGenKitTests/SourceGeneratorTests.swift
```diff
@@ -492,7 +492,7 @@ class SourceGeneratorTests: XCTestCase {
                 let pbxProj = try project.generatePbxProj()
                 try pbxProj.expectFile(paths: ["Sources", "A", "b.swift"], buildPhase: .sources)
                 try pbxProj.expectFile(paths: ["Sources", "F", "G", "h.swift"], buildPhase: .sources)
-                try pbxProj.expectFile(paths: ["../OtherDirectory/C/D", "e.swift"], names: ["D", "e.swift"], buildPhase: .sources)
+                try pbxProj.expectFile(paths: ["..", "OtherDirectory", "C", "D", "e.swift"], names: [".", "OtherDirectory", "C", "D", "e.swift"], buildPhase: .sources)
```
yonaskolb, line 495: Are there other test cases you can think of for this?

segiddins:
I'll try to pick this back up next week, thanks for the review!

Tests/XcodeGenKitTests/SourceGeneratorTests.swift
segiddins: not really

segiddins:
@yonaskolb this should be ready now, sorry for the delay!

segiddins:
@yonaskolb any other changes you'd like me to make?

Sources/XcodeGenKit/SourceGenerator.swift
```diff
@@ -294,11 +294,15 @@ class SourceGenerator {
             // lives outside the project base path
             let isOutOfBasePath = !path.absolute().string.contains(project.basePath.absolute().string)
 
+            // whether the given path is a strict parent of the project base path
+            // e.g. foo/bar is a parent of foo/bar/baz, but not foo/baz
+            let isParentOfBasePath = isOutOfBasePath && ((try? path.isParent(of: project.basePath)) == true)
+
             // has no valid parent paths
-            let isRootPath = (isBaseGroup && isOutOfBasePath) || path.parent() == project.basePath
+            let isRootPath = (isBaseGroup && isOutOfBasePath && isParentOfBasePath) || path.parent() == project.basePath
```
yonaskolb, line 302: I think this line can lose the `&& isParentOfBasePath`. 
`isOutOfBasePath` captures the intent and `isRootPath` is used in the line below. Or am I reading it wrong?

segiddins: I don't think it can. For example, for a project in ., ../foo/bar.txt shouldn't be considered the root path -- the root path should be a strict parent of the current directory

yonaskolb (approved):

segiddins:
Ping on merging this before the next release?

Issue #894, opened 2020-06-18
Playgrounds are no longer added to the project correctly


daltonclaybrook:
It looks like the 2.16.0 release may have introduced a regression where playgrounds are no longer being added to the Xcode project correctly. Instead, they are being added like any other folder, and the playground `Contents.swift` file is being added as a target member.

brentleyjones:
If I had to guess it would be from https://github.com/yonaskolb/XcodeGen/pull/862.

@KhaosT would you mind taking a look at it?

KhaosT:
Yeah sure.

@daltonclaybrook so in this case we are talking about a Xcode project with a Swift playground bundled in there?

daltonclaybrook:
@KhaosT Yah, that's right.

KhaosT:
Thanks.

So the underlying issue is that `.playground` has a UTI of `file.playground` when it's actually a folder, and that causes the check at [here](https://github.com/yonaskolb/XcodeGen/blob/379e0feb1d9730fc16af8b6e707fc4768122cfc7/Sources/XcodeGenKit/XCProjExtensions.swift#L69) to fail.

There are two possible fixes we can go with, one is to add this UTI to folder special case so it can be added correctly. The other one is to partially change the behavior introduced in #826 to only allow folders with extension that's not mapped to a known UTI.

@brentleyjones do you have a preference?

brentleyjones:
> The other one is to partially change the behavior introduced in #826 to only allow folders with extension that's not mapped to a known UTI.

That seems like the more wholistic change to me.

KhaosT:
👍 The PR is up for review ^^

KhaosT:
@daltonclaybrook quick question for you, when you add a Playground to Xcode, do you include it as part of a target? And if so, what do you expect when you build the app? Should the Swift playground get copied to the final target or not...?

daltonclaybrook:
@KhaosT We do want the playground to be brought into the Xcode project along with the sources for a target, but that playground should not be a member of the target and should not be copied into the target's bundle. That being said, if there is a [setting under Sources](https://github.com/yonaskolb/XcodeGen/blob/master/Docs/ProjectSpec.md#sources) that we can specify to get this behavior, I'm happy to use it.

KhaosT:
@daltonclaybrook Before you update to `2.16.0`, the Swift playground was not copied to the target right? Do you mind to share the segment of the configuration that included the playground?

[...]

PR #896, opened 2020-06-18
Treat all directories with known UTI as file wrapper

KhaosT:
This PR updated the logic that used to determine if we should apply special treatment for certain folders to mirror the behavior closer to what we had before #826. With this change, a folder will only be considered as a normal folder if it contains an extension that's not mapped to a known UTI.

This PR addressed the issue reported in #894.


Tests/Fixtures/TestProject/Project.xcodeproj/project.pbxproj
```diff
@@ -1736,6 +1740,14 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 		};
+		56F327C70680D6E4A222CB8A /* Resources */ = {
+			isa = PBXResourcesBuildPhase;
+			buildActionMask = 2147483647;
+			files = (
+				1A1225961174B8FB59FB139F /* MyPlayground.playground in Resources */,
```
brentleyjones, line 1747: I don't think we should copy playgrounds into the target's bundle. Not sure if that should get fixed here, or if it should get fixed in a separate PR (first).

KhaosT: It's a little unclear to me what's the desired behavior if someone decided to include the Swift playground on a target in `project.yml`... Let me check with daltonclaybrook and see what does he expect.

KhaosT: @brentleyjones yeah so verified that this particular behavior was not a regression, see [here](https://github.com/yonaskolb/XcodeGen/issues/894#issuecomment-646376190). 

We should probably track that separately around improving resource copy control to avoid unexpected resources getting copied to the target?

KhaosT:
Switched to using `fileGroups` instead of target so it doesn't get copied to the final destination.

brentleyjones:
@KhaosT Can you add a changelog entry?

KhaosT:
✅ Done~ Always forget that 😅

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@
 
 ## 2.16.0
 
+#### Fixed
+- Bugfix: Treat all directories with known UTI as file wrapper. [#896](https://github.com/yonaskolb/XcodeGen/pull/896) @KhaosT
```
brentleyjones, line 8: Can you move this up to the `## Next Version` section?

toshi0383:
Not sure if it's a bugfix though.

KhaosT:
Renamed title :)

Issue #898, opened 2020-06-22
Xcscheme for app extension requires `wasCreatedForAppExtension` to be 'YES'.


muizidn:
It seems that current XcodeGen doesn't support yet the capability to set .xcscheme for App Extension target for `wasCreatedForAppExtension` field. While the `tuist/Xcodeproj` has already supported it.

For my case, I need to implement notification service which it seems to work properly, we need to ensure that `wasCreatedForAppExtension` must be YES.

<img width="674" alt="Screen Shot 2020-06-22 at 13 49 40" src="https://user-images.githubusercontent.com/19355715/85256981-3aeaa280-b48f-11ea-89a7-ba763d5dfbca.png">

PR #899, opened 2020-06-22
Fix app extension scheme generation

muizidn:
This will fix issue #898.

Although this fix works in my case, I haven't yet updated the test so maybe need some more commits to ensure the new field be tested.

### DIFF
--- a/Sources/XcodeGenKit/SchemeGenerator.swift
+++ b/Sources/XcodeGenKit/SchemeGenerator.swift
@@ -50,7 +50,6 @@ public class SchemeGenerator {
 
         for target in project.targets {
             if let targetScheme = target.scheme {
-
                 if targetScheme.configVariants.isEmpty {
                     let schemeName = target.name
 
@@ -283,7 +282,9 @@ public class SchemeGenerator {
             launchAction: launchAction,
             profileAction: profileAction,
             analyzeAction: analyzeAction,
-            archiveAction: archiveAction
+            archiveAction: archiveAction,
+            wasCreatedForAppExtension: target
+                .flatMap({ $0.type.isExtension ? true : nil })
         )
     }
 
--- a/Sources/XcodeGenKit/SchemeGenerator.swift
+++ b/Sources/XcodeGenKit/SchemeGenerator.swift
@@ -283,8 +283,8 @@ public class SchemeGenerator {
             profileAction: profileAction,
             analyzeAction: analyzeAction,
             archiveAction: archiveAction,
-            wasCreatedForAppExtension: target
-                .flatMap({ $0.type.isExtension ? true : nil })
+            wasCreatedForAppExtension: schemeTarget
+                .flatMap { $0.type.isExtension ? true : nil }
         )
     }
 
--- a/Tests/Fixtures/TestProject/project.yml
+++ b/Tests/Fixtures/TestProject/project.yml
@@ -190,6 +190,7 @@ targets:
     sources: iMessageExtension
     settings:
       PRODUCT_BUNDLE_IDENTIFIER: com.project.iMessageApp.extension
+    scheme: {}
 
   iMessageStickersExtension:
     type: app-extension.messages-sticker-pack
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Change Log
 
 ## Next Version
+- Fixed issue when generating schemes for application extension should contain wasCreatedForAppExtension = YES. [#898](https://github.com/yonaskolb/XcodeGen/issues/898) @muizidn
 
 #### Fixed
 - Treat all directories with known UTI as file wrapper. [#896](https://github.com/yonaskolb/XcodeGen/pull/896) @KhaosT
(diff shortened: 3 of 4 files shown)


Sources/XcodeGenKit/SchemeGenerator.swift
```diff
@@ -321,6 +322,7 @@ extension Scheme {
     public init(name: String, target: Target, targetScheme: TargetScheme, project: Project, debugConfig: String, releaseConfig: String) {
         self.init(
             name: name,
+            wasCreatedForAppExtension: targetScheme.wasCreatedForAppExtension,
```
yonaskolb, line 325: The tests are failing because `wasCreatedForAppExtension = "NO"` is being added to the scheme where it's not expected. Xcode doesn't write this if it's false. This could be classified as a bug in XcodeProj, but to fix it here let's just use nil for false
```suggestion
            wasCreatedForAppExtension: targetScheme.wasCreatedForAppExtension ? targetScheme.wasCreatedForAppExtension : nil,
```

Sources/ProjectSpec/TargetScheme.swift
```diff
@@ -33,7 +35,8 @@ public struct TargetScheme: Equatable {
         commandLineArguments: [String: Bool] = [:],
         environmentVariables: [XCScheme.EnvironmentVariable] = [],
         preActions: [Scheme.ExecutionAction] = [],
-        postActions: [Scheme.ExecutionAction] = []
+        postActions: [Scheme.ExecutionAction] = [],
+        wasCreatedForAppExtension: Bool = false
```
yonaskolb, line 39: Let's use the default value here from above `wasCreatedForAppExtensionDefault`

Sources/ProjectSpec/Scheme.swift
```diff
@@ -570,6 +573,7 @@ extension Scheme: NamedJSONDictionaryConvertible {
 
     public init(name: String, jsonDictionary: JSONDictionary) throws {
         self.name = name
+        wasCreatedForAppExtension = jsonDictionary.json(atKeyPath: "wasCreatedForAppExtension") ?? false
```
yonaskolb, line 576: Can we add encoding to `toJSONValue` down below (not written if false)

Sources/XcodeGenKit/SchemeGenerator.swift
muizidn: ahh okayy. thanks

Sources/ProjectSpec/TargetScheme.swift
muizidn: okay.

brentleyjones:
@muizidn Can you put the changelog entry under `## Next Version`? Thanks!

muizidn:
Oh, my bad. Okay thanks!

Sources/XcodeGenKit/SchemeGenerator.swift
```diff
@@ -283,7 +283,8 @@ public class SchemeGenerator {
             launchAction: launchAction,
             profileAction: profileAction,
             analyzeAction: analyzeAction,
-            archiveAction: archiveAction
+            archiveAction: archiveAction,
+            wasCreatedForAppExtension: nil
```
yonaskolb, line 287: Sorry this was the bit that need the nil check. At the moment all this code doesn't do anything yet
```suggestion
            wasCreatedForAppExtension: scheme.wasCreatedForAppExtension ? scheme.wasCreatedForAppExtension : nil,
```
To test this, could you set a property on one of the schemes in TestProject fixture, and commit the diffs

muizidn: Can I add new target, e.g: NotificationService to Tests/Fixtures/TestProject/project.yml or I should use one of available extension target. For those target seems don't require wasCreatedForAppExtension to be YES.

yonaskolb: If there are no relevant app extensions, feel free to add one 👍

yonaskolb:
Thanks @muizidn. Now that I'm looking at the usage, I wonder if we can do this automatically rather than have a user setting. Do you know what sort of target require this? I wonder if `Target.type.isExtension` would suffice?

muizidn:
That's nicer!
I'll make research on it.

muizidn:
That's nicer!

So XcodeGen will infer the usage of this key based on it's target?

muizidn:
I got this. I'm working on this now.

Tests/Fixtures/TestProject/project.yml
```diff
@@ -190,6 +190,7 @@ targets:
     sources: iMessageExtension
     settings:
       PRODUCT_BUNDLE_IDENTIFIER: com.project.iMessageApp.extension
+    scheme: {}
```
muizidn, line 193: This one is required to force target generate scheme

Sources/XcodeGenKit/SchemeGenerator.swift
```diff
@@ -283,7 +282,9 @@ public class SchemeGenerator {
             launchAction: launchAction,
             profileAction: profileAction,
             analyzeAction: analyzeAction,
-            archiveAction: archiveAction
+            archiveAction: archiveAction,
+            wasCreatedForAppExtension: target
+                .flatMap({ $0.type.isExtension ? true : nil })
```
muizidn, line 287: `target?.type.isExtension` generates wasCreatedForAppExtension = "NO"

yonaskolb: I think we should use `schemeTarget`, and super small nitpick can we remove the parenthesis around flatMap 😄

muizidn:
hello!
can we merge this?

PR #908, opened 2020-07-14
Update XcodeProj to 7.13.0

brentleyjones:
This has support for App Clips.


yonaskolb:
Thanks @brentleyjones. Looks like some of the ids changed in the update, so you’ll just need to check in the generated fixtures to get the tests to pass

brentleyjones:
@yonaskolb Tests are passing now 🙂

PR #909, opened 2020-07-14
Add support for App Clips

brentleyjones:
Embeds App Clips into the containing app when they are a dependency.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 #### Added
 - Add `Scheme.Test.TestTarget.skipped` to allow skipping of an entire test target. [#916](https://github.com/yonaskolb/XcodeGen/pull/916) @codeman9
 - Added ability to set custom LLDBInit scripts for launch and test schemes [#929](https://github.com/yonaskolb/XcodeGen/pull/929) @polac24
+- Adds App Clip support. [#909](https://github.com/yonaskolb/XcodeGen/pull/909) @brentleyjones
 
 #### Fixed
 - Allow SDK dependencies to be embedded. [#922](https://github.com/yonaskolb/XcodeGen/pull/922) @k-thorat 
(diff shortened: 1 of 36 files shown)


brentleyjones:
@yonaskolb This is an Xcode 12 only feature. How do we want to go about testing this, since the figure will fail when tested with Xcode 11.

dflems:
@brentleyjones: Maybe `project.yml` could be templated (with mustache or some other template engine) to selectively ignore certain sections of the file unless a particular version of Xcode is selected?

dflems:
I fiddled with it a little locally. This might require a new generated project file :\

brentleyjones:
@yonaskolb FYI, we aren't actually testing this with CI because of the Spectre issue, but all of the infrastructure is in place for when we do add it.
```yaml
xcode: ["11"] # TODO: add 12_beta (blocked by https://github.com/kylef/Spectre/issues/41)
```

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 #### Added
 - Add `onlyCopyFilesOnInstall` option to targets for the Embed Files build phase. [#912](https://github.com/yonaskolb/XcodeGen/pull/912) @jsorge
+- Adds App Clip support. [#909](https://github.com/yonaskolb/XcodeGen/pull/909) @brentleyjones
```
dflems, line 7: @brentleyjones Can you tag me in on this as well? 😄

brentleyjones: Yep!

brentleyjones:
@yonaskolb This is green now.

dflems:
🙌 

Thanks @yonaskolb!

w.r.t the Spectre issue: in Xcode 12, `XCTest` now tracks with the latest version of macOS SDK but the SwiftPM default deployment target is `10.10` so test libraries that use `XCTest` get a little... weird. Gonna take another stab at trying to properly fix today.

dflems:
@yonaskolb @brentleyjones 

It looks like both Spectre and XcodeGen build fine on Xcode 12 beta 4 but GitHub Actions doesn't support it yet. They're currently targeting build `12A8169g`, which is beta 3 ([link](https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md)). Not sure when they're planning on supporting it, but once that's upgraded, we should be able to turn on the CI that will exercise these new codepaths.

Issue #608, opened 2019-07-15
Copy frameworks only when installing


Elland:
Is there a way to set this flag for a Copy Files phase?

yonaskolb:
Hi @Elland. What do you mean exactly by `Copy frameworks`?
At the time there isn't a way to control that on any of the build phases. This happens through a `buildActionMask` property on the build phase.
What is your use case?

Elland:
Hey @yonaskolb, I wanted the generated project configuration to only copy frameworks when installing, for faster build+launch times.

jsorge:
I'm running into this now, where I need this checkmark checked in my Embed Frameworks build phase:
<img width="1079" alt="Screen Shot 2020-03-29 at 9 58 04 PM" src="https://user-images.githubusercontent.com/2585841/77876637-694a4000-7208-11ea-8981-f536f6a3f298.png">

Here's what my project file looks like without the checkmark:
```
/* Begin PBXCopyFilesBuildPhase section */
		FC21B861D79C5E0F1CFC6599 /* Embed Frameworks */ = {
			isa = PBXCopyFilesBuildPhase;
			buildActionMask = 12;
			dstPath = "";
			dstSubfolderSpec = 10;
			files = (
				637EF3B657CC45F576440008 /* GooglePlaces.framework in Embed Frameworks */,
				9CB073B21DCB818D51414013 /* GoogleMaps.framework in Embed Frameworks */,
				84A3FAC8FC437CBCF002801A /* GoogleMapsCore.framework in Embed Frameworks */,
				238BC22DAEFDEA871E9C4988 /* GoogleMapsBase.framework in Embed Frameworks */,
			);
			name = "Embed Frameworks";
			runOnlyForDeploymentPostprocessing = 0;
		};
/* End PBXCopyFilesBuildPhase section */
```

And here's with the checkmark:
```
/* Begin PBXCopyFilesBuildPhase section */
		FC21B861D79C5E0F1CFC6599 /* Embed Frameworks */ = {
			isa = PBXCopyFilesBuildPhase;
			buildActionMask = 8;
			dstPath = "";
			dstSubfolderSpec = 10;
			files = (
				637EF3B657CC45F576440008 /* GooglePlaces.framework in Embed Frameworks */,
				9CB073B21DCB818D51414013 /* GoogleMaps.framework in Embed Frameworks */,
				84A3FAC8FC437CBCF002801A /* GoogleMapsCore.framework in Embed Frameworks */,
[...]

PR #912, opened 2020-07-21
Add Target option for copying frameworks when installing

jsorge:
There's a checkbox in a target's Embed Frameworks step that says "copy only when installing", and it defaults to un-selected. This PR adds the `onlyCopyFilesOnInstall` flag to `Target` which will check that box.

This resolves https://github.com/yonaskolb/XcodeGen/issues/608.


brentleyjones:
> I'm happy to add some tests around this feature, but I couldn't find the right spot in the project to add them.

A good start is the fixture tests, so we can confirm the actual project output.

jsorge:
@brentleyjones Added a test case to parse the JSON dictionary as well as the flag to a target in our yaml sample. Not sure where to assert that the right changes in the resulting project file occurred though.

Sources/ProjectSpec/Target.swift
```diff
@@ -46,6 +46,7 @@ public struct Target: ProjectTarget {
     public var deploymentTarget: Version?
     public var attributes: [String: Any]
     public var productName: String
+    public var onlyCopyFilesOnInstall: Bool
```
yonaskolb, line 49: Can you write this value in `toJSONValue` down below if the value is not false

yonaskolb:
Thanks @jsorge! It might be useful to also set a `Run script only when installing` on the Carthage build phase. Thoughts? 

In terms of the fixture test, I can see that it led to a diff and that's enough. That's used a simple integration and snapshot test. Uncommitted diffs and compilation errors cause a failure on CI.
If you'd like to add a unit test that would be great too. Then you can assert the exact values.

Could you also add a changelog entry.

jsorge:
@yonaskolb Thanks for the feedback! Addressed everything except the Carthage build phase, since I'm not sure how that works. My mental model is to have this mirror Xcode's checkbox (which is linked in issue #608). How does Carthage interact with the "Embed Frameworks" build phase?

yonaskolb:
@jsorge the generated Carthage build step basically extracts and copies the frameworks in, which is similar. I'm happy to leave this PR focused though. Thanks again!

Elland:
@jsorge thanks!

PR #913, opened 2020-07-21
Added support for SelectedTests in schemes `Test` configuration.

ooodin:


Docs/ProjectSpec.md
```diff
@@ -782,7 +782,9 @@ A multiline script can be written using the various YAML multiline methods, for
 - [x] **name**: **String** - The name of the target
 - [ ] **parallelizable**: **Bool** - Whether to run tests in parallel. Defaults to false
 - [ ] **randomExecutionOrder**: **Bool** - Whether to run tests in a random order. Defaults to false
+- [ ] **useTestSelectionWhitelist**: **Bool** - Whether to use a list of selected tests. Defaults to nil.
```
yonaskolb, line 785: Would you ever set `useTestSelectionWhitelist` to false while providing `selectedTests`? I wonder if we can just surface `selectedTests` and if it's not empty internally set `useTestSelectionWhitelist` on `XCScheme.TestableReference` generation

yonaskolb:
Thanks @ooodin. Could you add a changelog entry as well

Docs/ProjectSpec.md
ooodin: In case if `useTestSelectionWhitelist` is false or nil, `selectedTests` will be ignored. 
I thought to make a logic without using `useTestSelectionWhitelist`, but it seems too complicated and not so clear. For example, what if config provides both `skippedTests` and `selectedTests`, user can do this as well.

ooodin:
Hello @yonaskolb, I added a changelog entry

CHANGELOG.md
```diff
@@ -8,6 +8,11 @@
 #### Internal
 - Updated to XcodeProj 7.13.0 [#908](https://github.com/yonaskolb/XcodeGen/pull/908) @brentleyjones
 
+## 2.16.1
+
+#### Added
```
brentleyjones, line 13: This should probably be above `Fixed` in the `Next Version` section.

ooodin: @brentleyjones yeah, fixed)

yonaskolb:
Thanks for the PR @ooodin. I've updated this to use the presence of `selectedTests` to set `useTestSelectionWhitelist`, which I feel is an intuitive usage, and matches the ease of use of the Xcode UI

PR #915, opened 2020-07-28
Fix a bug in computing relative paths.

andrewreach:
Once a difference in path components between base and self has been
encountered, it is no longer valid to skip over common components.


yonaskolb:
Thanks @andrewreach. Could you add a changelog entry as well

andrewreach:
Sure, done.

PR #916, opened 2020-07-28
Add ability to skip the entire test target

codeman9:
Simply adds support for skipping the entire test target rather than always defaulting to `false`.

### DIFF
--- a/Sources/ProjectSpec/Scheme.swift
+++ b/Sources/ProjectSpec/Scheme.swift
@@ -500,6 +500,9 @@ extension Scheme.Test.TestTarget: JSONEncodable {
         if parallelizable != Scheme.Test.TestTarget.parallelizableDefault {
             dict["parallelizable"] = parallelizable
         }
+        if skipped {
+            dict["skipped"] = skipped
+        }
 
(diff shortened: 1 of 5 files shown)


Sources/ProjectSpec/Scheme.swift
```diff
@@ -474,6 +478,7 @@ extension Scheme.Test.TestTarget: JSONObjectConvertible {
         targetReference = try TargetReference(jsonDictionary.json(atKeyPath: "name"))
         randomExecutionOrder = jsonDictionary.json(atKeyPath: "randomExecutionOrder") ?? Scheme.Test.TestTarget.randomExecutionOrderDefault
         parallelizable = jsonDictionary.json(atKeyPath: "parallelizable") ?? Scheme.Test.TestTarget.parallelizableDefault
+        skipped = jsonDictionary.json(atKeyPath: "skipped") ?? false
```
yonaskolb, line 481: Can you also write the value in `toJSONValue` down below if the value is true

yonaskolb:
Not sure why the tests failed, might just be a random error

PR #917, opened 2020-07-29
Patch #909 to not fail CI

dflems:
* Pull app clips out of `TestProject/project.yml`
* Move to `TestProject/project-xcode-12.yml`
* Generate new fixtures
* Update build script to selectively build Xcode 12 project based on selected xcode version

Xcodegen itself can't build with Xcode 12 yet due to the Spectre dependency not building:

```
[44/69] Compiling JSONUtilities URL+JSONPrimitiveConvertible.swift
/Users/runner/work/XcodeGen/XcodeGen/.build/checkouts/Spectre/Sources/Spectre/XCTest.swift:35:14: error: 'recordFailure(withDescription:inFile:atLine:expected:)' is only available in macOS 10.13 or newer
    testCase.recordFailure(withDescription: "\(name): \(failure.reason)", inFile: failure.file, atLine: failure.line, expected: false)
```

### DIFF
--- a/Tests/Fixtures/TestProject/project-xcode-12.yml
+++ b/Tests/Fixtures/TestProject/project-xcode-12.yml
@@ -2,18 +2,9 @@
 # this file can be merged into project.yml. As-is, it generates
 # targets that Xcode 11 doesn't understand or know how to build.
 name: ProjectXcode12
-include: [environments.yml]
-options:
-  bundleIdPrefix: com.project
-  usesTabs: false
-  indentWidth: 2
-  tabWidth: 2
-  transitivelyLinkDependencies: true
-  groupSortPosition: top
-configFiles:
-  Test Debug: Configs/config.xcconfig
+include: [project.yml]
 targets:
-  App_iOS:
+  App_iOS_With_Clip:
     type: application
     platform: iOS
     attributes:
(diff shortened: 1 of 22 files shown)


Tests/Fixtures/TestProject/project-xcode-12.yml
```diff
@@ -0,0 +1,125 @@
+# NOTE: when Xcode 12 goes GM and the Xcode 11 CI can be dropped,
+# this file can be merged into project.yml. As-is, it generates
+# targets that Xcode 11 doesn't understand or know how to build.
```
brentleyjones, line 3: Could we instead have this file `include:` the normal one, and in here just have `App_Clip:` target?

brentleyjones: Well, and a version of the app that includes the clip.

dflems:
@brentleyjones Updated. Unfortunately, there's an issue with Spectre preventing Xcodegen from building with Xcode 12 at the moment, so I removed that entry from `ci.yaml`

Tests/Fixtures/TestProject/project-xcode-12.yml
dflems: Done

brentleyjones:
Thanks @dflems!

PR #918, opened 2020-07-30
Glob isDirectory cache

tcamin:
By running time profiler using Xcode 11.5 it turns out that a significant amount of time is spent getting/setting data to the cache, as shown below:

Screenshot 2020-07-30 at 15 31 05

Screenshot 2020-07-30 at 23 11 56

Screenshot 2020-07-30 at 23 12 59

Screenshot 2020-07-30 at 23 14 11

It is likely that this might be due to the fact that we have several include/exclude globs in our configuration yaml however a pretty simple fix is to replace the Dictionary with an NSCache or NSDictionary. Gotta be honest that I don't really understand what's going on here, but it seems that a significant amount of type is spent on checking equality (of keys?) as if there were lots of hash collisions going on internally. I tried increasing the capacity of the Dictionary and while things improved a bit they were very far from results obtained with NSCache or NSDictionary.

With NSDictionary generation went down from 16s to 4s. I'm profiling with release build configuration so optimizations are enabled.

Moreover, with our project I never got a cache hit unless I made the cache static so that previous evaluations are reused over multiple instances of Glob.

### DIFF
--- a/Sources/Core/Glob.swift
+++ b/Sources/Core/Glob.swift
@@ -57,7 +57,9 @@ public class Glob: Collection {
 
     public static let defaultBlacklistedDirectories = ["node_modules", "Pods"]
 
-    private var isDirectoryCache = [String: Bool]()
+    // NSDictionary showed much better get/set performances when compared to Swift native dictionary when
+    // instrumenting (Swift 5.2.4) the isDirectory(path:) method
+    private static let isDirectoryCache = NSMutableDictionary()
 
     public let behavior: Behavior
     public let blacklistedDirectories: [String]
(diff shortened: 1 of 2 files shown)


brentleyjones:
This is neat. Can you add a changelog entry? Thanks!

tcamin:
In a previous project I worked on I faced performance issues traversing the filesystem using NSFileManager, I ended using a Process invoking bash's find. What's your take on an approach like this one? https://github.com/yonaskolb/XcodeGen/compare/master...tcamin:exploredirectories_bashfind

It doesn't look nice but it significantly outperforms the NSFileManager based solution.

brentleyjones:
I would personally be fore it, since performance is a huge issue for me. @yonaskolb would have to weigh in on that though.

tcamin:
These are the 2 instrumentated sessions, with and without using NSFilemanager.

Screenshot 2020-07-31 at 15 18 43

Screenshot 2020-07-31 at 15 18 27

Again, maybe this is specific to our project but it would be nice if you could try it out to see if you get similar results.

brentleyjones:
Even if it only applies to a certain sort of project, that is a massive difference.

Sources/Core/Glob.swift
```diff
@@ -57,7 +57,7 @@ public class Glob: Collection {
 
     public static let defaultBlacklistedDirectories = ["node_modules", "Pods"]
 
-    private var isDirectoryCache = [String: Bool]()
+    private static let isDirectoryCache = NSCache<NSString, NSNumber>()
```
yonaskolb, line 60: Could you please add a comment mentioning that this is faster and why it's an NSCache

tcamin: I chose an NSCache because we're dealing with a cache here. It works with an NSDictionary as well, I can use whatever you think works best. Regarding the reason I can add a comment, although I can only mention that the finding is empirical using Instruments.

tcamin:
> I would personally be for it, since performance is a huge issue for me. @yonaskolb would have to weigh in on that though.

@yonaskolb what's your take on this alternative solution (https://github.com/yonaskolb/XcodeGen/compare/master...tcamin:exploredirectories_bashfind)?

Is there any large test repo available to see if this approach works for others as well?

PR #932, opened 2020-08-13
Improved Application Extension scheme generation

brentleyjones:
To better match Xcode generated schemes for Application Extensions:

- Defaults extensions to launchAutomaticallySubstyle "2"
- Doesn't use debugger launcher with extensions


yonaskolb:
Thanks @brentleyjones. There's quite a bit of logic in here, could we add some unit tests to `SchemeGeneratorTests`

brentleyjones:
@yonaskolb I'm on paternity leave for a while. I can add some tests when I get back, but that will probably be after Xcode 12 is released, and I think this and the subsequent WidgetKit PR would be good to land for that.

PR #935, opened 2020-08-14
Add better WidgetKit Extension support

brentleyjones:
This adds a new optional `subtype` property on `Target`. This is needed because Apple uses `app-extension` for WidgetKit Extensions, with only an entry in the Info.plist indicating which type of extension it is. By having a subtype declared, we are able to make a proper scheme for launching the extension, and provide proper defaults when using `info`.

This also adds the `hostTarget` property to `TargetScheme` and `Scheme.Run`. This allows specifying which target is hosting an extension or watch app. By specifying this it overrides the default logic in `TargetScheme` for selecting a host target to add to `Build` and it allows `SchemeGenerator` to add a `MacroExpansion` for WidgetKit Extensions.

### DIFF
--- a/Docs/ProjectSpec.md
+++ b/Docs/ProjectSpec.md
@@ -144,7 +144,7 @@ Describe an order of groups. Available parameters:
 
 ```yaml
 options:
-  groupOrdering: 
+  groupOrdering:
     - order: [Sources, Resources, Tests, Support files, Configurations]
     - pattern: '^.*Screen$'
       order: [View, Presenter, Interactor, Entities, Assembly]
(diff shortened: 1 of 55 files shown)


brentleyjones:
The scheme might need this section as well under `Launch`:

```xml
<MacroExpansion>
    <BuildableReference
        BuildableIdentifier = "primary"
        BlueprintIdentifier = "3B7AD31F24E1E1E5008832E1"
        BuildableName = "hosting_app.app"
        BlueprintName = "hosting_app"
        ReferencedContainer = "container:example.xcodeproj">
    </BuildableReference>
</MacroExpansion>
```

I'm investigating.

brentleyjones:
I added `hostTarget` to deal with the `MacroExpansion`.

Docs/ProjectSpec.md
```diff
@@ -783,6 +791,7 @@ A multiline script can be written using the various YAML multiline methods, for
 
 ### Run Action
 - [ ] **executable**: **String** - the name of the target to launch as an executable. Defaults to the first build target in the scheme
+- [ ] **hostTarget**: **String** - the name of the target that embeds this extension or watch app. If specified, the Launch action may be modified to support launching this target
```
brentleyjones, line 794: To match `executable`, should this be called `host` instead?

PR #936, opened 2020-08-18
Run CI against Xcode 12 beta

dflems:
Originally tried this in #909 / #917 but reverted because `Spectre` wouldn't build with Xcode 12b3. It _does_ build with b4, but GitHub Actions doesn't support it yet (incoming with https://github.com/actions/virtual-environments/pull/1409)

This PR will probably fail but once 12b4 is deployed on GH, it should pass on re-run 🤞 

@yonaskolb (cc @brentleyjones)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ jobs:
     name: Xcode ${{ matrix.xcode }}
     strategy:
       matrix:
-        xcode: ["11"]
+        xcode: ["11", "12_beta"]
     steps:
     - uses: actions/checkout@master
     - name: Set Xcode


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - Add `Scheme.Test.TestTarget.skipped` to allow skipping of an entire test target. [#916](https://github.com/yonaskolb/XcodeGen/pull/916) @codeman9
 - Added ability to set custom LLDBInit scripts for launch and test schemes [#929](https://github.com/yonaskolb/XcodeGen/pull/929) @polac24
 - Adds App Clip support. [#909](https://github.com/yonaskolb/XcodeGen/pull/909) @brentleyjones @dflems
```
yonaskolb, line 8: Can we put this into an `Internal` section

yonaskolb:
I saw that spectre has a new release for Xcode 12. We should probably update to that

CHANGELOG.md
dflems: updated

dflems:
@yonaskolb Unfortunately, it seems that Spectre `0.9.1` doesn't compile on beta 4 :(

Since GitHub is upgrading soon, it would work in the short term but then start failing once they roll up to beta 4.

```
/path/to/XcodeGen/.build/checkouts/Spectre/Sources/Spectre/XCTest@5.swift:11:21: error: 'XCTIssueReference' is not implicitly convertible to 'XCTIssue'; did you mean to use 'as' to explicitly convert?
    testCase.record(issue)
                    ^
                          as XCTIssue
```

dflems:
@yonaskolb Re-triggered. Looks like CI has `12A8179i` installed now so xcodegen _built_ fine, but now Carthage is failing when setting up the project fixtures -_-

Tests/Fixtures/TestProject/xcode12_workaround.xcconfig
```diff
@@ -0,0 +1,8 @@
+//
+// See https://github.com/Carthage/Carthage/issues/3019
+//
+// Skips building ARM slices for simulators until Carthage can support it
+//
+
+EXCLUDED_ARCHS__EFFECTIVE_PLATFORM_SUFFIX_simulator__NATIVE_ARCH_64_BIT_x86_64=arm64 arm64e armv7 armv7s armv6 armv8
+EXCLUDED_ARCHS=$(inherited) $(EXCLUDED_ARCHS__EFFECTIVE_PLATFORM_SUFFIX_$(EFFECTIVE_PLATFORM_SUFFIX)__NATIVE_ARCH_64_BIT_$(NATIVE_ARCH_64_BIT))
```
dflems, line 8: I don't think this will have any effect on Xcode versions before 12 (we use this same workaround internally)

Tests/Fixtures/TestProject/xcode12_workaround.xcconfig
```diff
@@ -0,0 +1,9 @@
+//
+// See https://github.com/Carthage/Carthage/issues/3019
+//
+// Skips building ARM slices for simulators until Carthage can support it
+//
+
+EXCLUDED_ARCHS__EFFECTIVE_PLATFORM_SUFFIX_simulator__NATIVE_ARCH_64_BIT_x86_64=arm64 arm64e armv7 armv7s armv6 armv8
+EXCLUDED_ARCHS_1200=$(inherited) $(EXCLUDED_ARCHS__EFFECTIVE_PLATFORM_SUFFIX_$(EFFECTIVE_PLATFORM_SUFFIX)__NATIVE_ARCH_64_BIT_$(NATIVE_ARCH_64_BIT))
+EXCLUDED_ARCHS=$(inherited) $(EXCLUDED_ARCHS_$(XCODE_VERSION_MAJOR))
```
dflems, line 9: This will only set `EXCLUDED_ARCHS` if building with Xcode 12. It works around an issue where Carthage fails to create fat Simulator + Device binaries. For instance, creation of fat dylibs/archives fail because `iphonesimulator` builds targeting Apple silicon share the `arm*` archs with `iphoneos`. We'll probably need this workaround until Carthage supports xcframeworks or some other means of being able to build frameworks per-platform.

dflems:
@yonaskolb Finally green after adding a workaround for Carthage :+1:

The fixtures including Xcode-12+-only targets are now successfully building in CI.

brentleyjones:
I heard that GitHub actions sometimes runs beta 4, but also sometimes runs beta 3 😕.

dflems:
Oof. Then maybe we hold off until it's confirmed in the GH actions docs.

dflems:
@yonaskolb @brentleyjones They merged https://github.com/actions/virtual-environments/pull/1409 today, so I would assume the rollout is 100% complete for beta4 on GitHub Actions.

Tests/Fixtures/TestProject/fixtures.xcconfig
```diff
@@ -0,0 +1,7 @@
+#include "xcode12_workaround.xcconfig"
+
+// Common settings for fixtures
+CODE_SIGN_IDENTITY =
+CODE_SIGNING_REQUIRED = NO
+CODE_SIGN_ENTITLEMENTS =
+CODE_SIGNING_ALLOWED = NO
```
yonaskolb, line 7: Anything not related to Xcode 12, can we put in the actual spec? I'd like to have that as buildable as possible without custom xcodebuild args

PR #942, opened 2020-08-27
Add support for specifying the UI testing screenshot behavior in a scheme test action

### LINKED ISSUE #941: Scheme Test Action: Capture screenshots automatically
Is there support to change the value of this toggle through the Scheme:test configuration?

Screen Shot 2020-08-26 at 12 41 30 PM

daltonclaybrook:
Resolves #941

### DIFF
--- a/Sources/ProjectSpec/Scheme.swift
+++ b/Sources/ProjectSpec/Scheme.swift
@@ -178,6 +178,8 @@ public struct Scheme: Equatable {
         public var region: String?
         public var debugEnabled: Bool
         public var customLLDBInit: String?
+        public var captureScreenshotsAutomatically: Bool
+        public var deleteScreenshotsWhenEachTestSucceeds: Bool
 
         public struct TestTarget: Equatable, ExpressibleByStringLiteral {
             public static let randomExecutionOrderDefault = false
@@ -236,7 +238,9 @@ public struct Scheme: Equatable {
             language: String? = nil,
             region: String? = nil,
             debugEnabled: Bool = debugEnabledDefault,
-            customLLDBInit: String? = nil
+            customLLDBInit: String? = nil,
+            captureScreenshotsAutomatically: Bool = true,
+            deleteScreenshotsWhenEachTestSucceeds: Bool = true
         ) {
             self.config = config
             self.gatherCoverageData = gatherCoverageData
@@ -251,6 +255,8 @@ public struct Scheme: Equatable {
             self.region = region
             self.debugEnabled = debugEnabled
             self.customLLDBInit = customLLDBInit
+            self.captureScreenshotsAutomatically = captureScreenshotsAutomatically
+            self.deleteScreenshotsWhenEachTestSucceeds = deleteScreenshotsWhenEachTestSucceeds
         }
 
         public var shouldUseLaunchSchemeArgsEnv: Bool {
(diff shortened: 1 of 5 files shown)


daltonclaybrook:
I'm guessing that the Xcode 12 job was recently added? The failure looks unrelated to this PR.

daltonclaybrook:
@brentleyjones Is there any appetite for merging this change soon? My team in particular is definitely looking forward to it.

brentleyjones:
I'm on paternity leave for a while.

@yonaskolb can you take a look?

Sources/XcodeGenKit/SchemeGenerator.swift
```diff
@@ -389,3 +390,16 @@ extension PBXProductType {
         }
     }
 }
+
+extension Scheme.Test {
+    var systemAttachmentLifetime: XCScheme.TestAction.AttachmentLifetime? {
```
yonaskolb, line 395: Seeing as though these 2 booleans end up as an enum, we could also just add a simple `attachmentLifetime` enum  to the spec instead. Thoughts?

daltonclaybrook: My thought was that the "attachment lifetime" is more of an implementation detail of the XCScheme rather than the Scheme in the project spec. From the user's perspective, the bools are a 1:1 mapping with these two checkboxes in the Xcode scheme editor. Happy to discuss this more, though.
Screen Shot 2020-09-06 at 12 07 11 PM

daltonclaybrook: @yonaskolb any more thoughts on this? I'm happy to make this change if you think it's the right call.

Sources/ProjectSpec/Scheme.swift
```diff
@@ -475,6 +481,8 @@ extension Scheme.Test: JSONObjectConvertible {
         region = jsonDictionary.json(atKeyPath: "region")
         debugEnabled = jsonDictionary.json(atKeyPath: "debugEnabled") ?? Scheme.Test.debugEnabledDefault
         customLLDBInit = jsonDictionary.json(atKeyPath: "customLLDBInit")
+        captureScreenshotsAutomatically = jsonDictionary.json(atKeyPath: "captureScreenshotsAutomatically") ?? true
+        deleteScreenshotsWhenEachTestSucceeds = jsonDictionary.json(atKeyPath: "deleteScreenshotsWhenEachTestSucceeds") ?? true
```
yonaskolb, line 485: Could you write these values below in `toJSONValue` if they differ from the default

Sources/ProjectSpec/Scheme.swift
```diff
@@ -498,6 +498,8 @@ extension Scheme.Test: JSONEncodable {
             "language": language,
             "region": region,
             "coverageTargets": coverageTargets.map { $0.reference },
+            "captureScreenshotsAutomatically": captureScreenshotsAutomatically,
+            "deleteScreenshotsWhenEachTestSucceeds": deleteScreenshotsWhenEachTestSucceeds,
```
yonaskolb, line 502: Thanks. Sorry if it wasn't clear before, but these values should only be written to the dictionary if they are not their default value (true), like below. This is so generated specs only contain the required information

daltonclaybrook: Ah, got it! This makes perfect sense. Will make the change right away.

yonaskolb (approved):
Thanks @daltonclaybrook!

Hally21 (commented):
👍

PR #945, opened 2020-09-07
Fixed adding Info.plists with custom names to Copy Bundle Resources build phase

anivaros:
Fixes warnings for custom named plists:

> warning: The Copy Bundle Resources build phase contains this target's Info.plist file 'HardwareSupport-Info.plist'. (in target 'HardwareSupport' from project 'LegacySupport')

- Extended check for Info.plist, because they usually named not just "Info.plist", but "<TargetName>-Info.plist"

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,8 @@
 #### Added
 - Add `useBaseInternationalization` to Project Spec Options to opt out of Base Internationalization. [#961](https://github.com/yonaskolb/XcodeGen/pull/961) @liamnichols
 
-#### Fixed
-- Adding Info.plists with custom names to Copy Bundle Resources build phase [#945](https://github.com/yonaskolb/XcodeGen/pull/945) @anivaros
+#### Changed
+- Info.plists with custom prefixes are no longer added to the Copy Bundle Resources build phase [#945](https://github.com/yonaskolb/XcodeGen/pull/945) @anivaros
 
 ## 2.18.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 #### Added
 - Add `useBaseInternationalization` to Project Spec Options to opt out of Base Internationalization. [#961](https://github.com/yonaskolb/XcodeGen/pull/961) @liamnichols
 
+#### Fixed
+- Adding Info.plists with custom names to Copy Bundle Resources build phase [#945](https://github.com/yonaskolb/XcodeGen/pull/945) @anivaros
+
 ## 2.18.0
 
 #### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 - Add `useBaseInternationalization` to Project Spec Options to opt out of Base Internationalization. [#961](https://github.com/yonaskolb/XcodeGen/pull/961) @liamnichols
 
 #### Changed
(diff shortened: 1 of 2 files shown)


yonaskolb (commented):
Thanks @anivaros.
I'm ok with this for now, but a proper fix might be to check the build settings of the target to find all the target's info plists and exclude them.
Could you add a changelog entry, and also rebase off or merge in master as the tests should be fixed there

anivaros:
> Could you add a changelog entry, and also rebase off or merge in master as the tests should be fixed there
Hello again! I've added changelog, thanks!

4brunu:
This PR introduced a behaviour change that makes the file `GoogleService-Info.plist` be recognised as the `info.plist` and it's not added to the `Copy Bundle Resources` build phase.
Is there any way to force the `GoogleService-Info.plist` file to be added to `Copy Bundle Resources` build phase?
Thanks

liamnichols:
I confirm the same issue. I'm able to work around it with the following addition to my spec (cc @4brunu):

```yml
    sources:
      - path: Global/SupportingFiles/GoogleService-Info.plist
        buildPhase: resources
```

I also noticed along the way that the docs are now also slightly incorrect:

> Note that `Info.plist` files will never be added to any build phases, no matter what this setting is. 

Since this change causes the behaviour to extend beyond just files named **Info.plist**. 

I had a dig through the code and it seems like there are a few ways that the `PBXProjGenerator` detects the `INFOPLIST_FILE` build setting:

https://github.com/yonaskolb/XcodeGen/blob/17b40a328889b3abca05dbee98598c1fec042116/Sources/XcodeGenKit/PBXProjGenerator.swift#L1170-L1181

1. Use the value defined explicitly in the `settings` for a target
2. Use `info.path` if it was specified
3. Search sources for a file named exactly **Info.plist**

This was never completely consistent with how the build rule was set since that had applied to any file named **Info.plist**, but now this is inconsistent in the way that the build rule applies to all files that have the suffix **Info.plist** (rather than a direct match). 

I guess, like @yonaskolb already mentioned, we probably need to instead work out exactly what source file is resolved to `INFOPLIST_FILE` and instead only set that file's build rule to `none` to completely align with the pbxproj generator logic. In the meantime though, it's good that the workaround above is able to unblock those who use **GoogleService-Info.plist**.

I had a look at figuring out how to pass in the resolved `INFOPLIST_FILE` reference into the source generator that seems impossible since this takes place after the sources are generated. I guess that instead we need to apply the build rule after but I need to dig a little further. Maybe a good start is getting some test coverage for all of the scenarios.. I might try and put up a Draft PR if I get a chance

4brunu:
Thanks, that workaround worked for me 👍

joshuapoq:
I think this change should be reverted. `GoogleServices-Info.plist` is going to be common amongst XcodeGen users.

Specifically declaring a workaround is not intuitive to how XcodeGen should work and there is nothing documented in the Project specs to clearly define this behaviour.

Agreed with @yonaskolb comment on doing this properly.

liamnichols:
I'm attempting to have a go at solving this in a way that works for both cases in #1027 btw. If it's something we can solve relatively soon then reverting might not be necessary.

In the meantime.. @anivaros, please could you let me know you configure your spec? In the `target`, do you set the `INFOPLIST_FILE` build setting in the `settings` property or do you define a `info` object with a `path` to your plist file?

PR #948, opened 2020-09-10
"Copy only when installing" for "Embed App Extensions"

### LINKED ISSUE #947: "Copy only when installing" for "Embed App Extensions" (NOT "Embed Frameworks") in "Build Phases"
We need to have the `Copy only when installing` checked for `Embed App Extensions` group in `Build Phases`. Do we have a way to do that? (I saw a couple of posts about the similar topic but they didn't help because they are for `Embed Frameworks` while we are asking for `Embed App Extensions`.)

Screen Shot 2020-09-10 at 1 25 36 PM

RomanPodymov:
Hello.
Thank you for XcodeGen.
In this pull request I fixed #947. Added `onlyCopyExtensionsOnInstall` and using it the same way as `onlyCopyFilesOnInstall` from #912 but for app extensions.
Also, I think `onlyCopyFilesOnInstall` should be renamed to `onlyCopyFrameworksOnInstall`.


yonaskolb:
Thanks for this @RomanPodymov. Would there be cases where `onlyCopyExtensionsOnInstall` would be a different value to `onlyCopyFilesOnInstall`? Otherwise we could just use the same property

RomanPodymov:
Hello @yonaskolb 
Well, I have a project where I should have selected **Embed App Extensions / Copy only when installing** and unselected **Embed Frameworks / Copy only when installing**.
That's why I think that `onlyCopyFilesOnInstall` is not a good name for this property, it should be something like `onlyCopyFrameworksOnInstall`.

RomanPodymov:
Hello @yonaskolb 
It works for me if I select both **Embed App Extensions / Copy only when installing** and **Embed Frameworks / Copy only when installing**. If you want I can use `onlyCopyFilesOnInstall` for all **Copy only when installing** checkboxes.

yonaskolb:
@RomanPodymov I think sticking with `onlyCopyFilesOnInstall` that controls all the copying is good for now. If down the line people need more discrete options we can look at adding `onlyCopyFrameworksOnInstall` and `onlyCopyExtensionsOnInstall`, but I don't think they shouldn't be necessary

RomanPodymov:
Hello @yonaskolb 
Now `onlyCopyFilesOnInstall` is using for all phases where possible.

yonaskolb (approved):

tkachenkovv:
@yonaskolb 
Hello.
Thank you for XcodeGen.
We REALLY need separate checkboxes for "Embed App Extensions" and "Embed Frameworks"

Issue #900, opened 2020-06-29
How to specify Platforms (`platformFilter`) for dependencies


ElfSundae:
![image](https://user-images.githubusercontent.com/526008/86011299-1a05ec80-ba4f-11ea-9022-5ebc77cded0b.png)

```
45B9DBB41F62A04002DAB9DB /* PBXTargetDependency */ = {
    isa = PBXTargetDependency;
    platformFilter = ios;
    target = 94E563A2FD80462D82C4E977 /* watchOS Example */;
    targetProxy = DAC46A191E3ACE32B698613E /* PBXContainerItemProxy */;
};
```

I'm trying to support Mac Catalyst for an iOS app, Xcode shows error:
```
Building for Mac Catalyst, but the embedded app 'watchOS Example.app' is building for watchOS. You may need to restrict the platforms for which this app should be embedded in the target editor.
```

So how to specify Platforms (`platformFilter`) for dependencies? Thanks.

FYI: https://developer.apple.com/documentation/xcode/creating_a_mac_version_of_your_ipad_app#3226802

PR #951, opened 2020-09-16
Add platform filter to dependencies

raptorxcz:
Solve issue #900.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Next Version
 
+- Added ability to set custom platform for dependency [#934](https://github.com/yonaskolb/XcodeGen/pull/934) @raptorxcz
+
 ## 2.22.0
 
 #### Added
--- a/Docs/ProjectSpec.md
+++ b/Docs/ProjectSpec.md
@@ -428,6 +428,7 @@ A dependency can be one of a 6 types:
 - [ ] **codeSign**: **Bool** - Whether the `codeSignOnCopy` setting is applied when embedding framework. Defaults to true
 - [ ] **removeHeaders**: **Bool** - Whether the `removeHeadersOnCopy` setting is applied when embedding the framework. Defaults to true
 - [ ] **weak**: **Bool** - Whether the `Weak` setting is applied when linking the framework. Defaults to false
+- [ ] **platform**: **String** - Add dependency to selected platforms. Available platforms are: **iOS**, **macOS** and **all**. Defaults is **all**
 
 **Implicit Framework options**:
 
--- a/Sources/ProjectSpec/Dependency.swift
+++ b/Sources/ProjectSpec/Dependency.swift
@@ -5,6 +5,7 @@ public struct Dependency: Equatable {
     public static let removeHeadersDefault = true
     public static let implicitDefault = false
     public static let weakLinkDefault = false
+    public static let platformDefault: Platform = .all
 
     public var type: DependencyType
     public var reference: String
@@ -14,6 +15,7 @@ public struct Dependency: Equatable {
     public var link: Bool?
     public var implicit: Bool = implicitDefault
     public var weakLink: Bool = weakLinkDefault
+    public var platform: Platform = platformDefault
 
     public init(
         type: DependencyType,
@@ -22,7 +24,8 @@ public struct Dependency: Equatable {
         codeSign: Bool? = nil,
         link: Bool? = nil,
         implicit: Bool = implicitDefault,
-        weakLink: Bool = weakLinkDefault
+        weakLink: Bool = weakLinkDefault,
+        platform: Platform = platformDefault
     ) {
         self.type = type
         self.reference = reference
@@ -31,6 +34,13 @@ public struct Dependency: Equatable {
         self.link = link
         self.implicit = implicit
         self.weakLink = weakLink
+        self.platform = platform
+    }
+    
+    public enum Platform: String, Equatable {
+        case all
+        case iOS
+        case macOS
     }
 
     public enum CarthageLinkType: String {
@@ -111,6 +121,12 @@ extension Dependency: JSONObjectConvertible {
         }
         if let bool: Bool = jsonDictionary.json(atKeyPath: "weak") {
             weakLink = bool
+        }
+        
+        if let platformString: String = jsonDictionary.json(atKeyPath: "platform"), let platform = Platform(rawValue: platformString) {
+            self.platform = platform
+        } else {
+            self.platform = .all
         }
     }
 }
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -336,7 +336,7 @@ public class PBXProjGenerator {
             return addObject(buildConfig)
         }
 
-        let dependencies = target.targets.map { generateTargetDependency(from: target.name, to: $0) }
+        let dependencies = target.targets.map { generateTargetDependency(from: target.name, to: $0, platform: nil) }
 
         let defaultConfigurationName = project.options.defaultConfig ?? project.configs.first?.name ?? ""
         let buildConfigList = addObject(XCConfigurationList(
@@ -352,7 +352,7 @@ public class PBXProjGenerator {
         aggregateTarget.dependencies = dependencies
     }
 
-    func generateTargetDependency(from: String, to target: String) -> PBXTargetDependency {
+    func generateTargetDependency(from: String, to target: String, platform: String?) -> PBXTargetDependency {
         guard let targetObject = targetObjects[target] ?? targetAggregateObjects[target] else {
             fatalError("Target dependency not found: from ( \(from) ) to ( \(target) )")
         }
@@ -368,6 +368,7 @@ public class PBXProjGenerator {
 
         let targetDependency = addObject(
             PBXTargetDependency(
+                platformFilter: platform,
                 target: targetObject,
                 targetProxy: targetProxy
             )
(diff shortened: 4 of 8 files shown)


davidrothera (approved):

raptorxcz:
Fixed conflict

yonaskolb:
Thanks for your work her @raptorxcz. A couple of points:
- lets rename `platformFilter` to just `platform` everywhere
- perhaps `all` is a more suitable name than `macOS+iOS`? What do you think, I can be persuaded otherwise

raptorxcz:
@yonaskolb thank for review. 

You have valid point. I updated PR.

raptorxcz:
Fixed merge conflict

Sources/ProjectSpec/Dependency.swift
```diff
@@ -31,6 +33,17 @@ public struct Dependency: Equatable {
...
         self.weakLink = weakLink
+        self.platform = platform
+    }
+    
+    public enum Platform: Equatable {
+        case all
+        case iOS
+        case mac
+        
+        public static func == (lhs: Self, rhs: Self) -> Bool {
```
yonaskolb, line 44: This equatable function isn't required, and in this case actually returns the wrong result

Sources/ProjectSpec/Dependency.swift
```diff
@@ -31,6 +33,17 @@ public struct Dependency: Equatable {
         self.link = link
         self.implicit = implicit
         self.weakLink = weakLink
+        self.platform = platform
+    }
+    
+    public enum Platform: Equatable {
+        case all
+        case iOS
+        case mac
```
yonaskolb, line 42: Should we use `macOS` to match the iOS and the platform names from the rest of Xcodegen

Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -1299,6 +1304,17 @@ public class PBXProjGenerator {
             targetObject.productType = target.type
         }
     }
+    
+    private func makePlatfor(for platform: Dependency.Platform) -> String? {
```
yonaskolb, line 1308: Platfor

Tests/Fixtures/TestProject/project.yml
```diff
@@ -113,12 +113,16 @@ targets:
       PRODUCT_BUNDLE_IDENTIFIER: com.project.app
     dependencies:
       - target: Framework_iOS
+        platform: mac+iOS
```
yonaskolb, line 116:
```suggestion
        platform: all
```

Tests/Fixtures/TestProject/project.yml
```diff
@@ -113,12 +113,16 @@ targets:
...
       - target: StaticLibrary_ObjC_iOS
       - carthage: Result
+        platform: mac
```
yonaskolb, line 119: if we change to macOS
```suggestion
        platform: macOS
```

Tests/ProjectSpecTests/SpecLoadingTests.swift
```diff
@@ -376,9 +376,9 @@ class SpecLoadingTests: XCTestCase {
             $0.it("parses target dependencies") {
                 var targetDictionary = validTarget
                 targetDictionary["dependencies"] = [
-                    ["target": "name", "embed": false],
-                    ["target": "project/name", "embed": false],
-                    ["carthage": "name", "findFrameworks": true],
+                    ["target": "name", "embed": false, "platform": "all"],
+                    ["target": "project/name", "embed": false, "platform": "mac"],
```
yonaskolb, line 380: If we change to macOS
```suggestion
                    ["target": "project/name", "embed": false, "platform": "macOS"],
```

Sources/ProjectSpec/Dependency.swift
raptorxcz: Updated in dfc471d8511f37738bfb9a84915d822bcc299f1e.

Sources/XcodeGenKit/PBXProjGenerator.swift
raptorxcz: Updated in dfc471d8511f37738bfb9a84915d822bcc299f1e.

Tests/Fixtures/TestProject/project.yml
raptorxcz: Updated in dfc471d8511f37738bfb9a84915d822bcc299f1e.

Tests/Fixtures/TestProject/project.yml
raptorxcz: Updated in dfc471d8511f37738bfb9a84915d822bcc299f1e.

Tests/ProjectSpecTests/SpecLoadingTests.swift
raptorxcz: Updated in dfc471d8511f37738bfb9a84915d822bcc299f1e.

Sources/ProjectSpec/Dependency.swift
yonaskolb: If you make Platform a RawRepresentable `Platform: String, Equatable` Swift can auto generate the Equatable conformance. This also makes the decoding code simpler as you can init with a string

raptorxcz: I refactored dependence equatable in a1ebf2240158f49f99a71ed3cac5a4a5c60308e4.

Tests/Fixtures/TestProject/Project.xcodeproj/project.pbxproj
```diff
@@ -507,6 +510,7 @@
 			files = (
 				535A98A3E3B74E09891D977F /* TestFramework.framework in Embed Frameworks */,
 				BD1419893577E6CEDF8CBA83 /* Result.framework in Embed Frameworks */,
+				BB06A57E259D0D2A001EA21F /* Result.framework in Embed Frameworks */,
```
raptorxcz, line 513: As you can see, new copy of Result dependency is added. This is result of fixing Platforms equatable. I tried to understand this behaviour, but I failed. I do not even know if is error or not.

jlott1:
> Thanks for your work her @raptorxcz. A couple of points:
> 
> * lets rename `platformFilter` to just `platforms` everywhere
> * perhaps `all` is a more suitable name than `macOS+iOS`, and easier to type? What do you think, I can be persuaded otherwise

@yonaskolb 
I've got a question.   Could this also be used to have platform specific dependencies, for example I need to install a framework dependency only for iOS but not include it for tvOS.   Is this possible with this change or can it be made possible?

yonaskolb:
That's a good question @jlott1. The work done here doesn't achieve that, this is more aimed at Catalyst, is that right @raptorxcz?
Being able to only add dependencies for a specific platform would be useful though for cross platform targets. I worry that people will think this is what that is. Thoughts @raptorxcz?

yonaskolb:
It would be great to have some unit tests around the generation of this as well

raptorxcz:
This PR add only support for platform option at Xcode.
Screen Shot 2021-02-15 at 9 16 26

raptorxcz:
Added some tests in 24e1ca0882a8b9a2000bfd14cfd065b207d0ddc6.

raptorxcz:
@yonaskolb Could you look at my PR?

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 - Added `basedOnDependencyAnalysis` to Project Spec Build Script to be able to choose not to skip the script. [#992](https://github.com/yonaskolb/XcodeGen/pull/992) @myihsan
 - Add `BuildRule.runOncePerArchitecture` to allow running build rules once per architecture. [#950](https://github.com/yonaskolb/XcodeGen/pull/950) @sascha
 - Adds discovered dependency file for a build script [#1012](https://github.com/yonaskolb/XcodeGen/pull/1012) @polac24 @fggeraissate
+- Added ability to set custom platform for dependency [#934](https://github.com/yonaskolb/XcodeGen/pull/934) @raptorxcz
```
freddi-kit, line 13: Please move it to `#Next Version` when solving conflicts

Sources/ProjectSpec/Dependency.swift
```diff
@@ -22,7 +23,8 @@ public struct Dependency: Equatable {
         codeSign: Bool? = nil,
         link: Bool? = nil,
         implicit: Bool = implicitDefault,
-        weakLink: Bool = weakLinkDefault
+        weakLink: Bool = weakLinkDefault,
+        platform: Platform = .all
```
freddi-kit, line 27: The default value is defined in top of this struct like below code

https://github.com/yonaskolb/XcodeGen/blob/24e1ca0882a8b9a2000bfd14cfd065b207d0ddc6/Sources/ProjectSpec/Dependency.swift#L5-L7

Could you help add like `platformDefault`?

Sources/ProjectSpec/Dependency.swift
```diff
@@ -31,6 +33,13 @@ public struct Dependency: Equatable {
         self.link = link
         self.implicit = implicit
         self.weakLink = weakLink
+        self.platform = platform
```
freddi-kit, line 36: Please apply `platformDefault` which mentioned above

Docs/ProjectSpec.md
```diff
@@ -428,6 +428,7 @@ A dependency can be one of a 6 types:
 - [ ] **codeSign**: **Bool** - Whether the `codeSignOnCopy` setting is applied when embedding framework. Defaults to true
 - [ ] **removeHeaders**: **Bool** - Whether the `removeHeadersOnCopy` setting is applied when embedding the framework. Defaults to true
 - [ ] **weak**: **Bool** - Whether the `Weak` setting is applied when linking the framework. Defaults to false
+- [ ] **platform**: **String** - Add dependency to selected platforms. Available platforms are: **iOS**, **macOS** and **all**. Defaults is **all**.
```
freddi-kit, line 431:
```suggestion
- [ ] **platform**: **String** - Add dependency to selected platforms. Available platforms are: **iOS**, **macOS** and **all**. Defaults is **all**
```

Sources/ProjectSpec/Dependency.swift
raptorxcz: Updated in 07df57d86dfd3172a9e49793ab04a9eb76e7414d.

Sources/ProjectSpec/Dependency.swift
raptorxcz: Updated in 07df57d86dfd3172a9e49793ab04a9eb76e7414d.

CHANGELOG.md
raptorxcz: Updated in 07df57d86dfd3172a9e49793ab04a9eb76e7414d.

raptorxcz:
Could you look at my PR? Thanks

yonaskolb:
Thanks @raptorxcz. Sorry for the delay on this

daltonclaybrook:
> That's a good question @jlott1. The work done here doesn't achieve that, this is more aimed at Catalyst, is that right @raptorxcz?
> Being able to only add dependencies for a specific platform would be useful though for cross platform targets. I worry that people will think this is what that is. Thoughts @raptorxcz?

@yonaskolb I am one such person who has found myself here because I thought this referred to platform-specific dependencies as mentioned by @jlott1. This is definitely a feature my team could benefit from. Is there a world where both of these features could exist? What would the spec need to look like to accommodate both?

PR #952, opened 2020-09-18
Fix appex's Runpath Search Paths under macOS target

rinsuki:
Unlike other operating systems, macOS app is not a flat bundle, so appex is in a deeper hierarchy than other OSes.

- macOS: `Host.app/Contents/PlugIns/Widget.appex/Contents/MacOS/Widget` (search path: `../../../../Frameworks`)
- Other OSes (e.g. iOS): `Host.app/PlugIns/Widget.appex/Widget` (search path: `../../Frameworks`)

so, we must be specify LD_RUNPATH_SEARCH_PATHS for macOS target, or dyld can't find Framework (and fail to launch).


yonaskolb:
Thanks @rinsuki. Could you add a changelog entry as well

rinsuki:
@yonaskolb Added!

yonaskolb (approved):

Issue #675, opened 2019-10-03
Incorrect scheme generation


kastiglione:
We have a case where we have a scheme with 3 targets: two static libraries and one iOS application. No matter what order these targets are listed in the scheme/build part of the spec, XcodeGen consistently picks the same static library as the runnable target of the `LaunchAction`. Either XcodeGen is sorting the targets by name, or putting them into a dictionary, and then picking the first one to be the runnable target.

Xcode silently accepts this scheme, but the result is the scheme can't be run.

kastiglione:
I will try to debug later, but wanted to report first.

PR #957, opened 2020-09-24
Select the first runnable target

codeman9:
Instead of just selecting the first target as the scheme target, search for and select the first runnable target. If there are no runnables found, then fallback to the first target of the scheme.

Fixes: #675


keith (approved):

yonaskolb (approved):
Great, thanks @codeman9!

PR #955, opened 2020-09-22
Enable Base Localization by default

liamnichols:
Closes #954.

So I'm still waiting for confirmation about the approach for the linked issue, but this PR is an implementation for my proposed solution based on how Xcode 12 now behaves. 

Let me know what you think 🙏

PR #961, opened 2020-10-02
Add useBaseInternationalization to SpecOptions

### LINKED ISSUE #954: Xcode 12 and Base Internationalization
👋 It's another year so time for a discussion about the handling of Base Internationalisation in Xcode 12. I'm trying to wrap my head around the history of this since it seems vast so to recap:

1. #680 - Original issue of _"Enable Base Internationalization"_ warning in Xcode 11
2. #685 - Fix for original issue that adds base localisation by default
3. #694 - Addendum to handle scenarios where there is no Base.lproj directory on disk
4. #791 - Reverts last amendment and instructs us to add a **.gitkeep** file instead.

Xcode 12 has now come along and I think we need to revisit things. My unlocalised framework project is now producing the dreaded warning:

Screenshot 2020-09-22 at 13 52 17

This warning now shows despite the fact that I have zero localisable content whatsoever.

I started a new framework project in Xcode 12 to check the default behaviour and discovered that both the development language and Base are included by default:

Screenshot 2020-09-22 at 13 53 17

I then unchecked the _"Enable Base Internationalization"_ box, removed `LastUpgradeCheck` and restarted Xcode to see the warning again so my gut feeling is that we should so back to an approach where we populate `knownRegions` with `Base`, `pbxProject.developmentRegion ?? "en"`, and `sourceGenerator.knownRegions`. 

Does that sound reasonable? Or am I missing something? I can go ahead and make these changes if so

liamnichols:
Closes #954 

So following a discussion in the linked issue, it was mentioned that it'll probably be easier just to provide a mechanism for opting back out of _"Use Base Internationalization"_ but to make it opt-in by default (how it was after #955). 

Do you think this does the job @yonaskolb?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Next Version
 
+#### Added
+- Add `useBaseInternationalization` to Project Spec Options to opt out of Base Internationalization. [#961](https://github.com/yonaskolb/XcodeGen/pull/961) @liamnichols
+
 ## 2.18.0
 
 #### Added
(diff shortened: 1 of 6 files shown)


yonaskolb (approved):
Great thanks @liamnichols!

Issue #944, opened 2020-09-02
Feature Request: Support StoreKit Testing Configuration file in Scheme


gemmakbarlow:
StoreKit testing will shortly be available in Xcode 12, which requires support for a `Configuration.storekit` file to be mapped into a scheme.

An overview of the upcoming functionality can be found [via this WWDC video](https://developer.apple.com/videos/play/wwdc2020/10659/).

More specifics available [via this Apple Documentation](https://developer.apple.com/documentation/xcode/setting_up_storekit_testing_in_xcode).

jcolicchio:
I'm going to take a look at implementing this, I'd also like to see XcodeGen handle this 👍

jcolicchio:
This'll require an update to the underlying XcodeProj dependency to support `StoreKitConfigurationFileReference`, however I'm running into an issue with Carthage and Xcode 12 when trying to build and tweak XcodeProj :\

gemmakbarlow:
@jcolicchio What issue are you seeing ? I'd be happy to take a look also.

jcolicchio:
The issue is that we have no way of specifying a scheme's Store Kit Configuration in yaml, so this field always resets on generating the project:

<img width="486" alt="Screen Shot 2020-10-08 at 3 56 58 PM" src="https://user-images.githubusercontent.com/2837288/95521676-f6700d80-097e-11eb-98d2-7cdbebb29c4b.png">

In order for XcodeGen to populate this field, we'll need XcodeProj to support encoding and decoding the `StoreKitConfigurationFileReference` child of the scheme's `LaunchAction`.

I've got an open PR for XcodeProj, which I think will make the appropriate change to enable us to implement this functionality in XcodeGen: https://github.com/tuist/XcodeProj/pull/573

jcolicchio:
The issue I was seeing with Carthage was related to Xcode 12 and the `lipo` command, but I found a fix and I also learned that I needed to run `swift package generate-xcodeproj` :P

jcolicchio:
Currently blocked by #965 due to the fact that `simulateLocation` and `storeKitConfiguration` both require a relative path that differs depending on whether the project is opened directly vs as part of a `.xcworkspace` :/

gemmakbarlow:
> The issue I was seeing with Carthage was related to Xcode 12 and the `lipo` command, but I found a fix and I also learned that I needed to run `swift package generate-xcodeproj` :P

Nice work figuring that out.

[...]

PR #964, opened 2020-10-09
Feature/Store Kit Configuration

jcolicchio:
Addresses issue #944 

This PR adds `storeKitConfiguration` as an optional property of the Run Scheme and TargetScheme

In the yaml we can pass the path to the `.storekit` as a string or a dictionary with `location` and optional `forWorkspace`.

`forWorkspace` is required in order to support usage with `xcworkspace`s


Sources/XcodeGenKit/SchemeGenerator.swift
```diff
@@ -188,7 +188,7 @@ public class SchemeGenerator {
                 parallelizable: testTarget.parallelizable,
                 randomExecutionOrdering: testTarget.randomExecutionOrder,
                 buildableReference: testBuilEntries.buildableReference,
-                skippedTests: testTarget.skippedTests.map(XCScheme.SkippedTest.init)
+                skippedTests: testTarget.skippedTests.map(XCScheme.TestItem.init)
```
jcolicchio, line 191: This change was required as part of the upgrade to `XcodeProj 7.15.0`

jcolicchio:
@yonaskolb Passing tests and ready for review!

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 #### Added
 - Add `useBaseInternationalization` to Project Spec Options to opt out of Base Internationalization. [#961](https://github.com/yonaskolb/XcodeGen/pull/961) @liamnichols
+- Add `Scheme.StoreKitConfiguration` to allow specifying StoreKit Configuration in Scheme and TargetScheme, supporting either xcodeproj and xcworkspace. [#964](https://github.com/yonaskolb/XcodeGen/pull/964) @jcolicchio
```
gemmakbarlow, line 7:
```suggestion
- Add `Scheme.StoreKitConfiguration` to allow specifying StoreKit Configuration in Scheme and TargetScheme, supporting either xcodeproj or xcworkspace. [#964](https://github.com/yonaskolb/XcodeGen/pull/964) @jcolicchio
```

Docs/ProjectSpec.md
```diff
@@ -833,6 +835,11 @@ targets:
 ```
 
 
+### StoreKitConfiguration
+- [x] **location**: **String** - path to `.storekit` configuration file
+- [ ] **forWorkspace**: **Bool** - whether the scheme will be used within a `.xcworkspace` rather than `.xcodeproj`
```
gemmakbarlow, line 840: Nice.

Sources/ProjectSpec/Scheme.swift
```diff
@@ -368,6 +412,8 @@ extension Scheme.Run: JSONObjectConvertible {
         region = jsonDictionary.json(atKeyPath: "region")
         debugEnabled = jsonDictionary.json(atKeyPath: "debugEnabled") ?? Scheme.Run.debugEnabledDefault
         simulateLocation = jsonDictionary.json(atKeyPath: "simulateLocation")
+        let maybeStoreKitLocation: String? = jsonDictionary.json(atKeyPath: "storeKitConfiguration")
+        storeKitConfiguration = jsonDictionary.json(atKeyPath: "storeKitConfiguration") ?? maybeStoreKitLocation.flatMap { Scheme.StoreKitConfiguration(location: $0) }
```
gemmakbarlow, line 416:
```suggestion
        storeKitConfiguration = maybeStoreKitLocation ?? maybeStoreKitLocation.flatMap { Scheme.StoreKitConfiguration(location: $0) }
```

Sources/ProjectSpec/TargetScheme.swift
```diff
@@ -68,6 +71,8 @@ extension TargetScheme: JSONObjectConvertible {
         }
         configVariants = jsonDictionary.json(atKeyPath: "configVariants") ?? []
         gatherCoverageData = jsonDictionary.json(atKeyPath: "gatherCoverageData") ?? TargetScheme.gatherCoverageDataDefault
+        let maybeStoreKitLocation: String? = jsonDictionary.json(atKeyPath: "storeKitConfiguration")
+        storeKitConfiguration = jsonDictionary.json(atKeyPath: "storeKitConfiguration") ?? maybeStoreKitLocation.flatMap { Scheme.StoreKitConfiguration(location: $0) }
```
gemmakbarlow, line 75:
```suggestion
        storeKitConfiguration = maybeStoreKitLocation ?? maybeStoreKitLocation.flatMap { Scheme.StoreKitConfiguration(location: $0) }
```

gemmakbarlow (approved):
Looks great ✅ 

Some very minor suggestions below, but none that should block this from being merged. See what you think !

CHANGELOG.md
jcolicchio: I can make this change, but I think it's an important distinction that a scheme can support one or the other, but not both simultaneously

jcolicchio: Oops I'm dumb, I read the diff backwards 😨

Sources/ProjectSpec/Scheme.swift
jcolicchio: Though it looks redundant, `jsonDictionary.json(atKeyPath: "storeKitConfiguration")` on L416 is actually attempting to decode "storeKitConfiguration" as a `Scheme.StoreKitConfiguration`, whereas on L415 it's attempting to decode it as a `String`. I've pushed a commit cleaning this up and moving it into `Scheme.StoreKitConfiguration`'s init, let me know what you think!

gemmakbarlow: Whoops, got it. Nice, this adjustment looks solid 👍🏻

Sources/ProjectSpec/Scheme.swift
```diff
@@ -58,6 +59,36 @@ public struct Scheme: Equatable {
...
+            self.location = location
+            self.forWorkspace = forWorkspace
+        }
+
+        public init?(parentJSONDictionary: JSONDictionary, keyPath: KeyPath = "storeKitConfiguration") {
+            if let storeKitConfiguration: StoreKitConfiguration = parentJSONDictionary.json(atKeyPath: keyPath) {
+                // attempt to decode as StoreKitConfiguration
+                self = storeKitConfiguration
+            } else if let storeKitLocation: String = parentJSONDictionary.json(atKeyPath: keyPath) {
+                // attempt to decode as String
```
gemmakbarlow, line 76: ❤️

jcolicchio:
@yonaskolb this is ready for review and needs one more reviewer if you'd like to do the honors. Moving forward, is there a process for PRs like this? Should I just tag you when something needs attention?

Sources/ProjectSpec/Scheme.swift
```diff
@@ -58,6 +59,36 @@ public struct Scheme: Equatable {
...
+                self = .init(location: storeKitLocation)
+            } else {
+                return nil
+            }
+        }
+
+        private var prefix: String {
+            return forWorkspace == true ? "../" : "../../"
```
yonaskolb, line 84: Are we sure the prefixes will never change or are other things in different scenarios? Maybe we should rename this to prefix, give it a default of "../../" and add info in the docs that it must be changed for workspaces. Thoughts?

yonaskolb: Personally I would have left has as a single `storekitConfiguration` string that required putting the prefix in and documented the usage of it. I'm hesitant to put too many smarts in XcodeGen for small side features, especially anything in Schemes as they change so much. But you've done the hard work here to make it more friendly already. If it's left in, it would be good to have some tests for the prefix changing with the boolean

yonaskolb: As XcodeGen gets older I have a greater appreciation for small features that don't need to updated when Xcode does 😄

jcolicchio: I absolutely see your point here, and I don't want to ship something that'll break with no recourse the next time an Xcode internal detail changes. How about this: instead of `forWorkspace`, I'll add `pathPrefix` with a default value of `../../`? `SimulateLocation` could also stand to benefit from this addition as it's currently broken for non-workspace projects...

jcolicchio: This way, the path can still be relative to the project root, same as other sources, but for `.xcworkspace` projects or when Xcode breaks, we can tweak as needed by setting `pathPrefix`? I'd just like for XcodeGen users not to have to worry about this weird wonky path detail until they actually need to

jcolicchio: @yonaskolb Ready for another round of review, does this look better?

jcolicchio:
@yonaskolb just merged master, assuming this still passes tests it could use some attention :0

jcolicchio:
@yonaskolb It seems like you're more active on GitHub now -- this PR has been open for a few months, can I get a quick re-review? Or is there someone else I can ping to get it reviewed and merged?

yonaskolb:
Hi @jcolicchio, apologies for the delay here. I've had another look through this, and have some follow up thoughts. In an effort to both simplify the `storekitConfiguration` and to make it easier to change the path for all `storekitConfiguration`s in a spec, what if the prefix was in the top level options, and `storekitConfiguration` was just a string. Then it could also be used for `SimulateLocation` as you say. Not sure of naming, `workspacePrefix`? Thoughts?
I promise this will get actioned immediately, and won't be left for ages again :)

jcolicchio:
> Hi @jcolicchio, apologies for the delay here. I've had another look through this, and have some follow up thoughts. In an effort to both simplify the `storekitConfiguration` and to make it easier to change the path for all `storekitConfiguration`s in a spec, what if the prefix was in the top level options, and `storekitConfiguration` was just a string. Then it could also be used for `SimulateLocation` as you say. Not sure of naming, `workspacePrefix`? Thoughts?
> I promise this will get actioned immediately, and won't be left for ages again :)

I think that's a great idea, and dovetails nicely into a similar fix affecting simulated locations. As for naming, I'll give it some thought. Ultimately, it's a prefix for relative paths within schemes, so maybe `schemePathPrefix` might be good?

jcolicchio:
@yonaskolb I've changed `storeKitConfiguration` to a string, added an option for path prefix (name negotiable and easily find/replaceable), updated project spec and tests, we're good to go if the name looks ok 👍

yonaskolb (approved):
Fantastic, great name and great work! 👏

Sources/ProjectSpec/FileType.swift
```diff
@@ -110,5 +110,6 @@ extension FileType {
         "xcfilelist": FileType(buildPhase: BuildPhaseSpec.none),
         "apns": FileType(buildPhase: BuildPhaseSpec.none),
         "pch": FileType(buildPhase: BuildPhaseSpec.none),
+        "storekit": FileType(buildPhase: BuildPhaseSpec.none),
```
liamnichols, line 113: 👋 @jcolicchio, please can you help me understand this addition quickly? 

When using a **.storekit** file in a scheme I think maybe makes sense that this wouldn't be included as a resource, however when updating to 2.19.0 today, my tested started failing because `SKTestSession(configurationFileNamed: "StoreKitConfiguration")` couldn't find my configuration file anymore. 

Could this have maybe been an oversight because to use the **.storekit** files with `SKTestSession` then they need to be bundled with the test target? 

Or is it that Xcode doesn't treat **.storekit** files as resources by default? I haven't actually tested adding one to Xcode manually to see how it behaves out of the box so maybe the prior behaviour was a side effect that I had been relying on. 

I can solve it in my spec file by overriding the `buildRule` in `sources`, but I'm wondering if maybe we want to refer this or thing actually 🤔

jcolicchio: Hey @liamnichols , sorry to hear my recent change has broken your setup! I'll look into how Xcode typically deals with *.storekit* files, it may indeed make sense for them to be copied by default. My use case involved referencing a storekit file in the scheme, which doesn't appear to require the *.storekit* file be copied. Apologies!

liamnichols: No stress, thanks for answering my question so quickly 🙌 

TBH I was also only just finding out that **.storekit** files could be used in two separate ways 😅 I only noticed that our project was doing this because our tests started failing when I upgraded XcodeGen 😬

jcolicchio: Hah all good, I guess we both learned something today :P

PR #976, opened 2020-10-27
Scheme config variants wrong assignment for similar config names

### LINKED ISSUE #975: Scheme config variants wrong assignment for similar config names
We've found out a strange bug when using multiple config names and generating schemes automatically. 
If config names are similar (example: `PROD` and `PREPROD`), the Prod scheme will be (wrongly) assigned to the `PREPROD` configuration.

#### Example `project.yml`:

```yml
name: Test
configs:
  DEV Debug: debug
  DEV Release: release
  PROD Debug: debug
  PROD Release: release
  PREPROD Debug: debug
  PREPROD Release: release    
settings:
      configs:
        DEV Debug:
           SWIFT_ACTIVE_COMPILATION_CONDITIONS: DEV DEBUG
        DEV Release:
           SWIFT_ACTIVE_COMPILATION_CONDITIONS: DEV
        PROD Debug:
           SWIFT_ACTIVE_COMPILATION_CONDITIONS: PROD DEBUG
        PROD Release:
           SWIFT_ACTIVE_COMPILATION_CONDITIONS: PROD
        PREPROD Debug:
           SWIFT_ACTIVE_COMPILATION_CONDITIONS: PREPRD DEBUG
        PREPROD Release:
           SWIFT_ACTIVE_COMPILATION_CONDITIONS: PREPRD
options:
  createIntermediateGroups: true
  deploymentTarget:
    iOS: "12.0"
targets:
  Test:
    type: application
    platform: iOS
    sources:
      - path: "Sources"
    settings:
        base:
            CURRENT_PROJECT_VERSION: 1
            MARKETING_VERSION: 0.26.0
            INFOPLIST_FILE: Sources/Info.plist
    scheme:
      configVariants:
        - DEV
        - PROD
        - PREPROD
```
#### Result
Opening the generated `Test PROD` scheme, there will be the `PREPROD Debug` and `PREPROD Release` configurations assigned.

stefanomondino:
This fixes #975 

To determine proper match between a config and a config variant, we should strip away the `Release`/`Debug` word from config name and check if the remaining part (trimmed out of whitespaces) is equal to the variant name.

Also, while investigating this and implementing proper tests, I found out a bug in usual tests with a mismatch between a `Prod` config variant and related `Debug Production` and `Release Production`, which is an error as per documentation. Tests were working only because `Debug Production` contains `Prod`, but it was just a coincidence.

This is why the `TestProject.swift` file was changed.


yonaskolb (commented):
Thanks for this @stefanomondino. Sorry for the delay on the review.

I think I can see a potential breaking change here if someone had a config name of `Production debug` for example, as the replacement you're using is case sensitive. Some people also don't put `debug` or `release` in their configs, and might have `Appstore` for example. This would be good test cases too.
Perhaps another approach would be to see if the whole word is in the config? Splitting by spaces, and then checking if that contains the variant?

Could you also please add a changelog entry?

stefanomondino:
Hey @yonaskolb, 

Thanks for the review (and for the huge amount of time you're saving me with this project <3 ) 

I'm not quite sure what you mean here, please correct me if I'm wrong.
`debug` and `release` are not part of the config name, they are the *type* of each config (should be some kind of default template provided by xcode itself). 
When using config variants, we're actually autogenerating schemes that follow a convention (the `debug` template of selected configuration for Run and Test, and the `release` one for archive).
If someone were using a direct config name like `Production debug` it would probably end up having two autogenerated variants, `Production debug debug` and `Production debug release`, which is... strange. 

Am I missing something?

Tests/XcodeGenKitTests/SchemeGeneratorTests.swift
```diff
@@ -182,32 +182,35 @@ class SchemeGeneratorTests: XCTestCase {
...
                 var target = app
                 target.scheme = TargetScheme(configVariants: configVariants)
                 let configs: [Config] = [
                     Config(name: "Test Debug", type: .debug),
-                    Config(name: "Production Debug", type: .debug),
+                    Config(name: "PREPROD Debug", type: .debug),
+                    Config(name: "PROD Debug", type: .debug),
                     Config(name: "Test Release", type: .release),
-                    Config(name: "Production Release", type: .release),
+                    Config(name: "PREPROD Release", type: .release),
```
yonaskolb, line 193: I mean if you were to change this to `Prepod release` I think you'd get a crash as the config can't be found with the variant name

yonaskolb:
Glad you’re getting some good use out of it! 😀
See my comment above. I also realised this lookup logic is duplicated in the spec validation, so it should be pulled out into somewhere common https://github.com/yonaskolb/XcodeGen/blob/0ac7a5f8c6652650d03d89c92da09ec74c6e6060/Sources/ProjectSpec/SpecValidation.swift#L95

stefanomondino:
Hey @yonaskolb I've pushed some changes as requested (extension over collection of `Config` objects) and updated the changelog, but still not getting your point about uppercase/lowercase... is this working for you now?

Tests/XcodeGenKitTests/SchemeGeneratorTests.swift
```diff
@@ -193,32 +193,35 @@ class SchemeGeneratorTests: XCTestCase {
...
                 var target = app
                 target.scheme = TargetScheme(configVariants: configVariants)
                 let configs: [Config] = [
                     Config(name: "Test Debug", type: .debug),
-                    Config(name: "Production Debug", type: .debug),
+                    Config(name: "PreProd Debug", type: .debug),
+                    Config(name: "Prod Debug", type: .debug),
                     Config(name: "Test Release", type: .release),
-                    Config(name: "Production Release", type: .release),
+                    Config(name: "PreProd Release", type: .release),
```
yonaskolb, line 204: I think this will fail the tests, and I'm sure it's not uncommon to have that lowercase
```suggestion
                    Config(name: "PreProd release", type: .release),
```

stefanomondino: @yonaskolb took me a while to get the point, then it totally made sense!

Should be fixed now as you originally suggested: whitespace tokenization and lowercase compare

Sources/ProjectSpec/Config.swift
```diff
@@ -10,10 +10,34 @@ public struct Config: Hashable {
...
     case debug
     case release
+    
+    public var name: String {
+        rawValue.prefix(1).uppercased() + rawValue.dropFirst()
+    }
+}
+
+public extension Collection where Element == Config {
+    func first(with configVariant: String, for type: ConfigType) -> Config? {
```
yonaskolb, line 26: Could you just add a comment saying `including` will be searched as a whole word
```suggestion
    func first(including configVariant: String, for type: ConfigType) -> Config? {
```

CHANGELOG.md
```diff
@@ -49,6 +49,9 @@
 
 [Commits](https://github.com/yonaskolb/XcodeGen/compare/2.18.0...2.19.0)
 
+#### Fixed
+- Scheme config variants wrong assignment for similar config names. [#976](https://github.com/yonaskolb/XcodeGen/pull/976) @stefanomondino
```
yonaskolb, line 53:
```suggestion
- Lookup scheme config variants by whole words, fixing incorrect assignment in names that contain subtrings of each other (eg PreProd and Prod) [#976](https://github.com/yonaskolb/XcodeGen/pull/976) @stefanomondino
```

Sources/ProjectSpec/Config.swift
```diff
@@ -10,10 +10,34 @@ public struct Config: Hashable {
...
+        first(where: { $0.type == type && $0.name.variantName(for: $0.type) == configVariant })
+    }
+}
+
+private extension String {
+    func variantName(for configType: ConfigType? ) -> String {
```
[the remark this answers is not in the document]
stefanomondino: @yonaskolb configType is declared as optional on `Config`, i've left that parameter as optional to avoid complex unwrap few lines above. I think it's best to leave it as it is, the method is private anyway.

CHANGELOG.md
```diff
@@ -49,6 +49,9 @@
 
 [Commits](https://github.com/yonaskolb/XcodeGen/compare/2.18.0...2.19.0)
 
+#### Fixed
+- Lookup scheme config variants by whole words, fixing incorrect assignment in names that contain subtrings of each other (eg PreProd and Prod) [#976](https://github.com/yonaskolb/XcodeGen/pull/976) @stefanomondino
```
yonaskolb, line 53: I'll move this on master

yonaskolb (approved):

PR #977, opened 2020-10-28
add support for default values in variable expansion

koenpunt:
This PR adds support for setting default values when using variable expansion, for both template variables and environment variables.

I wanted to use the default shell syntax for ["Parameter Substitution"](https://tldp.org/LDP/abs/html/parameter-substitution.html), so `${VARIABLE_NAME-default value}`, but I noticed that variables with dashes are also allowed. 
So I opted for the `:-` (`${VARIABLE_NAME:-default value}`) syntax, which for bash is described as; it will use the default value when the variable is absent, or the value is empty.


koenpunt:
Ping ✨

yonaskolb:
Thanks @koenpunt, this should be useful. Sorry for the delay. Could you please add a changelog entry as well. You'll need to rebase for that.

In terms of the syntax, another option is the one Xcode uses `${VALUE:default=VALUE_2}`. What do you think? Can't say I have a strong preference, but might be nice to align with the default tools.

PR #978, opened 2020-10-28
Refactor PBXProjGenerator and add tests around dependency order generation

asifmohd:
## Description

This PR refactors `PBXProjGenerator` and adds a few tests around dependency generation of static libraries and bundles

This should provide a good harness for the changes being done in PR #863. That is, these newly added tests should still pass when I rebase my branch with the changes from this PR.

I will be working on adding a few more tests in separate PRs before we can probably have the #863 change ready to be merged.

### DIFF
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -108,59 +108,20 @@ public class PBXProjGenerator {
 
         pbxProj.rootObject = pbxProject
 
-        for target in project.targets {
-            let targetObject: PBXTarget
-
-            if target.isLegacy {
-                targetObject = PBXLegacyTarget(
-                    name: target.name,
-                    buildToolPath: target.legacy?.toolPath,
-                    buildArgumentsString: target.legacy?.arguments,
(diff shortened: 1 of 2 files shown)


Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -108,59 +108,20 @@ public class PBXProjGenerator {
...
-                        lastKnownFileType: lastKnownFileType,
-                        path: target.filename,
-                        includeInIndex: false
-                    ),
-                    context: target.name
-                )
-
-                targetFileReferences[target.name] = fileReference
-            }
+        let generatedTargetResults = self.generateTargetInfo(targets: self.project.targets, aggregateTargets: self.project.aggregateTargets)
```
asifmohd, line 111: The motivation to move this logic to a different function was to break down the `generate` function into smaller functions.

And then the larger goal would be, to able to test out the `generateTargetInfo` function in isolation. But that requires marking the `generateTargetInfo` function as a public function so that it can be accessed in tests. 
It seems we could make this a pure function, i.e. pass in the targets to this function, and then the function returns this struct, which we can assert on, but still need to explore how feasible it would be to write tests around this function.

So I wanted to raise this PR and gather feedback if this approach makes sense, before proceeding further.

PR #988, opened 2020-11-21
Support Linux by upgrading XcodeProj and Spectre

elliottwilliams:
Hey there – I've been working upstream changes to [XcodeProj](https://github.com/tuist/XcodeProj/pull/585) and [Spectre](https://github.com/kylef/Spectre/pull/44) to get XcodeGen building and running on Linux hosts. Since both projects have made releases, I'm ready to move on to XcodeGen itself!

Linux support is obviously a little esoteric, but I think it's valuable for users that want to integrate XcodeGen into automated systems which are linux-based. For example: at Yelp, we have a ton of Linux CI capacity and lots of infra that makes it easy to run git-based workflows on linux servers, and being able to use XcodeGen from them would simplify things.

This branch passes tests on Ubuntu 18.04 using Docker and the [official swift 5.3 image](https://hub.docker.com/_/swift?tab=tags&page=1&name=5.3):

    docker run -itv $PWD:/code -u $(id -u $USER):$(id -g $USER) swift:5.3 \
    swift test --package-path /code --enable-test-discovery


Tests/CoreTests/GlobTests.swift
```diff
@@ -42,7 +42,7 @@ class GlobTests: XCTestCase {
     }
 
     private func newTmpDir() -> String {
-        var tmpDirTmpl = "/tmp/glob-test.XXXXX".cString(using: .utf8)!
+        var tmpDirTmpl = "/tmp/glob-test.XXXXXX".cString(using: .utf8)!
```
yonaskolb, line 45: Out of curiosity, what caused this change?

Tests/LinuxMain.swift
```diff
@@ -0,0 +1,17 @@
...
+import XcodeGenKitTests
+
+var tests = [XCTestCaseEntry]()
+tests += CoreTests.__allTests()
+tests += FixtureTests.__allTests()
+tests += PerformanceTests.__allTests()
+tests += ProjectSpecTests.__allTests()
+tests += XcodeGenKitTests.__allTests()
+
+XCTMain(tests)
```
yonaskolb, line 16: We could potentially replace the contents of this file with 
```
// LinuxMain.swift
fatalError("Run the tests with `swift test --enable-test-discovery`.")
```
And then delete all the XCTestManifests, as per https://oleb.net/2020/swift-test-discovery/

Sources/Core/Glob.swift
```diff
@@ -176,13 +176,13 @@ public class Glob: Collection {
...
                 }
-                let firstLevelPath = NSString(string: path).appendingPathComponent(subpath)
+                let firstLevelPath = NSString(string: path).appendingPathComponent(subpath).standardizingPath
                 guard isDirectory(path: firstLevelPath) else {
                     return nil
                 }
                 var subDirs: [String] = try FileManager.default.subpathsOfDirectory(atPath: firstLevelPath)
                     .compactMap { subpath -> String? in
-                        let fullPath = NSString(string: firstLevelPath).appendingPathComponent(subpath)
+                        let fullPath = NSString(string: firstLevelPath).appendingPathComponent(subpath).standardizingPath
```
yonaskolb, line 185: It looks like this is unavailable on macOS from the tests? `Foundation.StringProtocol:31:16: note: 'standardizingPath' has been explicitly marked unavailable here`

yonaskolb (commented):
This is fantastic, nice work across all 3 repos @elliottwilliams! 👏 
Could you please add a changelog entry as well, under `Added`

Sources/Core/Glob.swift
elliottwilliams: Yeah – seeing the same thing here. It might make sense to move some of the internal here over to URL to avoid Linux/Mac Foundation incompatibilities.

Tests/CoreTests/GlobTests.swift
elliottwilliams: This was a weird one! In glibc, `mkstemp` requires that [exactly six characters are `X`'s](https://man7.org/linux/man-pages/man3/mkstemp.3.html):

>  The last six characters of template must be "XXXXXX" and these are replaced with a string that makes the filename unique. Since it will be modified, template must not be a string constant, but should be        declared as a character array.

The BSD call only asks for "some number of `Xs' appended to it".

yonaskolb (approved):

elliottwilliams:
Thanks so much for the quick turnaround!

PR #992, opened 2020-12-02
Add baseOnDependencyAnalysis to Project Spec Build Script

myihsan:
From Xcode 12, here is a new option for Run Script Phase.

Screen Shot 2020-12-02 at 19 03 15

We can uncheck it to make sure the script runs every time.

This PR adds that option to Project Spec Build Script.


brentleyjones (approved):
Thanks for this @myihsan!

yonaskolb (approved):
Thanks @myihsan

PR #1003, opened 2021-01-08
Don't embed frameworks with MACH_O_TYPE=staticlib

mrabiciu:
There is a bug where if you setup a framework target with `MACH_O_TYPE=staticlib` it will incorrectly be added to the "Embed Frameworks" step resulting in duplicated frameworks and a large binary.

This change fixes an existing TODO created in: https://github.com/yonaskolb/XcodeGen/commit/0afd705 and correctly excludes these targets from the embed phase.


yonaskolb:
This is great, thanks @mrabiciu! Could you please add some tests and a changelog entry?

mrabiciu:
Sure will do!

mrabiciu:
@yonaskolb Added a test and changelog entry

Sources/ProjectSpec/XCProjExtensions.swift
```diff
@@ -54,17 +54,15 @@ extension PBXProductType {
...
-    public func shouldEmbed(_ dependencyType: PBXProductType) -> Bool {
-        switch dependencyType {
-        case .staticLibrary, .staticFramework:
-            // Some dependendencies should not be embed, independently of the target type
+    public func shouldEmbed(_ dependencyTarget: Target) -> Bool {
+        switch dependencyTarget.defaultLinkage {
+        case .static:
             return false
-
         default:
```
yonaskolb, line 61: Now that we don't have to enumerate the full type, can we make the cases explicit instead of using default

mrabiciu: Replaced `default` with explicit cases. I'm not sure if this reads any better or if I should do something like
```
if dependencyTarget.defaultLinkage == .static {
    // Static dependencies should never embed
    return false
} else if isApp {
    // If target is an app, all dependencies should be embed (unless they're static)
    return true
} else if isTest, [.framework, .bundle].contains(dependencyTarget.type) {
    // If target is test, some dependencies should be embed (depending on their type)
    return true
} else {
    // If none of the above, do not embed the dependency
    return false
}
```

yonaskolb (approved):

PR #1011, opened 2021-01-21
Readme improvements

literalpie:
I'm pretty sure this wording better communicates what you were trying to say


yonaskolb (approved):
Indeed, thank you!

PR #1012, opened 2021-01-25
Add Discovered Dependency File

polac24:
Adds support for build script discoveredDependencyFile, added to XcodeProj in tuist/XcodeProj#568

Screen Shot 2021-01-25 at 11 26 19


yonaskolb (approved):
Thanks @polac24

PR #1013, opened 2021-02-01
Allowing override LastUpgradeCheck and LastUpgradeVersion

Andre113:
Xcode will check for the value of `LastUpgradeCheck` and `LastUpgradeVersion` to display some warnings related to a recommended iOS deployment target. Just like this:

In some cases, it is not possible to change the deployment target because we still have to support some old iOS versions, and then the warnings stays there forever. This is really bad when we have a large number of targets.

With these changes, we can specify a different value for `LastUpgradeCheck` in `project.yml` . Like this:

```
name: MyApp

atributes:
  LastUpgradeCheck: '8.3'

configs:
  Debug: debug
  Homolog: debug
  Release: release
configFiles:
  Debug: Configs/DebugBase.xcconfig
  Homolog: Configs/HomologBase.xcconfig
  Release: Configs/ProdBase.xcconfig
```

This way, the warnings should be gone.


Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -285,9 +285,16 @@ public class PBXProjGenerator {
...
-            .merged(project.attributes)
-
+        var projectAttributes: [String: Any] = project.attributes
+        
+        // Set default LastUpgradeCheck if user did not specify
+        let lastUpgradeKey = "LastUpgradeCheck"
+        if !projectAttributes.contains(where: { (key, value) -> Bool in
+            key == lastUpgradeKey && value is String
+        }) {
+            projectAttributes[lastUpgradeKey] = project.xcodeVersion
```
yonaskolb, line 295: This code is the same as before, but it only replaces the key if it's a string, correct? Were you running into cases where devs were putting numbers in the project attributes?

yonaskolb:
Thanks @Andre113. Could you add a changelog entry as well

Sources/XcodeGenKit/PBXProjGenerator.swift
Andre113: Almost. It will replace the key if the user did not specify a value for `LastUpgradeCheck` or if this value is not a string. In this case, it will keep the same behavior as today, using the `project.xcodeVersion` value.
I  just felt it would be safer to validade if the value we're inserting is a `String` to avoid any strange values.

Andre113:
> Thanks @Andre113. Could you add a changelog entry as well

Sure! Just updated it. I also modified the readme to update the information regard the `LastUpgradeCheck` property.

yonaskolb:
Thanks @Andre113. We’ve had a release since you added your change log entry so could you move it up into the Next Version section

Andre113:
> Thanks @Andre113. We’ve had a release since you added your change log entry so could you move it up into the Next Version section

Oops.
Done, just updated it.

PR #1015, opened 2021-02-04
Change FRAMEWORK_SEARCH_PATH for xcframeworks

FranzBusch:
We ran into problems when switching from a regular `.framework` to an `.xcframework`. We had a static library target which depended on the `.framework` and after exchanging it for an `.xcframework` we got a compilation error that said `No such module XXX`. 

After investigating a bit, I figured out that due to the `FRAMEWORK_SEARCH_PATH` build setting that is set by XcodeGen for any framework dependency Xcode was not able to locate the proper file anymore. I think this is due to the setup of `.xcframework`s. They include multiple regular `.framework`s inside nested folders. Therefore, we need to pass a recursive path to `FRAMEWORK_SEARCH_PATH` so that it can find all the possible frameworks within the xcframework.

I adapted the code to handle this case, but I am not really sure if this is the proper approach. Would it make sense to create a new dependency type called `.xcframework` instead of special casing the `.framework` type?

It would be great to get your feedback @yonaskolb on this before I start writing tests.


yonaskolb:
Hi @FranzBusch, thanks for this! Yeah I'm happy with your approach, if you could add some tests and a changelog entry

FranzBusch:
@yonaskolb  I added some tests. Let me know if anything else is needed.

yonaskolb:
Thanks @FranzBusch. Could you add a changelog entry as well please. You'll need to rebase off master, as there has been a release since then

FranzBusch:
@yonaskolb done :)

Tests/XcodeGenKitTests/ProjectGeneratorTests.swift
```diff
@@ -1369,6 +1369,45 @@ class ProjectGeneratorTests: XCTestCase {
                 // generated plist should not be in buildsettings
                 try expect(targetConfig.buildSettings["INFOPLIST_FILE"] as? String) == predefinedPlistPath
             }
+            
+            describe("XCFramework dependencies") {
+                $0.context("with xcframework dependency") {
+                    $0.it("should add FRAMEWORK_SEARCH_PATHS_SEARCH_PATHS") {
```
yonaskolb, line 1375:
```suggestion
                    $0.it("should add FRAMEWORK_SEARCH_PATHS") {
```

Tests/XcodeGenKitTests/ProjectGeneratorTests.swift
```diff
@@ -1369,6 +1369,45 @@ class ProjectGeneratorTests: XCTestCase {
...
+                        let project = Project(name: "test", targets: [app])
+                        let pbxProject = try project.generatePbxProj()
+                        
+                        let target = pbxProject.nativeTargets.first!
+                        let configuration = target.buildConfigurationList!.buildConfigurations.first!
+                        try expect(configuration.buildSettings["FRAMEWORK_SEARCH_PATHS"] as? [String]) == ["$(inherited)", "\"some/folder/MyXCFramework.xcframework/**\""]
+                    }
+                }
+                $0.context("with regular framework") {
+                    $0.it("should add FRAMEWORK_SEARCH_PATHS_SEARCH_PATHS") {
```
yonaskolb, line 1393:
```suggestion
                    $0.it("should add FRAMEWORK_SEARCH_PATHS") {
```

PR #1020, opened 2021-02-15
Use carthage xcframeworks

yonaskolb:
This enables XcodeGen to look up the xcframeworks that Carthage 0.37 builds.
This PR is still a work in progress, but put up for reference

### DIFF
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -24,6 +24,7 @@ public class PBXProjGenerator {
     var packageReferences: [String: XCRemoteSwiftPackageReference] = [:]
 
     var carthageFrameworksByPlatform: [String: Set<PBXFileElement>] = [:]
+    var carthageXCFrameworks: Set<PBXFileElement> = []
     var frameworkFiles: [PBXFileElement] = []
     var bundleFiles: [PBXFileElement] = []
 
@@ -225,8 +226,8 @@ public class PBXProjGenerator {
         try project.targets.forEach(generateTarget)
         try project.aggregateTargets.forEach(generateAggregateTarget)
 
-        if !carthageFrameworksByPlatform.isEmpty {
-            var platforms: [PBXGroup] = []
+        if !carthageFrameworksByPlatform.isEmpty || !carthageXCFrameworks.isEmpty {
+            var children: [PBXFileElement] = []
             for (platform, files) in carthageFrameworksByPlatform {
                 let platformGroup: PBXGroup = addObject(
                     PBXGroup(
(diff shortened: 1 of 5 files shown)


Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -677,7 +681,7 @@ public class PBXProjGenerator {
 
         let targetSupportsDirectEmbed = !(target.platform.requiresSimulatorStripping &&
             (target.type.isApp || target.type == .watch2Extension))
-        let directlyEmbedCarthage = target.directlyEmbedCarthageDependencies ?? targetSupportsDirectEmbed
+        let directlyEmbedCarthage = target.directlyEmbedCarthageDependencies ?? project.options.carthageXCFrameworks || targetSupportsDirectEmbed
```
freddi-kit, line 684: ↓ may be corect?

```suggestion
        let directlyEmbedCarthage: Bool
            
        if let directlyEmbedCarthageDependencies = target.directlyEmbedCarthageDependencies {
            directlyEmbedCarthage = directlyEmbedCarthageDependencies
        } else if project.options.carthageXCFrameworks {
            directlyEmbedCarthage = project.options.carthageXCFrameworks
        } else {
            directlyEmbedCarthage = targetSupportsDirectEmbed
        }
```

This makes pass one failed test

freddi-kit (commented):
Review Ongoing

kateinoigakukun (commented):
IMHO we don't need to handle carthage dependency specially any more because XCFramework removed `carthage copy-framework` step.  We can add carthage dependency as a normal framework dependency like: `framework: Carthage/Build/Alamofire.xcframework`.

Yeah, but making things simple often leads some redundancy, so I understand the demand of shorthand.

But even if we provide a shorthand for `framework: Carthage/Build/Alamofire.xcframework`, we should separate API completely in the aspect of backward compatibility, I think.
IIUC, `carthage build` will produce XCFramework by default in the near future. So we may need to set `carthageXCFrameworks` true by default.

However, in that case, it will not be compatible with the old project.yml which uses `carthage: Alamofire` without explicit `carthageXCFrameworks` and doesn't use xcframeworks. If that project will update XcodeGen to the future version which enables `carthageXCFrameworks` by default, xcodeproj can't be generated without explicit `carthageXCFrameworks: false`.

On the other hand, if we won't make `carthageXCFrameworks` true by default, we have to write `carthageXCFrameworks: true` for the rest of our life.

So we need to re-think we should provide a shorthand or not, and the shorthand API should be separated from the current API or not.

CraigSiemens:
I tried this out with our project and I issue I'm noticing is that since carthage supports getting precompiled frameworks, it's possible to get a mix of xcframeworks and frameworks. This flag seems to assume everything will be available as a xcframework.

As far as I can tell, there isn't a great way to know whether a dependency will be a xcframework when the project is generated. The only ways I can think of would be:

- Use the currently added flag as a global default, allow each dependency to override that setting as needed with an `isXCFramework: {true|false}` or type option.
- At project generation time, look in the `Carthage/Build` folder to see whether it's a xcframework or a framework
  - May slow down the project generation a bit
  - Would require that the carthage dependencies are build before the project is generated, which isn't needed currently.

lsh-silpion:
Hi there, is there any progress here?

PR #1026, opened 2021-02-23
Bugfix/StoreKitConfiguration BuildPhase

### LINKED ISSUE #1025: [Regression] .storekit files are not considered bundled resources by default in 2.19.0
This follows my comment in https://github.com/yonaskolb/XcodeGen/pull/964/files#r581035944, I'm just opening it as a standalone ticket for a bit more visibility. 

The recent change in #964 causes **.storekit** files to not be defined a build rule by default so starting in 2.19.0 we have to explicitly declare the build rule of the source file if we want to use it in a Unit or UI Test target:

```yml
sources:
  - path: GlobalTests/Classes/Common/Utilities/Payments/StoreKitConfiguration.storekit
    buildPhase: resources
```

https://github.com/yonaskolb/XcodeGen/blob/17b40a328889b3abca05dbee98598c1fec042116/Sources/ProjectSpec/FileType.swift#L113

I see both sides to the story so I'm not completely sure what we want to do here:

- If the **.storekit** file is used purely for the StoreKit Configuration run option in a scheme, there is no need to bundle it as part of the target resources since Xcode loads it at runtime.
- If the **.storekit** file is however used by [`SKTestSession`](https://developer.apple.com/documentation/storekittest/sktestsession)'s `init(configurationFileNamed:) thows` method, then the file must be bundled as a resource to be loaded at runtime by the app (rather than Xcode). 

My gut feeling is that we should respect the previous behaviour and require that excluding the file from the bundle resources is an opt-in action, but I'm not 100% sure if I'm missing something or if this is inline with the strategy of similar scenarios?

jcolicchio:
This PR changes the default build phase for **.storekit** files from **.none** to **.resources**, since this is required for tests which want to load an arbitrary storekit configuration at runtime.

For use in schemes only, we can add the **.storekit** file via **fileGroups:** so that it won't be associated with any target, and won't be included in any bundle resources

Fixes #1025

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Next Version
 
+#### Fixed
+- Fixed regression on **.storekit** configuration files' default build phase. [#1026](https://github.com/yonaskolb/XcodeGen/pull/1026) @jcolicchio
+
 ## 2.19.0
 
 #### Added
--- a/Sources/ProjectSpec/FileType.swift
+++ b/Sources/ProjectSpec/FileType.swift
@@ -71,6 +71,7 @@ extension FileType {
         // resources
         "bundle": FileType(buildPhase: .resources),
         "xcassets": FileType(buildPhase: .resources),
+        "storekit": FileType(buildPhase: .resources),
 
         // sources
         "swift": FileType(buildPhase: .sources),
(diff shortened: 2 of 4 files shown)


jcolicchio:
@liamnichols Thanks for bringing this issue up, I'd appreciate your review on this PR -- If you could also install this version locally, remove your manual `buildPhase` override and validate that it fixes your issue, that'd be great!

liamnichols (approved):
Thanks @jcolicchio 💯 I checked out your branch and removed my workaround and I can confirm that this works just like it did in 2.18.0 🙇

jcolicchio:
> Thanks @jcolicchio 💯 I checked out your branch and removed my workaround and I can confirm that this works just like it did in 2.18.0 🙇

Awesome! @yonaskolb would you or someone else like to be the second reviewer? I think I saw @ainame liked the issue?

yonaskolb (approved):

PR #1027, opened 2021-02-23
Improve INFOPLIST_FILE handling to only omit used Info.plist's from Copy Bundle Resources Build Phase 

### LINKED ISSUE #1045: File not added to target membership in 2.19.0
Hello!
I have prepared a GoogleService-Info.plist file an Info.plist files for each target.
It worked fine up to 2.18.0, but after upgrading to 2.19.0, GoogleService-Info.plist and Info.plist are no longer added to my membership.
Need additional settings?
The folder structure and Xcodegen are as follows.
Xcodegen always help me. Thank you!🙏

```
...
Develop:
  type: application
  platform: iOS
  sources:
    - path: MyApp
    - Config/Develop
...
Staging:
  type: application
  platform: iOS
  sources:
    - path: MyApp
    - Config/Staging
...
Production:
  type: application
  platform: iOS
  sources:
    - path: MyApp
    - Config/Production
...
```

```
├── MyApp
│   ├── AppDelegate.swift
│   ├── ViewController.swift
│    ...
├── Config
│   ├── Develop
│   │   ├── GoogleService-Info.plist
│   │   └── Info.plist
│   ├── Staging
│   │   ├── GoogleService-Info.plist
│   │   └── Info.plist
│   └── Production
│       ├── GoogleService-Info.plist
│       └── Info.plist
├── project.yml
├── Mintfile
...
```

In 2.18.0

In 2.19.0

liamnichols:
Closes #1045

## Background

Prior to #945, a lot of XcodeGen was build assuming that a file named **Info.plist** was the value used in `INFOPLIST_FILE` build setting. This file is different to other plists because it shouldn't be included in the copy bundle resources build phase. 

If `INFOPLIST_FILE` was named anything other than **Info.plist**, by default, XcodeGen would include the file in the build phase which would result in a warning.

> warning: The Copy Bundle Resources build phase contains this target's Info.plist file 'HardwareSupport-Info.plist'. (in target 'HardwareSupport' from project 'LegacySupport')

#945 took a stab at fixing this by relaxing the checks so that any file that had been suffixed with **Info.plist** would then be treated as an **Info.plist** to support the old(?) pattern of using a file named `{TargetName}-Info.plist`. This however caused a noticeable regression for projects that include the **GoogleService-Info.plist** file which **should** be treated as a resource (this file is nothing to do with an bundles info property list and instead is a resource used by popular Google SDKs. 

So I looked back at @yonaskolb's initial comment:

>  but a proper fix might be to check the build settings of the target to find all the target's info plists and exclude them.

## Description

In this change, I am taking a stab at improving things for both scenarios. To begin with in 5a856b8, I checked out 2.18.0 and added a **GoogleService-Info.plist** file to the resources and regenerated the fixtures as expected. I then went back to 2.19.0 and also generated the fixtures with a target which defines the Info.plist as **App-Info.plist** in 
61e2c3b. 

Now that we have updated fixtures that reproduce both use cases, it should now be a matter of changing things until the tests pass 😬 

In `PBXProjGeneractor`, the final `INFOPLIST_FILE` build setting value for a given target isn't resolved until wayyyy after the array of `SourceFile` objects (which define the desired `buildPhase`) are created. I figured that the first task would be to resolve these values upfront so that we can use the actual `INFOPLIST_FILE` values to influence the `SourceFile`'s that we resolve (instead of guessing based on filename). So far, I've refactored the build setting resolution logic in 56ef412e6e87431a33ab71470a6ca130a1ac27c4 so that I now have a `[Config: String?]` dictionary that contains the `INFOPLIST_FILE` values for each build config of the target.

## What's Next

The next step is to somehow pass this information into `SourceGenerator.getAllSourceFiles(targetType:sources:)` so that it'll return a `SourceFile` with it's `buildPhase` value set to `BuildPhaseSpec.none` for the real Info.plist files (to make the tests pass). This is where I'm a little stuck. 

After trying to poke around, I felt like it was quite difficult to pass the set of files deep into the `SourceGenerator` and before I continue, I want to confirm that I'm on the right tracks.

Part of me did wonder if I could maybe instead just mutate the `target.sources` array before passing it into the `SourceGenerator` since I could just pretend that it was just set in the spec file, but I wondered if it was wrong to override the value that was potentially set in the project spec in doing this. Ideally we should probably prioritise the explicitly set value before and just use `.none` as a fallback before the result of `getDefaultBuildPhase(for:targetType:)`.

I've run out of time this evening on this one so before I look at it with a fresh pair of eyes later this weekend, I thought I'd push the changes up and maybe somebody else could give some input (👀 @yonaskolb 😬)


Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -1301,6 +1292,42 @@ public class PBXProjGenerator {
...
+            }
+
+            // If we haven't yet looked for the default info plist, try doing so
+            if searchForDefaultInfoPlist {
+                searchForDefaultInfoPlist = false
+
+                if let plistPath = getInfoPlist(target.sources) {
+                    let relative = (try? plistPath.relativePath(from: projectDirectory ?? project.basePath)) ?? plistPath
+                    defaultInfoPlist = relative.string
+                }
```
liamnichols, line 1318: In the original implementation, `plistPath` was assigned to the `BuildSettings` dictionary as type `Path`:

https://github.com/yonaskolb/XcodeGen/blob/05ea968f703792084cc70e5a3ddef5dfcc039f0a/Sources/XcodeGenKit/PBXProjGenerator.swift#L1178-L1180

But in other cases, it was a `String`:

https://github.com/yonaskolb/XcodeGen/blob/05ea968f703792084cc70e5a3ddef5dfcc039f0a/Sources/XcodeGenKit/PBXProjGenerator.swift#L1173

Since I'm now converting them to `String` here instead of relying on `Any` like the previous implementation, one of the tests started failing because it was casting to `Path`. But I also see other tests cast to `String`, so I'm wondering if it matters that I cast to string here?

Here is an example where the tests read the build setting as `String`:

https://github.com/yonaskolb/XcodeGen/blob/05ea968f703792084cc70e5a3ddef5dfcc039f0a/Tests/XcodeGenKitTests/ProjectGeneratorTests.swift#L1370

Here is the test failure:

> generate Info.plist: 0 is not equal to 2

https://github.com/yonaskolb/XcodeGen/blob/05ea968f703792084cc70e5a3ddef5dfcc039f0a/Tests/XcodeGenKitTests/ProjectGeneratorTests.swift#L1506-L1507

If I change line 1506 above in the tests to `as? String`, it will pass. But I just want to confirm that it's ok for me to be holding the Info.plists as `String` here? The build settings seem to convert to `String` at some point anywhere (I'm just not sure where) and given that `info.path` is already a string, I don't see this as a breaking change in behaviour but would like a second opinion 🙇

yonaskolb: Yep, fine to use String. I'd consider it a bug that the build settings contained a `Path`. On the extremely long todo list is changing the type from `Any` to an enum of type `BuildSettingValue` that has cases of `String` or `[BuildSettingValue]`

yonaskolb:
Thanks so much for looking at this @liamnichols, I like the way you're approaching it so far

Sources/XcodeGenKit/PBXProjGenerator.swift
liamnichols: Thanks for confirming, its good to know! I'll update the test to expect `String` then 🚀

liamnichols:
@yonaskolb: I hit a bit of a bump here while trying to think things through throughly...

In theory, the value of the `INFOPLIST_FILE` build setting could be something like `${SRCROOT}/SupportingFiles/${CONFIG_NAME}-Info.plist` and if this is the case, we'd have no way to be able to figure that one out how that maps to a `SourceFile` to be able to influence the buildPhase that is assigned. 

While I think it's probably uncommon to change the filename based on random build settings, I don't think it's unrealistic for people to have carried over `SRCROOT` or concepts like that as a way to use a path relative to the project root. 

These are the options that I came up with:

1. Just assume that if the user defined `INFOPLIST_FILE` in their spec, that the value is relative to the project root and does not include any variables that require expansion and accept that behaviour will be inconsistent. 
2. Attempt to expand any variables - But I think that this is impossible because we don't have the compile-time build settings? 
3. **Only** automatically alter the build phase in the following circumstances:
  i. If the user defined the `info.path` property in the spec
  ii. If the user neither defined `info.path` OR `INFOPLIST_FILE` and we instead XcodeGen automatically found the file called exactly **Info.plist** in the targets sources.

Option 3 would basically mean that if you defined `INFOPLIST_FILE` in your project spec and your info property list was called something other than **Info.plist** then you would have to explicitly stop that **.plist** from being treated as a bundle resource. I think that this is the most sensible option, but I'm interested in feedback.

Either way, we would probably have to document this a bit better somewhere.

yonaskolb:
@liamnichols. I think we want to optimise for a working project. At the end of the day, if the Info.plist file is bundled as a resource it's not a major deal, and there will always be a way to remove this by explicitly setting all those files to resource none. So we should use our best efforts to exclude it, but within reason. Option 2 is something I don't want to handle. I think option 1 is fine. We'll check against the path, and if that includes build variables it just won't match and won't be excluded.

liamnichols:
Before I started trying to pass the phases in, I found a couple of method arguments that seemed to be redundant in `SourceGenerator` and did a little cleanup in 7dd55fa5b91fb0166276034073e843b0cc7c146d

#### `getSourceFiles(targetType:targetSource:path:)` 

This method accepted a `path` value but it was always `project.basePath + targetSource.path` so seemed redundant injecting it given that we could compute the value inside the method. 

Doing this also avoided a bit of confusion where parts of the method used `path` and other parts declared a local  `folderPath` argument which was also the same - `let folderPath = project.basePath + Path(targetSource.path)`. 

I might have missed the initial intentions of injecting the `path` argument, but since its redundant today, it seems ok to remove it.

#### `generateSourceFile(targetType:targetSource:path:buildPhase:fileReference:)`

This method was my main target during refactoring. It accepted a `buildPhase` property (optional, `nil`) that if set, was used as a priority over the value specified as `targetSource.buildPhase`. 

It turns out that this argument was only ever used in one of the calls to this method, and instead of using this value as an override, the intention was that actually this value was to be used as a fallback in the event hat `targetSource.buildPhase` had not been set with had resulted in some duplicated code. 

To work around this, I removed the argument and actually just incorporated the use internally within the method by refactoring some common logic into `resolvedTargetSourceType(for:at:)` in order to override the buildPhase to be `.resources` when the `SourceType` is `.folder`. The reasoning behind this should hopefully be clear in the diff of this commit specifically

yonaskolb:
Looks good to me @liamnichols. Thanks for cleaning this up. This file and the PBXGenerator have had all sorts of hands in it changing different things, so they need a bit of love 😄

Sources/XcodeGenKit/SourceGenerator.swift
```diff
@@ -77,13 +77,19 @@ class SourceGenerator {
...
 
-    func getAllSourceFiles(targetType: PBXProductType, sources: [TargetSource]) throws -> [SourceFile] {
-        try sources.flatMap { try getSourceFiles(targetType: targetType, targetSource: $0) }
+    /// Collects an array complete of all `SourceFile` objects that make up the target based on the provided `TargetSource` definitions.
+    ///
+    /// - Parameters:
+    ///   - targetType: The type of target that the source files should belong to.
+    ///   - sources: The array of sources defined as part of the targets spec.
+    ///   - buildPhases: A dictionary containing any build phases that should be applied to source files at specific paths in the event that the associated `TargetSource` didn't already define a `buildPhase`. Values from this dictionary are used in cases where the project generator knows more about a file than the spec/filesystem does (i.e if the file should be treated as the targets Info.plist and so on).
+    func getAllSourceFiles(targetType: PBXProductType, sources: [TargetSource], buildPhases: [Path : BuildPhaseSpec]) throws -> [SourceFile] {
```
liamnichols, line 86: I'm not sure if trying to retroactively add documentation to this method myself is a great idea. I probably didn't describe it as well as I need to, but the important bit here was for me to explain what the `buildPhases` argument does.

Tests/XcodeGenKitTests/SourceGeneratorTests.swift
```diff
@@ -559,13 +559,13 @@ class SourceGeneratorTests: XCTestCase {
...
                     - file.h
-                    - Info.plist
+                    - GoogleService-Info.plist
                     - file.xcconfig
                   B:
                     - file.swift
                     - file.xcassets
                     - file.h
-                    - Info.plist
+                    - Sample.plist
```
liamnichols, line 568: I renamed these files in this test because the _real_ **Info.plist** is actually in group `C` defined on line 592 I assume because files in group `A` are all resources and files in group `B` had already been excluded automatically in the `TargetSource` definition. 

The previous logic ensured that **Info.plist** from group `C` had its `buildPhase` set to `.none` since it hadn't been previously overridden anywhere so it was falling back to the default value. I had to rename **A/Info.plist** and **B/Info.plist** because now the behaviour has changed so that only the resolved `INFOPLIST_FILE` gets it's build phase set to `.none` and since this test didn't define `info` in the spec (or the `INFOPLIST_FILE` value explicitly) then it falls back to the **first** file named **Info.plist** in the targets sources. 

Changing the names of the other two files just helps align with the _expected_ behaviour where the 'first' **Info.plist** is discovered as **C/Info.plist**.

Tests/XcodeGenKitTests/SourceGeneratorTests.swift
```diff
@@ -605,19 +605,19 @@ class SourceGeneratorTests: XCTestCase {
                     TargetSource(path: "B", buildPhase: BuildPhaseSpec.none),
                     TargetSource(path: "C", buildPhase: nil),
                 ])
-                let project = Project(basePath: directoryPath, name: "Test", targets: [target])
+                let project = Project(basePath: directoryPath.absolute(), name: "Test", targets: [target])
```
liamnichols, line 608: I had to change this to use `absolute()` here because otherwise the following logic wouldn't work:

https://github.com/yonaskolb/XcodeGen/blob/b8b457abdb29681b7692cf4148c299e213445d8d/Sources/XcodeGenKit/PBXProjGenerator.swift#L1178-L1180

Because `project.basePath` was just `TestDirectory/`, it couldn't work out the path relative to `plistPath` so instead was falling back to the absolute path. Is this just something that applied to this test case, or could we expect that `basePath` isn't always an absolute path?

I could move the `.absolute()` into my reimplementation of `getInfoPlists(for:)`, but since it wasn't already there, I don't know if I need to because there was already a bug here or because the `basePath` is already resolved elsewhere?

liamnichols (commented):
@yonaskolb: I've now _finished_ the implementation with 06bd547 to get the tests into a passing state. Again, feedback on the approach would be super appreciated 🙏 It was simpler than I was expecting but I wonder if I need to take extra care around resolving paths? 

I could probably do with adding some extra tests to **SourceGeneratorTests.swift** to catch some edge cases but just wondering how much to add. 

Cheers!

Tests/XcodeGenKitTests/SourceGeneratorTests.swift
yonaskolb: I don't think we can rely on `basePath` being absolute, so maybe better to add to `getInfoPlists` if it relies on it. We could potentially make `basePath` always absolute, but I'm not sure if that will break anything else

yonaskolb (commented):
Looks good @liamnichols!

Tests/XcodeGenKitTests/SourceGeneratorTests.swift
liamnichols: Ok thanks for confirming. I'll move the call to `absolute()` into `getInfoPlist(for:)` and try and add some tests around this to cover both scenarios

yonaskolb:
@liamnichols whenever you have time to update this, it would be good to release this, as it's a fix in some respects

liamnichols:
Ah thanks for reminding me. I’ll have some time over this weekend to get it ready 👍

liamnichols:
Things left to do:

- [ ] Rebase from `master`
- [ ] Update CHANGELOG.md
- [ ] Move the call to `Path.absolute()` into `getInfoPlist(for:)`
- [ ] Add some extra unit tests to **SourceGeneratorTests.swift** 
- [ ] Clarify behaviour in docs if needed

liamnichols:
> - [  ] Clarify behaviour in docs if needed

Not sure if i need to do anything here actually. Let me know what you think @yonaskolb. 

This is ready now 👍

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 #### Fixed
 - Fixed regression on **.storekit** configuration files' default build phase. [#1026](https://github.com/yonaskolb/XcodeGen/pull/1026) @jcolicchio
 - Fixed bug where schemes without a build target would crash instead of displaying an error [#1040](https://github.com/yonaskolb/XcodeGen/pull/1040) @dalemyers
+- Fix files such as **GoogleServices-Info.plist** from being omitted from the Copy Resources build phase. Now, only the resolved Info.plist path for a specific target is skipped. [#1027](https://github.com/yonaskolb/XcodeGen/pull/1027) @liamnichols
```
freddi-kit, line 13:
```suggestion
- Fix files with name ending `Info.plist` (such as **GoogleServices-Info.plist**) from being omitted from the Copy Resources build phase. Now, only the resolved `Info.plist` path for a specific target is skipped. [#1027](https://github.com/yonaskolb/XcodeGen/pull/1027) @liamnichols
```

liamnichols: Thanks @freddi-kit! 

Counter-suggestion: 

```suggestion
- Fixed files with names ending in **Info.plist** (such as **GoogleServices-Info.plist**) from being omitted from the Copy Resources build phase. Now, only the resolved info plist file for each specific target is omitted. [#1027](https://github.com/yonaskolb/XcodeGen/pull/1027) @liamnichols
```

(using **bold** for filenames)

freddi-kit: Thanks looks good!

yonaskolb (approved):
Looks good! Thank you again for your fantastic work here @liamnichols! 👏

PR #1041, opened 2021-03-24
CarthageDependencyResolver: ignore order-only dependencies

elliottwilliams:
If a target has a dependency which is `embed: false` and `link: false`, it's an "order-only" dependency which influences the build order of targets, but does not impact the target's build product. For these kinds of dependencies, it doesn't make sense to recursively embed their Carthage dependencies.

For example, consider a test bundle which needs to build after a framework (to import its swiftmodule), but shouldn't link it, because it'll be available at runtime from the host app:

```yaml
MyFramework:
  type: framework
  dependencies:
  - carthage: RxSwift
MyApp:
  type: application
  dependencies:
  - target: MyFramework
MyFrameworkTests:
  type: bundle.unit-test
  dependencies:
  - target: MyFramework
    embed: false
    link: false
```

Currently, MyFrameworkTests would redundantly embed RxSwift, even though it doesn't link or embed MyFramework. This change fixes that redundancy by preventing the dependency resolver from exploring targets which are order-only.


yonaskolb (commented):
Makes sense, thanks @elliottwilliams. The tests are failing as the fixture output has changed, can you run the tests, verify that the change is correct, and then commit the diff. 
Could you also add a changelog entry please

elliottwilliams:
Fixtures updated (and verified that the diff makes sense). Thanks for the changelog reminder, someday I'll get the hang of this 😆

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
 - Fixed framework search paths when using `.xcframework`s. [#1015](https://github.com/yonaskolb/XcodeGen/pull/1015) @FranzBusch
 - Fixed bug where schemes without a build target would crash instead of displaying an error [#1040](https://github.com/yonaskolb/XcodeGen/pull/1040) @dalemyers
 - Fixed files with names ending in **Info.plist** (such as **GoogleServices-Info.plist**) from being omitted from the Copy Resources build phase. Now, only the resolved info plist file for each specific target is omitted. [#1027](https://github.com/yonaskolb/XcodeGen/pull/1027) @liamnichols
+- Carthage frameworks are no longer embedded for "order-only" target dependencies. This avoid redundant embeds in situations where a target's sources _import_ a Carthage framework but do not have a binary dependency on it (like a test target which runs in a host app). [#1041](https://github.com/yonaskolb/XcodeGen/pull/1041) @elliottwilliams
```
yonaskolb, line 19: I'll move this in master

yonaskolb (approved):

Issue #1048, opened 2021-04-09
Multiple targets named 'Core' in: GraphViz, XcodeGen


d-date:
When install XcodeGen via Swift PM, Xcode says "Multiple targets named 'Core' in: GraphViz, XcodeGen".
Please consider to rename 'Core' to 'XcodeGenCore' as a workaround.

```swift
// swift-tools-version:5.3

import PackageDescription

let package = Package(
  name: "build-tools",
  platforms: [.macOS(.v10_11)],
  dependencies: [
    .package(url: "https://github.com/yonaskolb/xcodegen", from: "2.0.0"),
  ],
  targets: [.target(name: "BuildTools", path: "")]
)
```

### GraphViz/Package.swift

```swift
// swift-tools-version:5.2
// The swift-tools-version declares the minimum version of Swift required to build this package.

import PackageDescription

let package = Package(
    name: "GraphViz",
 ...
    targets: [
        .systemLibrary(name: "Clibgraphviz", pkgConfig: "libcgraph", providers: [
            .brew(["graphviz"]),
            .apt(["graphviz-dev"])
        ]),
        .target(
            name: "GraphViz",
            dependencies: ["Core", "DOT", "Builder", "Tools"]),
        .target(
            name: "Core",
            dependencies: []),
...
    ]
)
```

### XcodeGen/Package.swift

```swift
// swift-tools-version:5.0

import PackageDescription

let package = Package(
    name: "XcodeGen",
...
    targets: [
        .target(name: "XcodeGen", dependencies: [
            "XcodeGenCLI",
            "Version",
        ]),
   ...
        .target(name: "Core", dependencies: [
            "PathKit",
            "Yams",
        ]),
 ...

```

yonaskolb:
Hi @d-date, could you provide a sample package that includes XcodeGen and doesn't build. I don't think the first snippet you posted has enough information

freddi-kit:
![image](https://user-images.githubusercontent.com/13707872/114182224-5a4eaf80-997d-11eb-95e7-c6b17a03c492.png)

@yonaskolb 
It is easy to reproduce: adding XcodeGen by SPM to iOS project

freddi-kit:
i just made fix branch, please wait for a moment

freddi-kit:
Fixed by https://github.com/yonaskolb/XcodeGen/pull/1057

PR #1049, opened 2021-04-09
Fix multiple targets named 'Core' in: GraphViz, XcodeGen

d-date:
Fix #1048

PR #1057, opened 2021-04-14
Rename `Core` to avoid collisions with other packages (i.e. GraphViz)

elliottwilliams:
Target names must be globally unique. GraphViz added a `Core` target in https://github.com/SwiftDocOrg/GraphViz/pull/15, which collides with XcodeGen's `Core` target (meaning you can cause a build failure with `swift package update && swift build`). I'm going to open a PR in GraphViz to rename their target as well, but we should really do this in both repos, to avoid situations like this in the future.

I tried to follow the convention based on "XcodeGenKit" and "XcodeGenCLI" but please feel free to bikeshed on the name 🙂 

[SE-0226](https://github.com/apple/swift-evolution/blob/master/proposals/0226-package-manager-target-based-dep-resolution.md), once it's fully implemented, avoids target name collisions for _inactivate_ targets, which will allow for multiple packages to have same-named test targets, for instance. But the "minimal set" of targets to build still need to be uniquely named. There are some pretty deep conceptual issues which prevent a package from depending on two different swift modules that have the same name: https://forums.swift.org/t/multiple-target-issue-with-spm/16696/4


freddi-kit:
Hi, @elliottwilliams! Thank you for contribution! 
This is already done by other PR (https://github.com/yonaskolb/XcodeGen/pull/1049). Can you check it?

elliottwilliams:
@freddi-kit oh thanks, I missed that PR!

Seems like it also solves the problem. The only difference is that instead of moving the source files to a new `XcodeGenCore` directory, it provides a non-default sources path: https://github.com/yonaskolb/XcodeGen/pull/1049/files#diff-f913940c58e8744a2af1c68b909bb6383e49007e6c5a12fb03104a9006ae677eL55-R55

IMO it's cleaner to just move the files, but I can see the argument for keeping them at their current paths.

freddi-kit:
@yonaskolb Can you help to review it? this also helps to fix https://github.com/yonaskolb/XcodeGen/issues/1048. The difference with https://github.com/yonaskolb/XcodeGen/pull/1049 is directories name changing or not, please check which is prefer.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@
 
 #### Internal
 - Build universal binaries for release. XcodeGen now runs natively on Apple Silicon. [#1024](https://github.com/yonaskolb/XcodeGen/pull/1024) @thii
+- The `Core` target is renamed to avoid collisions with other packages. [#1057](https://github.com/yonaskolb/XcodeGen/pull/1057) @elliottwilliams
```
freddi-kit, line 22: Please add it to `Next Version`. This field is for already released version
https://github.com/yonaskolb/XcodeGen/pull/1057/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edL7

yonaskolb: Will fix on master

yonaskolb (approved):
Thanks @elliottwilliams. Strange that we have to do this and SwiftPM doesn't handle it for us

PR #1059, opened 2021-04-15
Add support for specifying Info.plist paths in an xcconfig using `SRCROOT` and `PRODUCT_NAME` variables

dalemyers:
If the info plist is set in an xcconfig then it has access to the same variables that Xcode uses. The most common of these for paths are `SRCROOT` and `PRODUCT_NAME`. This adds support for replacing those variables during evaluation.

Example: If you specify your Info.plist as `$(SRCROOT)/SomePath/$(PRODUCT_NAME)/Info.plist` then, while it will be found and used, it won't be recognised as the canonical one, and therefore will be added to the copy resources phase. With this change, `/some/path/on/disk/project/SomePath/MyTarget/Info.plist` and `$(SRCROOT)/SomePath/$(PRODUCT_NAME)/Info.plist` are treated as the same (assuming `SRCROOT` = `/some/path/on/disk/project` and `PRODUCT_NAME` = `MyTarget`)

### DIFF
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -1315,11 +1315,34 @@ public class PBXProjGenerator {
         var searchForDefaultInfoPlist: Bool = true
         var defaultInfoPlist: String?
 
+
+        /// Expand known variables in a path
+        ///
+        /// Notes: So far, variables supported are:
+        ///     * SRCROOT
+        ///     * PRODUCT_NAME
+        ///
+        /// - Parameter path: The path to expand the variable in
+        /// - Returns: The fully expanded path
+        func expandVariablesInPath(_ path: String) -> String {
+            var expanded = path.replacingOccurrences(of: "$(PRODUCT_NAME)", with: target.productName)
+
+            // Do the one with the prefix first as it's most likely and lets
+            // us handle paths correctly
+            if expanded.hasPrefix("$(SRCROOT)/") {
+                expanded = String(expanded.dropFirst("$(SRCROOT)/".count))
+            }
+
+            expanded = expanded.replacingOccurrences(of: "$(SRCROOT)", with: project.basePath.string)
+
+            return expanded
+        }
+
         let values: [(Config, String)] = project.configs.compactMap { config in
             // First, if the plist path was defined by `INFOPLIST_FILE`, use that
             let buildSettings = project.getTargetBuildSettings(target: target, config: config)
             if let value = buildSettings["INFOPLIST_FILE"] as? String {
-                return (config, value)
+                return (config, expandVariablesInPath(value))
             }
 
             // Otherwise check if the path was defined as part of the `info` spec
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## Next Version
 
+#### Added
+- Add support for specifying Info.plist paths in an xcconfig using `SRCROOT` and `PRODUCT_NAME` variables [#1058](https://github.com/yonaskolb/XcodeGen/pull/1058) @dalemyers
+
+#### Fixed
+
+#### Internal
+
+
 ## 2.20.0
 
 #### Added


Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -1315,11 +1315,34 @@ public class PBXProjGenerator {
...
+
+            return expanded
+        }
+
         let values: [(Config, String)] = project.configs.compactMap { config in
             // First, if the plist path was defined by `INFOPLIST_FILE`, use that
             let buildSettings = project.getTargetBuildSettings(target: target, config: config)
             if let value = buildSettings["INFOPLIST_FILE"] as? String {
-                return (config, value)
+                return (config, expandVariablesInPath(value))
```
dalemyers, line 1345: It's not clear to me if we should also do this on lines 1350 and 1367.

liamnichols:
This is a bit of a tricky one that I flagged up originally here - https://github.com/yonaskolb/XcodeGen/pull/1027#issuecomment-785410136

I think that its a bit more difficult to this in a predicable manner though because build setting variables are actually very hard to detect. For example, off the top of my head I can think of the following edge cases:

- `$SRC_ROOT` - no brackets 
- `${SRC_ROOT}` - curly
- `$(SRC_ROOT)` - not curly
- `$SOURCE_ROOT` - alias
- `${PRODUCT_NAME:rfc1034identifier}` -

liamnichols:
This is a bit of a tricky one that I flagged up originally here - https://github.com/yonaskolb/XcodeGen/pull/1027#issuecomment-785410136

I think that its a bit more difficult to this in a predicable manner though because build setting variables are actually very hard to detect. For example, off the top of my head I can think of the following edge cases:

- `$SRCROOT` - no brackets 
- `${SRCROOT}` - curly
- `$(SRCROOT)` - not curly
- `$SOURCE_ROOT` - alias
- `${PRODUCT_NAME:rfc1034identifier}` - does some magic
- `${PRODUCT_NAME:identifier}` - does some more magic
- `${PRODUCT_NAME:default=Hello}` - new in Xcode 11

There is probably some more as well 🙈 

---

But, this is an interesting problem. For `SRCROOT`, can you not just omit it? Since the `INFOPLIST_FILE` will be relative to src root anyway making these two build settings the same:

```
INFOPLIST_FILE = "App/Info.plist"
INFOPLIST_FILE = "$(SRCROOT)/App/Info.plist"
```

As for `PRODUCT_NAME`, it's a bit tricker I guess, but using templates we can work around it if we also use the generated `info.path` property in the spec and use XcodeGen's variable expansion instead. But its not really a work around to your issue and issues that other people might also face.

dalemyers:
Yeah, I figured it wasn't going to be this easy, but I'm new to the project, so figured I'd give it a go and see. 

As for why don't I just omit it? Simple: We have a project that I'm moving to XcodeGen and I'm currently generating the project.yml from the old pbxproj. I had a choice to make the fix in my script, or in XcodeGen. I figured doing it in XcodeGen (if it worked) would help the most people. 

Feel free to close this though!

liamnichols:
Ahh thanks, for the explanation. Sorry if it sounded dismissive in any way, I was just coming from a semi-related issue and I wanted to share some context of the issues that I ran into in the initial implementation 😄 

I think that in general, trying to predict and replace build setting values in XcodeGen is probably a bit out of scope of the tools capabilities since in theory the values can be anything and it would be too hard (maybe even impossible?) to achieve reliably. More holistically, I feel that if we try to do this in some places and not others then it's probably going to cause some confusion along the way (although adding documentation in the source code is a great start) so maybe instead there might be a better approach. 

In your existing Xcode Project, how were you managing the Info.plist's build phase setting? Was it just that you had manually removed them all from Copy Bundle Resources? 

Alternatively, you could actually leverage Xcode Gen's variable replacement (at generation time) and take advantage of the `${target_name}` variable that is available when resolving `targetTemplates`. This is how I do it in my project:

```yml
targetTemplates:
  FrameworksBase:
    templates: [VersionedInfoPlist]
    platform: iOS
    settings:
      base:
        INFOPLIST_FILE: ${target_name}/SupportingFiles/Info.plist
    sources:
      - ${target_name}

# Target Definitions
targets:
  FrameworkTargetOne:
    templates: [OtherTemplate, FrameworksBase]
    scheme:
      testTargets:
        - FrameworkTargetOneTests
  FrameworkTargetOneTests:
    templates: [FrameworksBase]
    type: bundle.unit-test
    dependencies:
      - target: FrameworkTargetOne
  FrameworkTargetTwo:
    templates: [OtherTemplate, FrameworksBase]
    scheme:
      testTargets:
        - FrameworkTargetTwoTests
  FrameworkTargetTwoTests:
    templates: [FrameworksBase]
    type: bundle.unit-test
    dependencies:
      - target: FrameworkTargetTwo
```

In this setup, the `FrameworkTargetOne`, `FrameworkTargetOneTests`, `FrameworkTargetTwo` and `FrameworkTargetTwoTests` targets all leverage the same target template and all have their **Info.plist** files managed properly.

While it looks like `${target_name}` is a build setting that would end up in the .pbxproj file, this is actually handled by the generators variable expansion which will be done by XcodeGen when loading the project spec and as a result, the `PBXProjGenerator` will already have the resolved `INFOPLIST_FILE` and will be able to properly configure the build phase for it.

Would this help you?

dalemyers:
Not dismissive at all. You are 100% right. 

And for our end, it's much easier than that. I can just run basically exactly the same code on our end (since it's all `$(SRCROOT)` and `$(TARGET_NAME)` and we know those for certain. 

>Was it just that you had manually removed them all from Copy Bundle Resources?

They don't get automatically added by Xcode as far as I'm aware.

yonaskolb:
Thank you for your work here @dalemyers. As @liamnichols says though, I don't think this is a rabbit hole we want to go down in this project.

PR #1062, opened 2021-04-23
Update graphviz to 0.4.0

swiftty:
When I use XcodeGen as dependency, SwiftPM fetches latest GraphViz package.
So it causes compilation error.

related? #1057 #1049


yonaskolb:
Fixed by https://github.com/yonaskolb/XcodeGen/pull/1057

PR #1066, opened 2021-05-02
Fix CI: install system dependdencies for graphviz

yanamura:
Fix this issue
https://github.com/yonaskolb/XcodeGen/pull/1065#issuecomment-830730388

Ref
https://github.com/SwiftDocOrg/GraphViz/blob/main/.github/workflows/ci.yml#L59-L61
https://github.com/SwiftDocOrg/GraphViz/blob/main/.github/workflows/ci.yml#L83-L86

PR #1065, opened 2021-05-01
Fix no such module 'DOT'

yanamura:
## Problem
create `cli` folder and put this Package.swift
```
// swift-tools-version:5.1
import PackageDescription

let package = Package(
    name: "Tools",
    dependencies: [
        .package(url: "https://github.com/yonaskolb/XcodeGen", .branch("master"))
    ]
)
```

and run
`swift run --package-path cli xcodegen`

cause this error
```
XcodeGen/Sources/XcodeGenKit/GraphVizGenerator.swift:1:8: error: no such module 'DOT'
import DOT`
```

## Solution
Update GraphViz 0.2.0 to 0.4.0
[diff](https://github.com/SwiftDocOrg/GraphViz/compare/0.2.0...0.4.0)

With this update, we no longer to import Dot.
```
Changed package to produce a single library named `GraphViz` so that
  `Dot`, `Tools`, and `GraphVizBuilder` need not be imported separately anymore.
```

PR #1067, opened 2021-05-03
Fix no such module 'DOT' by  pinning GraphViz 0.2.0

yanamura:
- Reverts yonaskolb/XcodeGen#1065
  - issue
    - https://github.com/yonaskolb/XcodeGen/pull/1066#issuecomment-830784929

## Solution
- pinned GraphViz 0.2.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,6 @@
 
 ## Next Version
 
-#### Fixed
-- Fixed no such mocule `DOT` error [#1065](https://github.com/yonaskolb/XcodeGen/pull/1065) @yanamura
-
 ## 2.21.0
 
 #### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Next Version
 
+#### Fixed
+- Fixed no such module `DOT` error [#1067](https://github.com/yonaskolb/XcodeGen/pull/1067) @yanamura
+
 ## 2.21.0
 
 #### Added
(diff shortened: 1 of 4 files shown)


Package.swift
```diff
@@ -19,7 +19,7 @@ let package = Package(
         .package(url: "https://github.com/tuist/XcodeProj.git", from: "7.18.0"),
         .package(url: "https://github.com/jakeheis/SwiftCLI.git", from: "6.0.0"),
         .package(url: "https://github.com/mxcl/Version", from: "2.0.0"),
-        .package(url: "https://github.com/SwiftDocOrg/GraphViz.git", from: "0.4.0"),
+        .package(url: "https://github.com/SwiftDocOrg/GraphViz.git", from: "0.1.0"),
```
yonaskolb, line 22: To fix the original issue in https://github.com/yonaskolb/XcodeGen/pull/1065, should this be pinned to 0.2?

CHANGELOG.md
```diff
@@ -2,9 +2,6 @@
 
 ## Next Version
 
-#### Fixed
-- Fixed no such mocule `DOT` error [#1065](https://github.com/yonaskolb/XcodeGen/pull/1065) @yanamura
```
yonaskolb, line 6: If we pin the version below I believe this should still be fixed, so you an add it back with this PR, but fix `mocule` -> `module`

Package.swift
yanamura: pinned version in Package.resolved is 0.2.0 , so both 0.1.0 and 0.2.0 are available.

yonaskolb: Yes, but I think `Package.resolved` isn't used if the package is used as a dependency in another package, which is when the issue occurs

yanamura: oh, I got it! thanks for your suggestion. I'll fix now

yonaskolb (approved):

PR #1085, opened 2021-05-27
Add ability to set executable to an external application

### LINKED ISSUE #1084: Referencing an external application for the run action executable
## Current Behavior
When setting the executable of a run action to an external application (by inserting a path), XCodeGen does not recognize this and the executable of the created project is still set to None.

For instance, I want to create a plugin template for an existing application. For debugging, this application provides a launch argument that loads the plugin, therefore I want to set the executable to the app:

```yml
schemes:
  ExampleScheme:
    run:
      executable: /Applications/Example.app
```
image

## Possible Solution
This https://github.com/tuist/XcodeProj/pull/521 PR added `PathRunnable` class to `LaunchAction` to the XcodeProj repo.

Either template generation could distinguish between paths and target names in the executable setting and create a runnable path or a buildable reference, respectively, or another option for adding a path could be added to the run action.

_XcodeGen Version: 2.23.1_

platysma:
Fixes #1084

### Summary
Added ability to set executable to an external application by providing a `filePath` in the Run Action

### Changes
Added additional field `filePath` in the Run Action. If set, a `PathRunnable` will be created in the Run Action.

```yml
schemes:
  ExampleScheme:
    run:
      filePath: /Applications/Example.app
```

### DIFF
--- a/Sources/XcodeGenKit/SchemeGenerator.swift
+++ b/Sources/XcodeGenKit/SchemeGenerator.swift
@@ -161,10 +161,14 @@ public class SchemeGenerator {
             return XCScheme.ExecutionAction(scriptText: action.script, title: action.name, environmentBuildable: environmentBuildable)
         }
 
-        let schemeTarget: Target?
-
-        if let targetName = scheme.run?.executable {
-            schemeTarget = project.getTarget(targetName)
+        var schemeTarget: Target? = nil
+        var pathRunnable: XCScheme.PathRunnable? = nil
+        
+        if let executable = scheme.run?.executable {
+            schemeTarget = project.getTarget(executable)
+        } else if let filePath = scheme.run?.filePath {
+            // Only use pathRunnable if no executable is provided.
+            pathRunnable = XCScheme.PathRunnable(filePath: filePath)
         } else {
             guard let firstTarget = scheme.build.targets.first else {
                 throw SchemeGenerationError.missingBuildTargets(scheme.name)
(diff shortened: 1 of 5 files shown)


Sources/XcodeGenKit/SchemeGenerator.swift
```diff
@@ -161,10 +161,14 @@ public class SchemeGenerator {
...
-        if let targetName = scheme.run?.executable {
-            schemeTarget = project.getTarget(targetName)
+        var schemeTarget: Target? = nil
+        var pathRunnable: XCScheme.PathRunnable? = nil
+        
+        if let executable = scheme.run?.executable {
+            schemeTarget = project.getTarget(executable)
+        } else if let filePath = scheme.run?.filePath {
+            // Only use pathRunnable if no executable is provided.
+            pathRunnable = XCScheme.PathRunnable(filePath: filePath)
```
freddi-kit, line 171: According to the ProjSpec you wrote, it looks filePath is only available for the absolute path. 
How about throwing SchemeGenerationError if the path is not the absolute path?

You can check it by using `path.isAbsolute` in `Path` class.

freddi-kit:
It looks good, please check my comments. Sorry for late

Sources/XcodeGenKit/SchemeGenerator.swift
platysma: Good idea. Added.

freddi-kit (approved):
Great, thank you!

Sources/XcodeGenKit/SchemeGenerator.swift
```diff
@@ -369,6 +377,8 @@ enum SchemeGenerationError: Error, CustomStringConvertible {
             return "Unable to find project reference named \"\(project)\" in project.yml"
         case .missingBuildTargets(let name):
             return "Unable to find at least one build target in scheme \"\(name)\""
+        case .pathNotAbsolute(let path):
+            return "Provided filePath \"\(path)\" needs to be an absolute path"
```
yonaskolb, line 381: Thanks for this error, but can we make it a `SpecValidationError.ValidationError` instead which is thrown in `SpecValidation.swift`, with an associated test case in `ProjectSpecTests.testValidation`

PR #1087, opened 2021-06-03
Add support for conditionally linking dependencies based on platform

daltonclaybrook:
This PR makes it possible to conditionally link dependencies based on a specified list of platforms. It addresses an issue raised in #951 that the "platform filter" parameter in Xcode is currently not granular enough to prevent a framework from linking against tvOS and watchOS targets.

This is currently implemented as a new field on `Dependency` called `conditionalPlatforms`. In all honestly, I would prefer that the existing `platform` setting be renamed to something like `platformFilter` to disambiguate from this new setting, but I'm open to suggestions.

(cc: @raptorxcz, @yonaskolb, @jlott1)

### DIFF
--- a/Sources/ProjectSpec/Dependency.swift
+++ b/Sources/ProjectSpec/Dependency.swift
@@ -16,6 +16,7 @@ public struct Dependency: Equatable {
     public var implicit: Bool = implicitDefault
     public var weakLink: Bool = weakLinkDefault
     public var platform: Platform = platformDefault
+    public var conditionalPlatforms: Set<ProjectSpec.Platform>?
 
     public init(
         type: DependencyType,
@@ -25,7 +26,8 @@ public struct Dependency: Equatable {
         link: Bool? = nil,
         implicit: Bool = implicitDefault,
         weakLink: Bool = weakLinkDefault,
-        platform: Platform = platformDefault
+        platform: Platform = platformDefault,
+        conditionalPlatforms: Set<ProjectSpec.Platform>? = nil
     ) {
         self.type = type
         self.reference = reference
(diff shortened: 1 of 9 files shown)


daltonclaybrook:
Spitballing here: What if we changed the other field to be called `platformFilter` and this one simply `platforms`. Possible changes to docs:

>- [ ] **platformFilter**: **String** - This field is specific to Mac Catalyst. It corresponds to the "Platforms" dropdown in the Frameworks & Libraries section of Target settings in Xcode. Available options are: **iOS**, **macOS** and **all**. Defaults is **all**
>- [ ] **platforms**: **[Platform](#platform)** - List of platforms this dependency should apply to. Defaults to all applicable platforms.

Sources/ProjectSpec/Dependency.swift
```diff
@@ -137,6 +144,7 @@ extension Dependency: JSONEncodable {
             "embed": embed,
             "codeSign": codeSign,
             "link": link,
+            "conditionalPlatforms": conditionalPlatforms.map(Array.init)?.map(\.rawValue)
```
yonaskolb, line 147: Can we add a sort to make this deterministic. And I also don't think the Array init is required

yonaskolb (commented):
Thanks @daltonclaybrook. I agree with the rename of platformFilter, so please proceed. It can be marked as a breaking change in the changelog.

daltonclaybrook:
@yonaskolb Just pushed up those changes. How do you feel about the name of the new field (currently "`conditionalPlatforms`")? Should this be changed as well?

daltonclaybrook:
@yonaskolb Please let me know what the next steps are on this. I'd really like to get this landed since my team needs it for an upcoming project.

yonaskolb:
> @yonaskolb Just pushed up those changes. How do you feel about the name of the new field (currently "`conditionalPlatforms`")? Should this be changed as well?

Yes, let's rename this to `platforms`. Sorry for the delay. Will approve after your change

yonaskolb (approved):
Great, thanks @daltonclaybrook!

PR #1091, opened 2021-06-15
Add Support for DocC in Xcode 13

### LINKED ISSUE #1089: Feature Request: DocC Support
I know it's early, but I figure this can be used for tracking. DocC seems to be a simple folder with the .docc extension appended to the name. I would be pretty helpful if XcodeGen could detect that.

Currently I have found that running XcodeGen will revert .docc back to a normal folder, so the compiler won't see it as documentation next time.

I'll try to take a look, and see if I can help at all.

brevansio:
DocC "files" are actually folders `.docc` appended to the name, but Xcode 13 treats them differently. Therefore, we need to exclude them from the normal BuildPhase.

Resolves #1089

About the PR: 

I'm a little uneasy about the tests. First, I added to an existing test because it seems like the best fit and I have confirmed that it both fails/passes when it should. If I need to move the test code or create a new test instead of modifying the existing test, let me know. Second, I'm also completely lost on the `project.pbxproj` because I couldn't figure out how to get it to fail the diff, so if I did something wrong let me know.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Next Version
 
+### Added
+
+- Added support for DocC Catalogs [#----](https://github.com/yonaskolb/XcodeGen/pull/----) @brevansio
+
 ## 2.23.1
 
 ### Changed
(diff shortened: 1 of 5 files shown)


yonaskolb (approved):
All looks great, thank you @brevansio!

Issue #933, opened 2020-08-13
Not embedding project reference targets in multiple targets


sky-duartecosta:
I have a project reference to be embedded in my main project, using its targets, but it only gets embedded in one target.

```
projectReferences:
  MyProject:
    path: Projects/MyProject/MyProject.xcodeproj
```

Targets:

```
MyTarget:
  dependencies:
      - target: MyProject.xcodeproj/MyProject.xcodeproj
        link: true
        embed: true

MyTargetOne:
  dependencies:
      - target: MyProject.xcodeproj/MyProject.xcodeproj
        link: true
        embed: true
```

Only one of the targets is embedded.

MatheusMBispo:
This is happening with my project too 😕

bluesource:
For me too. The depended targets are embedded only on the alphabetical sorted first target. 
I have 4 targets, which are defined by a template. The template define the dependencies. The dependencies are only embedded in the first target.

mqzkim:
Could I ask this is intended spec.. or bug? Are there any updates?

yonaskolb:
This is not intended behaviour, and should be treated a bug. I haven't had a chance to look into this. 
PR's welcome, even if all the PR does is add a test case that demonstrates this.

mqzkim:
I just made a project which can reproduce this bug 100%.
When I add a file in `TargetExample` target, `Core` framework disappeared.

(cf. When I made the application target to top order of targets in project, the bug doesn't occur.
-> ex. If I change the name `TargetExample` to `000_TargetExample`, `Core` framework doesn't disappear.)


https://github.com/mqzkim/xcodegen-test

Here are `project.yml` files below.

`Core/project.yml`
```
name: Core
options:
  bundleIdPrefix: com.xcodgentest
targets:
  Core:
    sources: Core
    type: framework
    platform: iOS
    deploymentTarget: "11.0"
```

`Target/project.yml`
```
name: Target
options:
  bundleIdPrefix: com.xcodgentest
projectReferences:
  Core:
    path: ../Core/Core.xcodeproj
targets:
  Target:
    sources: Target
    type: framework
    platform: iOS
    deploymentTarget: "11.0"
    dependencies:
      - target: Core/Core
  TargetExample:
    sources: TargetExample
    type: application
    platform: iOS
    deploymentTarget: "12.0"
    transitivelyLinkDependencies: true
    dependencies:
      - target: Target
```

mqzkim:
[...]

PR #1108, opened 2021-07-19
Embedding project reference targets in multiple targets

mqzkim:
To solve this. https://github.com/yonaskolb/XcodeGen/issues/933

- Added test code for MultipleXcodeProject with MultipleTargetDependencies

Also)
- Project framework would be included frameworkFiles.


PaulTaykalo:
if there are mutliple targets with the same depdency on same target from project reference, this fix will set incorrect dependencies

mqzkim:
> if there are mutliple targets with the same depdency on same target from project reference, this fix will set incorrect dependencies

@PaulTaykalo Would you provide any example that I can reproduce your issue?

simonpierreroy:
I tried to see what an Xcode generated setup would look like. It seems this PR is definitely on the correct path.
Unique `PBXContainerItemProxy` are needed per target.

@PaulTaykalo Did you have more info on the issues with this PR?

Screenshot 2023-03-15 at 5 55 01 PM

PaulTaykalo:
Tried this branch again - have the same result.
this update will break dependencies generation with complex depencies.

This is an oversimplified example of our project.
If arrows are dependencies, and `B` is target from project reference, then for project C, dependency on B target is not present in the XML. (Or it's invalid. Hard to tell).
```mermaid
graph TD
A.pbxproj --> B[Target from B.pbxproj]
C.pbxproj --> B[Target from B.pbxproj] 
linkStyle 1 stroke: red, stroke-width: 2px;
```

I can't provide more detailed explanation on this, but this PR breaks that part.
Not sure if I'll be able to drill down to the root cause of the problem

PaulTaykalo:
Like these deps are not generated:
## Incorrect (this PR)
image

## Correct
image

PR #1110, opened 2021-07-20
Fix issue that caused static framework target dependencies from having their default `link` value inferred as `true`

liamnichols:
I'm updating our project to statically link some framework targets instead of using dynamic linking for everything and I spotted some warnings appearing in CI when doing do:

> [...]/usr/bin/libtool: file: /[...]/MyFramework1.framework/MyFramework1 is a dynamic library, not added to the static library

It appears that this happens when I have three targets:

1. `MyFramework1` - Dynamic Framework  (`framework`)
2. `MyFramework2` - Static Framework (`framework.static`) with dependency on `MyFramework1`
3. `MyApp` - iOS app target with dependency on `MyFramework2` (and a transitive dependency on `MyFramework1`)

In this setup, the `MyFramework2` target should not have a "Link Binary With Libraries" build phase because the libraries should be linked into the executable target `MyApp` instead. After looking at the docs, and digging into the current fixtures, it appeared that this was defaulting the `link` property to the correct value if `MyFramework2` was a static library (`library.static`) but not for `framework.static` or `framework` with the `MACH_O_TYPE` build setting set to `staticlib`.

After setting up some extra targets in the fixtures to reproduce the setup (1f391f1), i tracked the issue down to `PBXProjectGenerator` and updated it to use the `defaultLinkage` property that is also used for figuring out if the dependency should be embedded or not (159f719). You can find the fixtures regenerated after making the changes in 
8e19482.

### DIFF
--- a/Docs/ProjectSpec.md
+++ b/Docs/ProjectSpec.md
@@ -127,7 +127,7 @@ Note that target names can also be changed by adding a `name` property to a targ
 - [ ] **fileTypes**: **[String: [FileType](#filetype)]** - A list of default file options for specific file extensions across the project. Values in [Sources](#sources) will overwrite these settings.
 - [ ] **preGenCommand**: **String** - A bash command to run before the project has been generated. If the project isn't generated due to no changes when using the cache then this won't run. This is useful for running things like generating resources files before the project is regenerated.
 - [ ] **postGenCommand**: **String** - A bash command to run after the project has been generated. If the project isn't generated due to no changes when using the cache then this won't run. This is useful for running things like `pod install` only if the project is actually regenerated.
-- [ ] **useBaseInternationalization**: **Bool** If this is `false` and your project does not include resources located in a **Base.lproj** directory then `Base` will not be included in the projects 'known regions'. The default value is `true`. 
+- [ ] **useBaseInternationalization**: **Bool** If this is `false` and your project does not include resources located in a **Base.lproj** directory then `Base` will not be included in the projects 'known regions'. The default value is `true`.
 - [ ] **schemePathPrefix**: **String** - A path prefix for relative paths in schemes, such as StoreKitConfiguration. The default is `"../../"`, which is suitable for non-workspace projects. For use in workspaces, use `"../"`.
 
 ```yaml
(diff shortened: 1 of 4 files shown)


liamnichols:
Before I mark this as ready for review, there are a couple of other places in `PBXProjGenerator` that also perform similar checks and I'm wondering if they also need similar patches? I've personally only ever used the `target` type of dependency so i'm not 100% sure how these other areas need to behave.

## Requires ObjC Linking

https://github.com/yonaskolb/XcodeGen/blob/a4d7a61a6805f62b6da9f7ffb5cdf13a18d16a40/Sources/XcodeGenKit/PBXProjGenerator.swift#L456

https://github.com/yonaskolb/XcodeGen/blob/a4d7a61a6805f62b6da9f7ffb5cdf13a18d16a40/Sources/XcodeGenKit/PBXProjGenerator.swift#L720-L723

I'm not completely sure if "ObjC linking" is the same and if it applies to static frameworks or not?

## Other Dependency Types

### `.framework`

https://github.com/yonaskolb/XcodeGen/blob/a4d7a61a6805f62b6da9f7ffb5cdf13a18d16a40/Sources/XcodeGenKit/PBXProjGenerator.swift#L802-L808

I think this is intentionally specific to libraries so can remain 👍 

### `.carthage`

https://github.com/yonaskolb/XcodeGen/blob/a4d7a61a6805f62b6da9f7ffb5cdf13a18d16a40/Sources/XcodeGenKit/PBXProjGenerator.swift#L897-L904

I think _maybe_ this needs patching,  but i've not used Carthage in a longgg time so I can't be sure

### `.package`

https://github.com/yonaskolb/XcodeGen/blob/a4d7a61a6805f62b6da9f7ffb5cdf13a18d16a40/Sources/XcodeGenKit/PBXProjGenerator.swift#L926-L937

Again, I don't have enough context to know the right answer

### `.bundle`

https://github.com/yonaskolb/XcodeGen/blob/a4d7a61a6805f62b6da9f7ffb5cdf13a18d16a40/Sources/XcodeGenKit/PBXProjGenerator.swift#L951-L953

Looks good, its specifically checking for libraries (not link types).


----


@yonaskolb: Any input on the bits above would be great, thanks! 🙏

yonaskolb:
Hi @liamnichols. Thanks for this, your change makes sense. As for the other places, I'm honestly not sure. Maybe just go with your change, and someone else might bring up if something else is not working. For this could you add a changelog entry, might need to rebase off master to get the latest changelog.
I know it's a been a while, but if you had the time to add some a unit test for this case it would be great too

Issue #1111, opened 2021-07-30
Feature request: Expose merged/resolved yaml


namolnad:
Hi there,

I'm wondering if it would be possible to somehow expose the merged/fully-resolved yaml/json after templates and templateAttributes have been replaced — likely via a new `xcodegen` command.
### Background
My team has broken down our project spec into a series of `BUILD.yml` files as well as separate template files all added together with our `project.yml` file via `includes` directives. We've recently started adding metadata to the yml target definitions to provide our tooling with additional context about a given framework. I've been able to merge the yaml together by creating a custom yaml loader and some merging functions, but the templateAttributes is something I haven't taken on yet and I'm curious if all of this is something that would be relatively simple to expose from XcodeGen. Would love to hear your thoughts. Thanks!

yonaskolb:
Hi @namolnad, you're in luck, as I believe the dump command already does what you're looking for.
```
xcodegen dump --type yaml
```
See help for the different formats you can output
```
xcodegen dump help
```

Closing this, but you can reopen if you run into any troubles.

namolnad:
🤦‍♂️ can't believe I missed this :) Thanks so much for the quick response!

PR #1112, opened 2021-07-31
[bug fix] Ensure fileTypes are mapped to JSON value

namolnad:
After finding out that #1111 was already possible via `xcodegen dump`, I tried it out for my project, but noticed that there was an error for my project when trying to use the `parsed-yaml` or the `parsed-json` options. It turns out there was a JSON/Yaml serialization error due to `fileTypes` still being the raw Swift type vs the JSON value.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 ### Fixed
 
 - Fix broken codesign option for bundle dependency [#1104](https://github.com/yonaskolb/XcodeGen/pull/1104) @kateinoigakukun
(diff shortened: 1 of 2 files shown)


yonaskolb (commented):
Thanks @namolnad! Could you please add a changeling entry too

namolnad:
> Thanks @namolnad! Could you please add a changeling entry too

Sure thing! Added!

yonaskolb (approved):

namolnad:
@yonaskolb do you have any estimates on when you think the next release may go out?

PR #1120, opened 2021-08-25
Speed up path checking

PaulTaykalo:
For checking if path is subpath of another, we don't actually need to use `contains` method.
`hasPrefix` will do all the job and will be faster to check.

This is because in the method, when we're checking paths, we're operating with full path's, so using `contains` is not only slower, but can also get false positives in some rare cases

Using this fix only, decreased project generation from about 14s to 5s on my project.

## Before
image

## After
image


PaulTaykalo:
#1122 makes this PR irrelevant

PR #1122, opened 2021-09-09
Use binary search for Checking if path is included in included files

PaulTaykalo:
in this PR, instead of checking if each file contains in the paths, we're using binary search.
This PR speed ups checks, especially in cases when multiple files are added (i.e few thousands)

Before this PR, complexity of the checking if file is included is O(N * M) where N - files to check, M - included files

This PR will decrease complexity to O(N * log(M)) which significantly improves operations on big projects.


PaulTaykalo:
@yonaskolb is there something I can help in order this PR to be merged in?

Sources/XcodeGenCore/ArrayExtensions.swift
```diff
@@ -0,0 +1,79 @@
...
+    ///   and returns a Boolean value that indicates whether the passed element
+    ///   represents a match.
+    ///
+    /// - Returns: The index of the first element for which `predicate` returns
+    ///   `true`. If no elements in the collection satisfy the given predicate,
+    ///   returns `nil`.
+    ///
+    /// - Complexity: O(log(*n*)), where *n* is the length of the collection.
+    @inlinable
+    func firstIndex(where predicate: (T) throws -> Bool) rethrows -> Int? {
```
yonaskolb, line 48: Could you please add some tests for this

yonaskolb (commented):
Thanks for this @PaulTaykalo! Do you have examples of before and afters in a large project?
Could you please add a changeling entry as well

Sources/XcodeGenCore/ArrayExtensions.swift
PaulTaykalo: Done

Tests/XcodeGenCoreTests/ArrayExtensionsTests.swift
```diff
@@ -0,0 +1,25 @@
+import XCTest
+@testable import XcodeGenCore
+
+class ArrayExtensionsTests: XCTestCase {
```
yonaskolb, line 4: Another thing the SortedArray does is sort the elements it's initialised with if we could test that too

yonaskolb:
@Keith are you still using this at Lyft? Would be great to see the results on your project, as it's the biggest I've heard of

Tests/XcodeGenCoreTests/ArrayExtensionsTests.swift
PaulTaykalo: Done

namolnad:
> @Keith are you still using this at Lyft? Would be great to see the results on your project, as it's the biggest I've heard of

I can also test this on our project at Instacart and report back!

keith:
Here's some tests with a very large configuration for us:

master xcodegen:

  Time (mean ± σ):     42.438 s ±  4.635 s    [User: 64.061 s, System: 4.952 s]
  Range (min … max):   36.565 s … 49.839 s    10 runs

changed xcodegen:

  Time (mean ± σ):     40.603 s ±  2.978 s    [User: 62.073 s, System: 4.567 s]
  Range (min … max):   36.873 s … 44.839 s    10 runs


so this definitely helps a bit, but maybe this just isn't a path we're hurt much by

PaulTaykalo:
@keith it seems that your case is - many configurations, many small projects. Am I right?
My case is more like - few configurations, heavy projects (over 200-1000+ files).

PR #1123, opened 2021-09-13
Fix platformFilter for package dependencies

raptorxcz:
If dependency is package then is not possible to set platformFilter. According to documentation it should be possible.

https://github.com/yonaskolb/XcodeGen/blob/master/Docs/ProjectSpec.md#dependency


yonaskolb (approved):
This is great, thank you @raptorxcz 🙏

PR #1128, opened 2021-09-17
Add Debug Logging

dalemyers:
We've run into an issue where we are getting some kind of race condition, so we increased logging to try and debug. It made sense to contribute it back. None of these messages (or the mechanism) are set in stone. 

Basically, `-q` works as it did before, but now `-d` displays all debug messages as well. If you specify both, `q` takes priority. 

Testing it on our project, there is no difference in run time to generate.

### DIFF
--- a/Sources/XcodeGenCLI/Commands/ProjectCommand.swift
+++ b/Sources/XcodeGenCLI/Commands/ProjectCommand.swift
@@ -50,6 +50,8 @@ class ProjectCommand: Command, LogRenderer {
 
         let projectSpecPath = (spec ?? "project.yml").absolute()
 
+        Logger.shared.debug("Checking for spec file: \(projectSpecPath)")
+
         if !projectSpecPath.exists {
             throw GenerationError.missingProjectSpec(projectSpecPath)
         }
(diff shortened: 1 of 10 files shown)


Sources/XcodeGenCLI/Commands/ProjectCommand.swift
```diff
@@ -46,8 +67,31 @@ class ProjectCommand: Command {
...
+    func debug(_ string: String) {
+        stdout.print("[DEBUG] \(string)")
+    }
+
+    func info(_ string: String, wasSuccess: Bool) {
+        var coloredString = string
+        if wasSuccess {
+            coloredString = coloredString.green
+        }
+        stdout.print("[INFO] \(coloredString)")
```
dalemyers, line 87: This is the only "visible" change of this PR. Instead of seeing:

```
⚙️  Generating plists...
⚙️  Generating project...
⚙️  Writing project...
```

You'll now see:

```
[INFO] ⚙️  Generating plists...
[INFO] ⚙️  Generating project...
[INFO] ⚙️  Writing project...
```

dalemyers:
This was to help debug https://github.com/yonaskolb/XcodeGen/issues/1131 but I found a fix anyway.

PR #1135, opened 2021-10-05
Extend possible paths for SettingsPresets

SofteqDG:
* Search for presets in Bundle.main.resourcesPath dir (if exists)

### DIFF
--- a/Sources/XcodeGenKit/SettingsBuilder.swift
+++ b/Sources/XcodeGenKit/SettingsBuilder.swift
@@ -193,6 +193,10 @@ extension SettingsPresetFile {
             Path(#file).parent().parent().parent() + relativePath,
         ]
 
+        if let resourcePath = Bundle.main.resourcePath {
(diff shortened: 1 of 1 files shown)


yonaskolb:
Thanks @SofteqDG. Is this for using XcodeGenKit as a dependency, or what is your workflow here?

SofteqDG:
> Thanks @SofteqDG. Is this for using XcodeGenKit as a dependency, or what is your workflow here?

Hello. Yes, it is for XcodeGenKit as a dependency. I'm using XcodeGenKit as SPM dependency and need to include presets in app bundle to make my app work on other macs.

SofteqDG:
@yonaskolb,
Hey. Any thoughts on this change? Can it be merged?

yonaskolb (commented):
Sorry for the late reply on this @SofteqDG. I'm merging this now, I'll update the changelog entry as well.
Having to do this should be negated by implementing https://github.com/yonaskolb/XcodeGen/issues/1267

Your use case sounds cool 🙂

Issue #937, opened 2020-08-18
Config to show/hide specific schemes


mime29:
Is there a way to show/hide specific schemes to keep the scheme selection clean?
In the scheme manager, there is a `show` checkbox that allows to hide some schemes from the active scheme selector.
So far I wasn't able to modify this parameter with xcodegen.

Thank you

wendyliga:
@mime29 after digging the rabbit hole on how `xcode` show or hide scheme, looks like the value is store on `xcuserdata`, specifically `xcschememanagement.plist`.

i don't sure xcodegen touch this before.
i honestly interensting on this idea, will look into it

mime29:
Thank you, I'm really looking forward to see that working

PR #1142, opened 2021-10-21
Added support for scheme management

wendyliga:
From this issue https://github.com/yonaskolb/XcodeGen/issues/937, we need some way to hide/show `Scheme`.
By recent addition to `tuist's xcodeproj` https://github.com/tuist/XcodeProj/pull/565, we are now able to create `XCSchemeManagement`.


teameh:
Haha you beat me to it. I needed this functionality this morning, saw the open issues and thought it would be a nice challenge to see how far I could get. 

Until I just saw you submitted this pr. Hehe. 

Well, for what it's worth, here's my code https://github.com/yonaskolb/XcodeGen/compare/master...teameh:scheme_management. I still needed to write out the plist to disk and add tests etc.. 

I took the same approach with TargetScheme but also added the other 2 properties (`orderHint` and `shared`) that were also added to xcodeproj in 8.5.0. I also added everything to [Scheme](https://github.com/teameh/XcodeGen/commit/ebf7fffeca8cd7e811d29cb5ff85aa76c2861080#diff-e30c15bc7cec124685379111f6d1d4ffffccf09b92070348d479786915bbb0a2R39-R54) because I thought users should be able to use this on `Scheme` as well when they need to use `Scheme` instead of `TargetScheme` when they need more level of control? 

Well maybe you could cherry pick the commit where I added [Scheme.Management](https://github.com/teameh/XcodeGen/commit/ebf7fffeca8cd7e811d29cb5ff85aa76c2861080) and combine our code? :)  

let me know if I can be of any help!

teameh:
@wendyliga anything I can do to help?

wendyliga:
@teameh can you test it?

teameh:
Hey @wendyliga you cherry picked only one commit, I think this won't compile. You need to cherry pick ebf7fffeca8cd7e811d29cb5ff85aa76c2861080 as well. 

Cheers!

wendyliga:
> Hey @wendyliga nice work! But you cherry picked only one commit, I think this won't compile. You need to cherry pick [ebf7fff](https://github.com/yonaskolb/XcodeGen/commit/ebf7fffeca8cd7e811d29cb5ff85aa76c2861080) as well. Cheers!

done

yageek:
Any update on this? How could we help to get it merged?

teameh:
Sorry I didn't make any time to test this. You could test the feature branch to see if everything is ready

huaqixue:
🙏🙏🙏，if you are free, please don't forget this PR, it will help me manage my schemes, thank you ! ! !

mbaikovets:
Any update on this? How could we help to get it merged?
@yonaskolb, please, take a look into the PR contents and help us to resolve this problem. 

It's much needed feature for large-scale projects.

Sources/ProjectSpec/Scheme.swift
```diff
@@ -31,6 +33,25 @@ public struct Scheme: Equatable {
...
+
+    public struct Management: Equatable {
+        public static let sharedDefault = true
+
+        public var shared: Bool
+        public var orderHint: Int?
+        public var isShown: Bool?
+
+        public init(
+            shared: Bool = true,
```
yonaskolb, line 47: Can we use `sharedDefault`

yonaskolb (commented):
Looks great all. Sorry, my review on this was still pending in Github.
A couple of things to add:
- a changelog entry
- a parsing test https://github.com/yonaskolb/XcodeGen/blob/master/Tests/ProjectSpecTests/SpecLoadingTests.swift
- some generation tests https://github.com/yonaskolb/XcodeGen/blob/master/Tests/XcodeGenKitTests/SchemeGeneratorTests.swift
- An example of setting these properties in the test project fixture https://github.com/yonaskolb/XcodeGen/blob/master/Tests/Fixtures/TestProject/project.yml

Sources/ProjectSpec/Scheme.swift
teameh: You mean this right? 

```
shared: Bool = Scheme.Management.sharedDefault
```

wendyliga: done https://github.com/yonaskolb/XcodeGen/pull/1142/commits/8edccc68abe50f44399480bff6757b48b99734db

Sources/XcodeGenKit/ProjectGenerator.swift
```diff
@@ -30,6 +30,25 @@ public class ProjectGenerator {
         let sharedData = XCSharedData(schemes: schemes)
         return XcodeProj(workspace: workspace, pbxproj: pbxProj, sharedData: sharedData)
     }
+    
+    public func generateSchemeManagement() -> XCSchemeManagement {
```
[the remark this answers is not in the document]
wendyliga: i have moved the function to SchemeGenerator. https://github.com/yonaskolb/XcodeGen/pull/1142/commits/5822e6252d7c56b350a555875a025b7c2c795741

wendyliga:
please help re review @yonaskolb

yonaskolb (commented):
Sorry for the delay on this. I've just left a comment above. Will this work for schemes defined on the project level?

Sources/XcodeGenKit/SchemeGenerator.swift
```diff
@@ -359,6 +359,24 @@ public class SchemeGenerator {
         )
     }
 
+    public static func generateSchemeManagement(project: Project) -> XCSchemeManagement {
+        let userStateSchemes = project.targets.map { target -> XCSchemeManagement.UserStateScheme in
+            XCSchemeManagement.UserStateScheme(
+                name: target.name + ".xcscheme",
```
yonaskolb, line 365: This won't work with `configVariants`, as the name isn't just the target name. This logic is container with `SchemeGenerator.generateSchemes` Maybe it's safer to loop over the generated schemes. Perhaps the `SchemeGenerator.generateSchemes` function could be split into `generateSchemes() throws -> [Scheme]` and `generateXCSchemes() throws -> [XCScheme]`. This could then use the former and get both all the schemes, and their proper names

teameh: > Maybe it's safer to loop over the generated schemes.

You mean the generated `Schema`, not `XCScheme` right? 

> Perhaps the SchemeGenerator.generateSchemes function could be split 

So something like this? 

    public func generateSchemes(project: Project) throws -> [Scheme] {
        // generating list of all schemas
    }

    public func generateXCSchemes(schemes: [Scheme]) throws -> [XCScheme] {
        // call generateScheme
    }

The only problem is, that the `generateScheme` function needs the optional `ProjectTarget` as well.. 

We could do something like this though: 

    static public func generateSchemes(project: Project) throws -> [(Scheme, ProjectTarget?)] {
        // generating list of all schemas
    }

    public func generateXCSchemes(schemes: [(Scheme, ProjectTarget?)]) throws -> [XCScheme] {
        try schemes.map { (schema, projectTarget) in
            try generateScheme(schema, for: projectTarget)
        }
    }

But this has another problem. `generateSchemeManagement` is currently not called in the `ProjectGenerator` but in the `GenerateCommand`. Which means that we need to call `generateSchemes` twice, which can't be good for performance.

Sources/ProjectSpec/TargetScheme.swift
```diff
@@ -22,6 +23,9 @@ public struct TargetScheme: Equatable {
     public var environmentVariables: [XCScheme.EnvironmentVariable]
     public var preActions: [Scheme.ExecutionAction]
     public var postActions: [Scheme.ExecutionAction]
+    public var shared: Bool
+    public var orderHint: Int?
+    public var isShown: Bool?
```
teameh, line 28: @wendyliga what was the reason again why we gave this struct 3 separate properties and `Scheme` a `management: Management` property? I think the api would be more straight forward if we keep this consistent?

teameh:
On my own fork I've merged the upstream master changes in and also updated the documentation index, maybe you could merge these changes in? See https://github.com/teameh/XcodeGen/commits/hide_scheme. 

@wendyliga Shall I also see if I can address the other changes or would you rather finish the PR yourself?

Sources/ProjectSpec/TargetScheme.swift
wendyliga: if I remember it correctly, I think it because of the `Scheme`  will create a directory called management, and for `TargetScheme`, it will only add new properties to existing file.

wendyliga:
> On my own fork I've merged the upstream master changes in and also updated the documentation index, maybe you could merge these changes in? See https://github.com/teameh/XcodeGen/commits/hide_scheme.
> 
> @wendyliga Shall I also see if I can address the other changes or would you rather finish the PR yourself?

sure, no problem

Sources/ProjectSpec/TargetScheme.swift
teameh: I don't think that is the case. Schemes should be written into different folders depending whether or not a scheme should be shared. 

Shared schemes should be written to: `*.xcodeproj/xcshareddata/xcschemes`
Non shared / personal schemes to: `*.xcodeproj/xcuserdata/<user-name>.xcuserdatad/xcschemes`

The plist with the management info should always be in: `*.xcodeproj/xcuserdata/<user-name>.xcuserdatad/xcschemes/xcschememanagement.plist`

So if we want to support the `shared` boolean option properly, we should move the schemes from the shared schemes folder to the personal one if it is set....  I'll see if I can add support for this in XcodeProj first.

teameh: Maybe it would be better to just remove the `shared: Bool` option for now? Since we're generating the schemes I wonder if setting the `shared` option would be used a lot since it would only determine where the generated scheme is going to be placed (`xcshareddata` or `xcuserdata`). I guess it would only make sense if you use XcodeGen to generate a project once that you're then going to check into git? 

Would you agree that it's fine to remove the option for now until there's native support for `xcuserdata` schemes at  XcodeProj? The alternatives are either waiting on XcodeProj support, or manually moving the generated schemes from xcshareddata if the `shared` option is set to `false`.

teameh: Maybe it would be better to just remove the `shared: Bool` option for now? Since we're generating the schemes I wonder if setting the `shared` option would be used a lot since it would only determine where the generated scheme is going to be placed (`xcshareddata` or `xcuserdata`). I guess it would only make sense if you use XcodeGen to generate a project once that you're then going to check into git? 

@wendyliga @yonaskolb would you agree that it's fine to not add the `shared` option for now until there's native support for `xcuserdata` schemes at  XcodeProj? The alternatives are either waiting on XcodeProj support, or manually moving the generated schemes from xcshareddata if the `shared` option is set to `false`.

teameh:
I've submitted a [PR to XcodeProj](https://github.com/tuist/XcodeProj/pull/739) to support `XCUserData` so we can write non shared / user schemes to: *.xcodeproj/xcuserdata/<user-name>.xcuserdatad/xcschemes. 

Using that code here makes to possible to fully support the `shared: bool` property and cleans up the code a bit. @wendyliga maybe you could take a look and merge my branch into yours if you agree this is the best way forward? https://github.com/teameh/XcodeGen/commits/hide_scheme 

I think the alternative would be to drop the `shared` option until XcodeProj merges the `XCUserData` PR. What do you say?

wendyliga:
@teameh done.
I also have added you to the branch so feel free to push directly to the branch

teameh:
Nice, thanks! 

Seems like the [XcodeProj PR](https://github.com/tuist/XcodeProj/pull/739) will take some tweaking before it will get merged. So if we want to move this forward we'll have to drop the `shared: Bool` support for now or just wait a while for the XcodeProj PR.

teameh:
> Seems like the https://github.com/tuist/XcodeProj/pull/739 will take some tweaking before it will get merged.

🥳 It's merged! Now lets wait for a release, hehe.

Sources/ProjectSpec/TargetScheme.swift
teameh: This is now done by XcodeProj \0/

Sources/XcodeGenKit/SchemeGenerator.swift
teameh: This has been addressed. The `generateSchemes` method now [returns a tuple of the shares schemes, the user schemes and the management](https://github.com/yonaskolb/XcodeGen/pull/1142/files#diff-5353e75c7700fffa19ecb24d2f9d5cae41af5fc25ef10e57c2219696c933e5f5R43-R47) which can then be used by the `ProjectGenerator` so it can be [written to the right files](https://github.com/yonaskolb/XcodeGen/pull/1142/files#diff-5b091a15a76b16e0bcb40ce6236420161ffd95493c4cb8414a81ec2e3f89f129R27-R36)

teameh:
> 🥳 It's merged! Now lets wait for a release, hehe.

It's already in! :O 

I've updated this branch and it all works great! 

@yonaskolb could you maybe do a re-review? I've also addressed your previous comments.

yonaskolb (approved):
Looks good. Great work everyone.
I'm slightly hesitant to add the concept of user data and having to get a username but looks like it's the only way to make a scheme hidden

yonaskolb:
Released in 2.34.0

teameh:
Awesome! Just updated our production app ymls and could finally hide/sort all the scheme mess. Works great, thanks for merging!

PR #1147, opened 2021-11-10
[PBXProjGenerator] Speed up `PBXProjGenerator.generate()`

jpsim:
By processing file groups and target generation concurrently leveraging `DispatchQueue.concurrentPerform(...)`.

On one particularly large project I benchmarked, this speeds up total generation time by 57%:

| | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| Before | 46.247 ± 4.589 | 39.267 | 53.397 | 1.57 ± 0.16 |
| After | 29.412 ± 0.773 | 28.487 | 31.060 | 1.00 |

I found and fixed data races one at a time by running XcodeGen with Thread Sanitizer enabled.

This is really just a proof of concept rather than something we can merge as-is for a few reasons:

1. ~I had to disable error propagation for getting file groups and target generation.~ Update: fixed in 2f75c953aa2d2126202ac6d83d74909b7d24557d
2. The mutation that happens in concurrent steps is hard to reason about. This was already the case before, but this change makes it worse. It's not easy to understand which mutations need to be performed on the mutation queue. This could be improved by using value types instead of reference types and performing copies rather than in-place mutation.

### DIFF
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -276,7 +276,7 @@ public class PBXProjGenerator {
             derivedGroups.append(group)
         }
 
-        mainGroup.children = Array(sourceGenerator.rootGroups)
+        mainGroup.children = sourceGenerator.rootGroupsArray
         sortGroups(group: mainGroup)
         setupGroupOrdering(group: mainGroup)
         // add derived groups at the end
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -1,3 +1,4 @@
+import Dispatch
 import Foundation
 import PathKit
 import ProjectSpec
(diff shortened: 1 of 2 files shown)


Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -222,7 +225,13 @@ public class PBXProjGenerator {
             pbxProject.projects = subprojects
         }
 
-        try project.targets.forEach(generateTarget)
+        // TODO: Generate targets concurrently
+        // At the moment, there's some non-determinism involved, leading to different framework orders in the
```
jpsim, line 229: @yonaskolb do you know where these are added? Sorting them at the end would probably fix this.

jpsim:
I don't think I'll have time to fix the non-determinism involved here. I took a stab at moving the source generator to use value types but the heavy use of PBXObject-derived classes and the object graph nature of the source tree makes it quite challenging to do in its entirety.

So I'm closing this PR for now. If I find time and an approach to complete this work I'll open a new PR.

PR #1169, opened 2022-02-02
Support test target for local Swift Package

freddi-kit:
Reopen https://github.com/yonaskolb/XcodeGen/pull/1074 for some reason
Close: #1050

PR #1074, opened 2021-05-05
Support test target for local Swift Package

freddi-kit:
Close: https://github.com/yonaskolb/XcodeGen/issues/1050

note: This PR needs to wait for releasing new version of tuist/XcodeProj with https://github.com/tuist/XcodeProj/pull/605

PR #1168, opened 2022-01-25
Remove TestableTargetReference and enhance TargetReference instead (related to #1074)

skofgar:
Hello,

This is related to #1074. I believe the changes added through `TestableTargetReference` could be made directly in `TargetReference` instead and make the code easier to maintain.

PR #1170, opened 2022-02-04
Remove TestableTargetReference and enhance TargetReference instead (related to #1169)

skofgar:
Also related to former #1074


skofgar:
This is replacing https://github.com/yonaskolb/XcodeGen/pull/1168

skofgar:
@yonaskolb asked me to update the `Docs.md`, I'm not sure yet what has to change there, so any pointers would be appreciated!

yonaskolb:
You can see the changes that were made here https://github.com/yonaskolb/XcodeGen/pull/1169/files#diff-40d41e90935483eb005b3c01dfd43f3a735ae19e9d3a2f615add07f07b136c05

Sources/ProjectSpec/SpecValidation.swift
```diff
@@ -159,7 +159,7 @@ extension Project {
                         if getProjectTarget(dependency.reference) == nil {
                             errors.append(.invalidTargetDependency(target: target.name, dependency: dependency.reference))
                         }
-                    case .project(let dependencyProjectName):
+                    case .project(let dependencyProjectName), .package(let dependencyProjectName):
```
freddi-kit, line 162: It is better to not check package by `getProjectReference`, it is out of range for `getProjectReference` and not preferred usage even it can check.  Please check by right way to check package or create  `getSwiftPackageReference`.

Sources/XcodeGenKit/SchemeGenerator.swift
```diff
@@ -107,6 +107,8 @@ public class SchemeGenerator {
                 }
                 pbxProj = try getPBXProj(from: projectReference)
                 projectFilePath = projectReference.path
+            case .package:
+                fatalError("Currently unsupported to use package for: \(target)")
```
freddi-kit, line 111: I think fatalError is not good for user to tell the error. You may can use throw error at here. Throwed error is shows as proper format as XcodeGen output Error.

Sources/XcodeGenKit/SchemeGenerator.swift
```diff
@@ -120,6 +122,8 @@ public class SchemeGenerator {
             switch target.location {
             case .project:
                 buildableName = pbxTarget.productNameWithExtension() ?? pbxTarget.name
+            case .package:
+                fatalError("Currently unsupported to use package for: \(target)")
```
freddi-kit, line 126: Ditto

freddi-kit (commented):
I think it is okay to merge it if @yonaskolb agree. However there are reason why I added `TestableTargetReference`. Please note them.

First, Swift Package is only supported for target reference of testing in my PR. If the supporting swift package in `TargetReference`, it also means supporting Swift Package in other all target place. It will accept Swift Package in target even if it cannot be used and it is very confusing for user of XcodeGen. So I prevent if by preparing by `TestableTargetReference` to support it as separated syntax to not allow `package` in `target`. 

Second, for making easy to maintain and read code. I know you want to merge them to make it easy to manage, but I also made `TestableTargetReference` for easy maintaining from other perspective. As I said above, I wanna make the role of target param separated in test and other place. I think `TargetReference` may has too exaggerated responsibility if it supports `package` param. It is because `TargetReference` object needs to be checked whether `package` is valid or not in many places (for example, you added fatalError in validation code in replacement in scheme generator).  It means `package` is okay to not be supported in `TargetReference` and useless. So I created `TestableTargetReference` make clear the difference of them to make it easy to read code.

They are what I mentioned "difficulty" in https://github.com/yonaskolb/XcodeGen/pull/1074#issuecomment-947664218

But as I said, It is also okay to merge it. And I added some comments.

freddi-kit:
Anyway, thank you for good suggestion

skofgar:
Thanks for all the feedback @yonaskolb and @freddi-kit. I'll work on updating my PR asap. Also, don't let this effort stop you from merging #1169. This can always be merged in later.

Also, thanks for the explanation why you introduced `TestableTargetReference`. If it's cleaner to keep `TestableTargetReference` and `TargetReference` apart, I don't want to impose this solution. Maybe @yonaskolb or someone else can weigh in on this regard.

Sources/XcodeGenKit/SchemeGenerator.swift
skofgar: Thanks for pointing that out! 

I'm actually not sure why that isn't supported, I wonder if I did a mistake when updating this.

I think I can add it by updating it to:

```
case .package(let packageName):
    guard let package = self.project.getPackage(packageName),
        case .local(let path) = package else {
        throw SchemeGenerationError.missingPackage(packageName)
    }
    return XCScheme.BuildableReference(
        referencedContainer: "container:\(path)",
        blueprintIdentifier: target.name,
        buildableName: target.name,
        blueprintName: target.name
    )
```

Sources/XcodeGenKit/SchemeGenerator.swift
skofgar: The solution above would resolve this as well.

Sources/ProjectSpec/SpecValidation.swift
skofgar: I updated this in the latest version, I hope this resolves this issue.

skofgar:
This isn't completely resolved as I had a unit test fail, because apparently a `Info.plist` included that shouldn't be. I'm investigating that and will add an update as soon as I figured out what the problem is.

Sources/XcodeGenKit/SchemeGenerator.swift
skofgar: I added my suggestion, because it's probably easier to review this way.

PR #1185, opened 2022-03-01
Skip compile sources for watch apps

evandcoleman:
Including an empty compile sources build phase for targets with the `application.watchapp2` product type causes a build error.


yonaskolb (approved):
Thanks @evandcoleman

yonaskolb:
As you can see from the test fixture diff, this was successfully building before. Is it on a certain version of Xcode that it's failing to build? CI is using 12.5.1 and 13.0 at the moment

evandcoleman:
The change to the test fixture was intentional. The compile sources build phase was removed for the watch target since these shouldn't be generated.

yonaskolb:
I mean that the project was compiling fine on CI before your change, but I can see now it wasn't compiling the watch app

PR #1197, opened 2022-03-23
Run target source pattern matching in parallel

alvarhansen:
As this transform closure does not access anything outside of its
closure and does not mutate any singletons, then it seems to be safe
to run this mapping in parallel.

In our project with 47 Targets and ~7000 files across them, the project generation time went from 6s to 3.2s:
```
> hyperfine 'current_xcodegen' 'parallel_xcodegen'
Benchmark 1: current_xcodegen
  Time (mean ± σ):      5.917 s ±  0.123 s    [User: 5.374 s, System: 2.368 s]
  Range (min … max):    5.757 s …  6.078 s    10 runs
 
Benchmark 2: parallel_xcodegen
  Time (mean ± σ):      3.210 s ±  0.045 s    [User: 5.603 s, System: 1.439 s]
  Range (min … max):    3.135 s …  3.276 s    10 runs
 
Summary
  'parallel_xcodegen' ran
    1.84 ± 0.05 times faster than 'current_xcodegen'
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - Fix Monterey macOS shell version, shell login flag for environments [#1167](https://github.com/yonaskolb/XcodeGen/issues/1167) @bimawa
 
+### Changed
+
+- Run target source pattern matching in parallel [#1197](https://github.com/yonaskolb/XcodeGen/pull/1197) @alvarhansen
+
 ## 2.27.0
 
 #### Added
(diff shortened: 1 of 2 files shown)


yonaskolb (commented):
Great, thanks @alvarhansen! Could you add a changelog entry too please

alvarhansen:
> Great, thanks @alvarhansen! Could you add a changelog entry too please

@yonaskolb, fixed. Added changelog entry.

yonaskolb (approved):
Thanks!

PR #1218, opened 2022-06-06
Example of vscode usage as a development environment

markst:
See: https://github.com/markst/hotreloading-vscode-ios

Using xcodegen a task can be setup to generate xcodeproj.

Combined with xcodebuild this means app's can be built & deployed to device without ever opening xcode.


yonaskolb (approved):
Very cool!

PR #1219, opened 2022-06-07
Fix `watchapp2-container` product name

lucabartoletti:
As far I can understand the right name is `application.watchapp2-container`.

This PR updates the documentation to reflect it.


yonaskolb (approved):
Good catch @lucabartoletti, thank you

PR #1221, opened 2022-06-13
Speed up SettingsBuilder

jpsim:
It's unnecessary to build up a whole grouped dictionary only to check if all platforms are identical and then immediately discard the dictionary.

Instead we can check if all targets match the first platform, which avoids creating a new dictionary but also allows bailing early as soon as a non-matching platform is found.

Generating a large project (36MB json spec) on an M1 Max machine leads to a ~6% total speedup: 28.48s vs 30.07s.


keith (approved):

yonaskolb (changes requested):
Cool, thanks @jpsim. Could you please add a changelog entry as well under `Next Version`.

Given your large amount of targets I imagine there are a wealth of small performance optimisations like this.

jpsim:
Changelog entry added

jpsim:
> Given your large amount of targets I imagine there are a wealth of small performance optimisations like this.

The number one thing we could do to optimise project generation would be to parallelise the 

From the 10s mark to the 45s mark in this trace, the bottleneck is recursively walking the directory tree in [SourceGenerator.getSourceChildren(...)](https://github.com/yonaskolb/XcodeGen/blob/2.29.0/Sources/XcodeGenKit/SourceGenerator.swift#L395-L415), which runs single-threaded.

image

I've tried parallelising that work but it's challenging given that the XcodeProj structure that gets built up at that stage uses a lot of reference types that are unsafe to modify from multiple threads. I've tried introducing locks and serial queues to address that, but haven't gotten it to work so far. I also tried to use Swift Concurrency to help parallise this processing stage safely, but the CLI library XcodeGen uses doesn't support async commands, so I gave up on that as well. [swift-argument-parser](https://github.com/apple/swift-argument-parser) does though, if you wanted to go down that route.

My next attempt would build up the directory tree in memory in a parallelised way and then use that tree for further work that is challenging to parallelise, but I'm not sure how much of a win we'll get from that.

yonaskolb (approved):
Yeah been thinking a move to [swift-argument-parser](https://github.com/apple/swift-argument-parser) would be good in the long run

asifmohd:
Thank you for this improvement @jpsim  👍  It has helped with an improvement of 25% for our project ❤️ 

> I've tried introducing locks and serial queues to address that, but haven't gotten it to work so far.

@jpsim Would be amazing if you could try out a build from this MR and share your feedback 🙏 

https://github.com/yonaskolb/XcodeGen/pull/863

I've been using a build from this branch for the past 2 years, and it has been super useful to save time spent generating projects. On an average I am seeing a 50% improvement in project generation times compared to master.

Issue #1224, opened 2022-06-23
Extension Kit extension support for Xcode14


freddi-kit:
(I'll work for it, this issue is for notifying to other developer)

Currently, we cannot create some new extensions from Xcode15 (e,g, App Intents). It is because the product type is new from Xcode 15 and XcodeGen does not create appropriate copy phase for this new product type.

alessionossa:
Unless you have access to a super-early-beta of next-year release of Xcode, I think you are referring to Xcode 14.

freddi-kit:
@alessionossa Thank you. Actually I'm asking my coworker to handle it. Also, it requires to change tuist/XcodeProj repo code and wait it for releasing new version. So please wait a moment 🙇

alessionossa:
I see that support for _ExtensionKit_ extensions has already been added there https://github.com/tuist/XcodeProj/issues/687.

freddi-kit:
Yes but we need to wait tuist to release new version

alessionossa:
https://github.com/tuist/XcodeProj/releases/tag/8.8.0 🎉

freddi-kit:
We are preparing it, please wait a moment

yonaskolb:
Resolved by https://github.com/yonaskolb/XcodeGen/pull/1228 and released in 2.30.0 👍

freddi-kit:
No, it is still not solved because it is necessary to customize copy framework path for extension kit.

yonaskolb:
Ok cool, didn't know. Are you still going to work on that @freddi-kit? That would be great

freddi-kit:
@yonaskolb my coworker @mtj0928 now published PR, we can review it

alessionossa:
Where can we expect a new release with the actual support for Extension Kit extensions?

alessionossa:
To completely support ExtensionKit Extension, I think we are still missing the correct project configuration to import Frameworks. See #1255.

Dennisdezhangaho:
this is not resolved yet, Xcode 14 for me to do this change, and cause my app crash. I am so confused till google guide me to this page. I think I just to back to Xcode 13.4.1

PR #1228, opened 2022-07-15
Fix XcodeGen building after XcodeProj update to 8.8.0

aleksproger:
**Reason**
- XcodeProj has been updated and has API breaking changes

**Content**
- Added new enum case handling in `Linkage`
- Renamed the enum case name for `XCWorkspaceDataFileRef.init`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,14 @@
 
 ## Next Version
 
-### Changed
+### Added
+- Added support for new target type `extensionkit-extension` in Xcode 14 #1228 @aleksproger
 
+### Changed
 - Speed up generating build settings for large projects #1221 @jpsim
-- Fix XcodeGen building after XcodeProj update to 8.8.0 #1228 @aleksproger
+
+### Fixed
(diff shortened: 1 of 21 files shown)


aleksproger:
@yonaskolb we should merge this and make a new release, because XcodeProg introduced an API breaking changes.
XcodeGen fails to build.
BTW I think it's a good idea to consider using .exact versions of dependencies to provide stable and reproducible builds

yonaskolb:
Thanks for this @aleksproger. I just added some small doc changes. This should also resolve https://github.com/yonaskolb/XcodeGen/issues/1224

Currently the `Package.resolved` tracks the exact version when building the package in isolation. The problem occurs when XcodeGen is pulled in as a library into another Package, and SPM ignores`Package.resolved`. I agree that having an exact version would help in some cases. Though doing that would prevent people using the library to update their version of XcodeProj in case there is a fix, which they can do now. Another way of resolving this is to list an explicit version of XcodeProj in your own package. 
So having a `from` feels like the lesser of two evils, though I could easily be convinced people would rather have it as an exact version. Happy for anyone else reading this to chime in too.

PR #1233, opened 2022-07-21
Fix data races in SettingsBuilder.

kotovmax:
**Reason:**
- `configFileSettings` and `settingPresetSettings` caches were just global dictionaries. Data races were taken place at these points when multiple projects are generated in parallel.

**Content:**
- Dictionaries for caches moved to NSCache which is thread safe.

### DIFF
--- a/Sources/XcodeGenKit/NSCacheExtensions.swift
+++ b/Sources/XcodeGenKit/NSCacheExtensions.swift
@@ -1,0 +1,37 @@
+import XcodeProj
+import Foundation
+
+enum Cached<T> {
+    case cached(T)
+    case nothing
+
+    var value: T? {
+        switch self {
+        case let .cached(value): return value
+        case .nothing: return nil
+        }
+    }
+}
+
+final class CacheContainer {
+    let value: Cached<BuildSettings>
+
+    init(value: Cached<BuildSettings>) {
+        self.value = value
+    }
+}
+
+extension NSCache where KeyType == NSString, ObjectType == CacheContainer {
+    subscript(aKey: String) -> Cached<BuildSettings>? {
+        get {
+            object(forKey: aKey as NSString)?.value
+        }
+        set {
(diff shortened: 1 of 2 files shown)


Sources/XcodeGenKit/NSCacheExtensions.swift
```diff
@@ -0,0 +1,37 @@
...
+
+    var value: T? {
+        switch self {
+        case let .cached(value): return value
+        case .nothing: return nil
+        }
+    }
+}
+
+final class CacheContainer {
```
aleksproger, line 16: Why do we need this container? Why not to store plain Cached<BuildSetting>?

aleksproger: And minor - do we actually need Cached enum? It just duplicates Optional.

aleksproger (approved):

Sources/XcodeGenKit/NSCacheExtensions.swift
kotovmax: > Why do we need this container? Why not to store plain Cached?
NSCache works only with ref types

> And minor - do we actually need Cached enum? It just duplicates Optional.
It was already there, let's keep it. 
IMO this adds more clarity at the point of use.

Sources/XcodeGenKit/NSCacheExtensions.swift
```diff
@@ -0,0 +1,37 @@
+import XcodeProj
```
aleksproger, line 1: Btw, maybe the proper place for this file is `XcodeGenCore` target
Because, `XcodeGenKit` target mostly contains public business logic, closely related to project generation

kotovmax:
@yonaskolb review this PR please) Thanks in advance!)

kotovmax:
@yonaskolb, please, review this PR)

PR #1187, opened 2022-03-03
Add checks for duplicate dependencies in targets

dalemyers:
Dependencies can have multiple properties set on them, but having the same type and reference is a collision and could generate unexpected results.

PR #1234, opened 2022-07-22
Added duplicate dependencies validation

aleksproger:
**Reason**
 - More strict validation of added dependencies

**Contents**
 - Added changelog entry
 - Added check for duplicates in validation stage
 - Added test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,18 +1,24 @@
 # Change Log
 
 ## Next Version
+
 ### Added
+
 - Added a new CopyFilesBuildPhase, "Embed ExtensionKit Extensions" #1230 @mtj0928
+- Added duplicate dependencies validation #1234 @aleksproger
 
 ## 2.30.0
 
 ### Added
+
 - Added support for new target type `extensionkit-extension` in Xcode 14 #1228 @aleksproger
 
 ### Changed
+
 - Speed up generating build settings for large projects #1221 @jpsim
 
 ### Fixed
+
 - Fix XcodeGen building as library after breaking XcodeProj update 8.8.0 #1228 @aleksproger
 
 ## 2.29.0
@@ -35,7 +41,7 @@ Some support for Xcode Test Plans has been added. For now test plans are not gen
 
 #### Added
 
-- Support for specifying custom group locations for SPM packages. #1173 @John-Connolly 
+- Support for specifying custom group locations for SPM packages. #1173 @John-Connolly
 
 ### Fixed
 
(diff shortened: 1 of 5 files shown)


aleksproger:
I saw that https://github.com/yonaskolb/XcodeGen/pull/1187 tried to introduce this feature, but got abandoned 
So, I also want to see that feature and made the similar validation, but a bit more optimized for example got rid of unnecessary iterations over all targets. Also, considered all the comments that were left there.

CHANGELOG.md
```diff
@@ -67,11 +73,12 @@ Some support for Xcode Test Plans has been added. For now test plans are not gen
 
 ### Changed
 
-- Speed up source inclusion checking for big projects #1122 @PaulTaykalo 
+- Speed up source inclusion checking for big projects #1122 @PaulTaykalo
 
 ## 2.25.0
 
 ### Added
+
```
yonaskolb, line 81: There are a lot of auto-reformatting changes here, but I'm ok with them, as it's a stricter markdown

Sources/ProjectSpec/SpecValidation.swift
```diff
@@ -154,38 +154,17 @@ extension Project {
...
-                    if packages[dependency.reference] == nil {
-                        errors.append(.invalidSwiftPackage(name: dependency.reference, target: target.name))
-                    }
-                default: break
+                let dependencyValidationErrors = try validate(dependency, in: target)
+                errors.append(contentsOf: dependencyValidationErrors)
+
+                if uniqueDependencies.contains(dependency) {
+                    errors.append(.duplicateDependencies(target: target.name, dependencyReference: dependency.reference))
+                    continue
```
yonaskolb, line 165: If a target has more than one duplicated dependency, this will bail out after the first one and not show all the errors. It will also prevent any further dependency validation errors from appearing. Do we need to `continue` here?

yonaskolb (commented):
Thanks for picking this up @aleksproger!

Sources/ProjectSpec/SpecValidation.swift
aleksproger: That is an inner continue statement, it will just continue to the next dependency, not the next target
It's just, but actually I can delete it, because I use if-else and there won't be extra insert to set attempt anyway

aleksproger: Removed the `continue` statement, because it only introduced ambiguity
And explicitly added cases which you mentioned to `ProjectSpecTests`

yonaskolb (approved):

freddi-kit:
@yonaskolb @aleksproger 
I would like to make this optional as option to pass to xcodegen command. it is kind of breaking change since validaiton rule is changed very much and I confirmed some project cannot be generated.

PR #1240, opened 2022-07-31
Correct name of package in example yaml

hisaac:
Hey there, I just noticed that one of the example yaml snippets had the wrong package name specified.


yonaskolb (approved):
Good catch thanks @hisaac

PR #1241, opened 2022-08-01
Fix resolving paths to avoid duplicate targets

rgnns:
Fixes duplicate targets found whenever include paths have different relative paths.


rgnns:
Can someone please take a look at this? This fixes an issue that is causing errors on our project. Thanks 🙏

yonaskolb:
Hi @rgnns, does https://github.com/yonaskolb/XcodeGen/pull/1275 also solve this issue?

rgnns:
Hi @yonaskolb , just tested with this change and it does solve our issue. However, worth pointing out the time it takes to generate our project went up from ~35s to ~65s.

Issue #1226, opened 2022-07-06
Generating schemes with BuildableProductRunnable for watchOS targets causes errors in Xcode 14 beta


SSheldon:
With an example `project.yml` like this:
``` yaml
name: Foo

targets:
  Foo:
    type: framework
    platform: watchOS

schemes:
  Foo:
    build:
      targets:
        Foo: build
```

And performing the following steps:
```
xcodegen -s project.yml
xcodebuild -project Foo.xcodeproj -scheme Foo -showBuildSettings
```

When using Xcode 14 beta 2, the final step errors with:
```
[MT] DVTAssertions: ASSERTION FAILURE in /System/Volumes/Data/SWE/Apps/DT/BuildRoots/BuildRoot8/ActiveBuildRoot/Library/Caches/com.apple.xbs/Sources/IDEFrameworks/IDEFrameworks-21257.0.0.0.25/IDEFoundation/Execution/RunDestinations/IDERunDestinationCLI.m:565
Details:  Unhandled/unexpected case where no run destinations were produced.
Object:   <IDERunDestinationCLI>
Method:   +resolveRunDestinationsWithWorkspace:scheme:buildAction:schemeCommand:schemeTask:destinationSpecifications:architectures:timeout:runDestinationManager:deviceManager:fallbackPreferredSDK:fallbackPreferredArchitectures:skipUnsupportedDestinations:shouldSkipRunDestinationValidation:didDisambiguate:disambiguatedMatches:disambiguatedMatchesDescription:error:
Thread:   <_NSMainThread: 0x60000213c280>{number = 1, name = main}
Hints: 

Backtrace:
  0   -[DVTAssertionHandler handleFailureInMethod:object:fileName:lineNumber:assertionSignature:messageFormat:arguments:] (in DVTFoundation)
  1   _DVTAssertionHandler (in DVTFoundation)
  2   _DVTAssertionFailureHandler (in DVTFoundation)
  3   _sortDevicesForDisplay (in IDEFoundation)
  4   -[Xcode3CommandLineBuildTool _resolveRunDestinationsForBuildAction:] (in Xcode3Core)
  5   -[Xcode3CommandLineBuildTool _resolveInputOptionsWithTimingSection:] (in Xcode3Core)
  6   -[Xcode3CommandLineBuildTool run] (in Xcode3Core)
  7   XcodeBuildMain (in libxcodebuildLoader.dylib)
  8   start (in dyld)
```

[...]

PR #1245, opened 2022-08-11
Fix profile action to not run frameworks

SSheldon:
This change applies the logic added for the launch action in #328 to the profile action. Profile actions that attempt to run frameworks just crash instruments; for watchOS frameworks, they even cause Xcode 14 betas to fail to build, as described in #1226.

## Validation

Using Xcode 13.0, I created schemes for a bunch of different kinds of targets and observed the profile action in the scheme it generated. I then opened these schemes with Xcode 13.3 and 14.0b5 and confirmed the action didn't change.

The following types of targets had build product runnables in the profile scheme:
* ios app
* watch app
* extension
* command line tool

And the following types had buildable reference macro expansions: 
* framework
* library
* metal library
* bundle
* xpc service

These line up with the logic for `shouldExecuteOnLaunch` used for the launch action.

I also ran tests, confirmed they passed, and added the only fixture change.

Testing in the Xcode 14 betas has confirmed that a change like this to the scheme for a watchOS framework fixes the build issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Fixed
 
-- Fix profile action to not run frameworks @SSheldon
+- Fix profile action to not run frameworks #1245 @SSheldon
 
 ## 2.31.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


yonaskolb (approved):
Great, thanks @SSheldon for your investigation into this and the fix

Issue #1246, opened 2022-08-11
Scheme not generated for Aggregate Target


CraigSiemens:
Our project/workspace has `IDEWorkspaceSharedSettings_AutocreateContextsIfNeeded` set to false to prevent Xcode from autogenerating schemes, to avoid opening the project being really slow and enforce that all the schemes come from the project spec.

If we add an aggregate target, a scheme will not be generated for it.

```yml
name: Example
configs:
  Debug: debug
  Release: release
targets:
  Foo:
    type: bundle
    platform: macOS
    scheme:
      environmentVariables:
        MY_ENV_VAR: VALUE
aggregateTargets:
  Run-Script:
    buildScripts:
      - name: Run script
        script: |
          echo("Hello")
    scheme:
      environmentVariables:
        MY_ENV_VAR: VALUE
```

## Steps to reproduce
1. Run `xcodegen` on the above `project.yml`.
2. Open the project
3. Go to Project > Schemes > Manage Schemes...
4. Uncheck "Autocreate schemes"
    <img width="25%" alt="Padded Screen Shot 2022-08-11 at 10 23 28 AM" src="https://user-images.githubusercontent.com/680442/184182215-44f082ec-de10-4a41-a1dd-62fea21a7556.png">
5. Run xcodegen again

## Expected Result
There should be a scheme generated for the aggregate target `Run-Script`.

## Actual Result
There was no scheme generated.

## Workaround
A scheme has to be added to `schemes` for it to be generated. It still doesnt use any of the settings from scheme under [Aggregate Target](https://github.com/yonaskolb/XcodeGen/blob/master/Docs/ProjectSpec.md#aggregate-target)
```yml
schemes:
  Run-Script:
    build:
      targets:
        Run-Script: all
```

PR #1250, opened 2022-08-19
Added scheme generation for aggregate targets

CraigSiemens:
- #1246

## Changes
- Updated `SchemeGenerator` to generate schemes for all `projectTargets`.
  - Added product type to the `ProjectTarget` protocol.
  - Updated `SchemeGenerator` to work with `ProjectTarget`s instead of `Target`s.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,9 @@
 # Change Log
 
 ## Next Version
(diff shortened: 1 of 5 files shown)


yonaskolb (approved):
Nice one, thanks @CraigSiemens

Issue #1139, opened 2021-10-15
Can I disable GPU Frame Capture?


kkk669:
I know GPU Frame Capture is useful, but sometimes [it causes a crash of SCNView's renderer](https://developer.apple.com/forums/thread/670852?answerId=655717022#655717022). So, I want XcodeGen to generate an `xcodeproj` file that GPU Frame Capture is disabled. Is that possible?

| | |
|-|-|
| <img width="925" alt="Screen Shot 2021-10-15 at 9 31 06 PM" src="https://user-images.githubusercontent.com/601636/137489070-0630f0b0-790f-41e9-8597-51eff8a70226.png"> | <img width="929" alt="Screen Shot 2021-10-15 at 9 31 23 PM" src="https://user-images.githubusercontent.com/601636/137488807-5fc1a20d-b201-4c92-a062-d9b975bd4078.png"> |

If not, I would like to request that feature.

If I select "Disabled" for "GPU Frame Capture" on the scheme settings in Xcode 13.0 (13A233), `*.xcodeproj/xcshareddata/xcschemes/*.xcscheme` will be changed like:

```diff
*** before	2021-10-15 21:23:01.000000000 +0900
--- after	2021-10-15 21:22:51.000000000 +0900
***************
*** 32,51 ****
--- 32,52 ----
     </TestAction>
     <LaunchAction
        buildConfiguration = "Debug"
        selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
        selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
        launchStyle = "0"
        useCustomWorkingDirectory = "NO"
        ignoresPersistentStateOnLaunch = "NO"
        debugDocumentVersioning = "YES"
        debugServiceExtension = "internal"
+       enableGPUFrameCaptureMode = "3"
        allowLocationSimulation = "YES">
        <BuildableProductRunnable
           runnableDebuggingMode = "0">
           <BuildableReference
              BuildableIdentifier = "primary"
              BlueprintIdentifier = "89050A44268B21FE0026DB69"
```

`enableGPUFrameCaptureMode` specs (Xcode 13.0 (13A233)):

- If "Automatically" is selected (default), there is no `enableGPUFrameCaptureMode`
- If "Metal" is selected, `enableGPUFrameCaptureMode  = "1"`
- If "OpenGL ES" is selected, `enableGPUFrameCaptureMode  = "2"`
[...]

PR #1251, opened 2022-08-22
Add enableGPUFrameCaptureMode to Scheme

bsudekum:
This implements https://github.com/yonaskolb/XcodeGen/issues/1139 and allows control over `GPU Frame Capture`.

### DIFF
--- a/Docs/ProjectSpec.md
+++ b/Docs/ProjectSpec.md
@@ -657,6 +657,7 @@ This is a convenience used to automatically generate schemes for a target based 
 - [ ] **testTargets**: **[[Test Target](#test-target)]** - a list of test targets that should be included in the scheme. These will be added to the build targets and the test entries. Each entry can either be a simple string, or a [Test Target](#test-target)
 - [ ] **gatherCoverageData**: **Bool** - a boolean that indicates if this scheme should gather coverage data. This defaults to false
 - [ ] **coverageTargets**: **[[Testable Target Reference](#testable-target-reference) - a list of targets to gather code coverage. Each entry can either be a simple string, a string using [Project Reference](#project-reference) or [Testable Target Reference](#testable-target-reference)
+- [ ] **enableGPUFrameCaptureMode**: **GPUFrameCaptureMode** - Property value set for `GPU Frame Capture`.
 - [ ] **disableMainThreadChecker**: **Bool** - a boolean that indicates if this scheme should disable the Main Thread Checker. This defaults to false
 - [ ] **stopOnEveryMainThreadCheckerIssue**: **Bool** - a boolean that indicates if this scheme should stop at every Main Thread Checker issue. This defaults to false
 - [ ] **buildImplicitDependencies**: **Bool** - Flag to determine if Xcode should build implicit dependencies of this scheme. By default this is `true` if not set.
--- a/Docs/ProjectSpec.md
+++ b/Docs/ProjectSpec.md
@@ -657,7 +657,6 @@ This is a convenience used to automatically generate schemes for a target based 
 - [ ] **testTargets**: **[[Test Target](#test-target)]** - a list of test targets that should be included in the scheme. These will be added to the build targets and the test entries. Each entry can either be a simple string, or a [Test Target](#test-target)
 - [ ] **gatherCoverageData**: **Bool** - a boolean that indicates if this scheme should gather coverage data. This defaults to false
 - [ ] **coverageTargets**: **[[Testable Target Reference](#testable-target-reference) - a list of targets to gather code coverage. Each entry can either be a simple string, a string using [Project Reference](#project-reference) or [Testable Target Reference](#testable-target-reference)
-- [ ] **enableGPUFrameCaptureMode**: **GPUFrameCaptureMode** - Property value set for `GPU Frame Capture`. Possible values are `autoEnabled`, `metal`, `openGL`, `disabled`. Default is `autoEnabled`.
 - [ ] **disableMainThreadChecker**: **Bool** - a boolean that indicates if this scheme should disable the Main Thread Checker. This defaults to false
 - [ ] **stopOnEveryMainThreadCheckerIssue**: **Bool** - a boolean that indicates if this scheme should stop at every Main Thread Checker issue. This defaults to false
 - [ ] **buildImplicitDependencies**: **Bool** - Flag to determine if Xcode should build implicit dependencies of this scheme. By default this is `true` if not set.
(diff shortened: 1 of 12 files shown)


Docs/ProjectSpec.md
```diff
@@ -657,6 +657,7 @@ This is a convenience used to automatically generate schemes for a target based
 - [ ] **testTargets**: **[[Test Target](#test-target)]** - a list of test targets that should be included in the scheme. These will be added to the build targets and the test entries. Each entry can either be a simple string, or a [Test Target](#test-target)
 - [ ] **gatherCoverageData**: **Bool** - a boolean that indicates if this scheme should gather coverage data. This defaults to false
 - [ ] **coverageTargets**: **[[Testable Target Reference](#testable-target-reference) - a list of targets to gather code coverage. Each entry can either be a simple string, a string using [Project Reference](#project-reference) or [Testable Target Reference](#testable-target-reference)
+- [ ] **enableGPUFrameCaptureMode**: **GPUFrameCaptureMode** - Property value set for `GPU Frame Capture`.
```
keith, line 660: can you include the possible values?

bsudekum: Added https://github.com/yonaskolb/XcodeGen/pull/1251/commits/cf29bd17cb5da6201383e5e9059885210adff11b

keith (dismissed):

Docs/ProjectSpec.md
keith: wait are these the right values, or do they have to be the corresponding numbers?

keith: the docs should probably mention that this is only valid for the run scheme settings

bsudekum: Yeah these are the right values. The string backed enum was added here: https://github.com/tuist/xcodeproj/pull/368.

Tests/ProjectSpecTests/SpecLoadingTests.swift
```diff
@@ -870,6 +871,7 @@ class SpecLoadingTests: XCTestCase {
                 try expect(scheme.build.parallelizeBuild) == false
                 try expect(scheme.build.buildImplicitDependencies) == false
                 try expect(scheme.build.runPostActionsOnFailure) == true
+                try expect(scheme.test?.enableGPUFrameCaptureMode) == .metal
```
keith, line 874: this isn't actually supported in test mode is it? I think this should be moved to the run section of the test instead?

bsudekum: Moved in https://github.com/yonaskolb/XcodeGen/pull/1251/commits/303c4c24d616559d742e4fb018388ce392f1ce33

Makefile
```diff
@@ -1,6 +1,6 @@
 TOOL_NAME = XcodeGen
 export EXECUTABLE_NAME = xcodegen
-VERSION = 2.32.0
+VERSION = 2.32.0-patch-3
```
keith, line 3: these version changes should be reverted

README.md
```diff
@@ -113,7 +113,7 @@ swift run xcodegen
 Add the following to your Package.swift file's dependencies:
 
 ```swift
-.package(url: "https://github.com/yonaskolb/XcodeGen.git", from: "2.32.0"),
+.package(url: "https://github.com/yonaskolb/XcodeGen.git", from: "2.32.0-patch-3"),
```
keith, line 116: revert

Sources/XcodeGen/main.swift
```diff
@@ -3,6 +3,6 @@ import ProjectSpec
 import XcodeGenCLI
 import Version
 
-let version = Version("2.32.0")
+let version = Version("2.32.0-patch-3")
```
keith, line 6: revert

Docs/ProjectSpec.md
```diff
@@ -657,6 +657,7 @@ This is a convenience used to automatically generate schemes for a target based
 - [ ] **testTargets**: **[[Test Target](#test-target)]** - a list of test targets that should be included in the scheme. These will be added to the build targets and the test entries. Each entry can either be a simple string, or a [Test Target](#test-target)
 - [ ] **gatherCoverageData**: **Bool** - a boolean that indicates if this scheme should gather coverage data. This defaults to false
 - [ ] **coverageTargets**: **[[Testable Target Reference](#testable-target-reference) - a list of targets to gather code coverage. Each entry can either be a simple string, a string using [Project Reference](#project-reference) or [Testable Target Reference](#testable-target-reference)
+- [ ] **enableGPUFrameCaptureMode**: **GPUFrameCaptureMode** - Property value set for `GPU Frame Capture`. Possible values are `autoEnabled`, `metal`, `openGL`, `disabled`. Default is `autoEnabled`.
```
keith, line 660: in my testing the values mentioned here are not actually valid, and you have to use the numbers, can you add a test using these strings instead of their numeric values?

bsudekum: Do you think this would be covered by this?

https://github.com/yonaskolb/XcodeGen/blob/9422de9925fcd5701aa0b4e12f7b48582b8e2866/Tests/ProjectSpecTests/SpecLoadingTests.swift#L842

Sources/ProjectSpec/Scheme.swift
```diff
@@ -864,3 +876,34 @@ extension XCScheme.EnvironmentVariable: JSONEncodable {
...
+        case "autoEnabled":
+            return .autoEnabled
+        case "metal":
+            return .metal
+        case "openGL":
+            return .openGL
+        case "disabled":
+            return .disabled
+        default:
+            assert(false, "Invalid enableGPUFrameCaptureMode value. Valid values are: autoEnabled, metal, openGL, disabled")
```
keith, line 905: i wonder if this should fail in some way that the user sees? otherwise they may not realize they have an invalid value, since this won't fail in release, what do other places like this do?

keith (approved):
looks good!

Sources/ProjectSpec/Scheme.swift
bsudekum: looks like fatalError is preferred: https://github.com/search?q=repo%3Ayonaskolb%2FXcodeGen%20fatalError&type=code

bsudekum: updated: https://github.com/yonaskolb/XcodeGen/pull/1251/commits/d5aa44faba4052e75f67330cdfed0078c871dd56

bsudekum:
@yonaskolb do you think you could review this?

yonaskolb (approved):
Looks good thanks @bsudekum

PR #1255, opened 2022-08-26
Add LD_RUNPATH_SEARCH_PATHS to extensionkit-extension preset

alessionossa:
Add `LD_RUNPATH_SEARCH_PATHS` option as it is for other types of extensions (e.g. app extension).
I followed the same logic of #272, copying the value from the current version of [SettingPresets/Products/app-extension.yml](https://github.com/yonaskolb/XcodeGen/blob/master/SettingPresets/Products/app-extension.yml)

### DIFF
--- a/SettingPresets/Products/extensionkit-extension.yml
+++ b/SettingPresets/Products/extensionkit-extension.yml
@@ -1,0 +1,1 @@
+LD_RUNPATH_SEARCH_PATHS: ["$(inherited)", "@executable_path/Frameworks", "@executable_path/../../Frameworks"]


freddi-kit:
I started thinking it also should be fixed by XcodeProj (yes, this pr also should be okay), could you check it if you have time?

https://github.com/tuist/XcodeProj/blob/b6de1bfe021b861c94e7c83821b595083f74b997/Sources/XcodeProj/Utils/BuildSettingsProvider.swift#L224-L229

freddi-kit (approved):

alessionossa:
> I started thinking it also should be fixed by XcodeProj (yes, this pr also should be okay), could you check it if you have time?
> 
> https://github.com/tuist/XcodeProj/blob/b6de1bfe021b861c94e7c83821b595083f74b997/Sources/XcodeProj/Utils/BuildSettingsProvider.swift#L224-L229

I see that `targetSettings(platform:, product:)` is only called by `targetDefault(variant:, platform:, product:, swift:)` and we are not calling the latter method anywhere. Are you think of using that instead of "custom overrides"?

yonaskolb:
Thanks @alessionossa. Could you add a changelog entry as well please.

And at the moment we aren't using the BuildSettingsProvider from XcodeProj, as that didn't exist at the time. I've created a new issue to track that here https://github.com/yonaskolb/XcodeGen/issues/1267

Issue #1261, opened 2022-09-08
Including yml files with the same `filePath` causes the second to be ignored.


CraigSiemens:
Given the following project setup, `OutgoingTransferDetail.yml` will not be loaded causing targets that depend on it to fail.

## Steps to Reproduce

```
.
├── project.yml
└── Modules
    ├── Modules.yml
    ├── Foo
    │   └── main.swift
    ├── Incoming
    │   ├── Incoming.yml
    │   └── Transfer
    │       ├── Transfer.yml
    │       └── IncomingTransferDetail
    │           ├── IncomingTransferDetail.yml
    │           └── Sources
    │               └── main.swift
    └── Outgoing
        ├── Outgoing.yml
        └── Transfer
            ├── Transfer.yml
            └── OutgoingTransferDetail
                ├── OutgoingTransferDetail.yml
                └── Sources
                    └── main.swift
```

Each folder that has a `.yml` file only contains includes for its subdirectories.

For example `Module.yml` is just
```
include:
- Incoming/Incoming.yml
- Outgoing/Outgoing.yml
```

`Foo`, `IncomingTransferDetail`, and `OutgoingTransferDetail` are targets. `Foo` depends on the other two.

## Expected Result

The project should load with a `Foo` target that depends on the `IncomingTransferDetail` and `OutgoingTransferDetail` targets.

## Actual Result
```
Spec validation error: Target "Foo" has invalid dependency: "OutgoingTransferDetail"
Program ended with exit code: 1
```

## Additional Details
It seems to be ab issue with how `SpecFile` tracks `mergedTargets`. It only uses the file name which fails because `Modules/Incoming/Transfer/Transfer.yml` and `Modules/Outgoing/Transfer/Transfer.yml` both get a `filePath` of `Transfer/Transfer.yml` when using relative paths.

I'me working on a PR for this but just wanted to capture the details for the issue here.

PR #1262, opened 2022-09-08
Handle imports with same relative path

CraigSiemens:
- #1261

## Changes
- Identify merged includes using the full path to the file rather than just `filePath` which can be relative to the `basePath` and not unique.
- Added comments to the path properties to make it clearer what they contain.
- Other minor cleanup (renaming a parameter and making things private).

### DIFF
--- a/Sources/ProjectSpec/Project.swift
+++ b/Sources/ProjectSpec/Project.swift
@@ -155,8 +155,7 @@ extension Project: Equatable {
 extension Project {
 
     public init(path: Path) throws {
-        var cachedSpecFiles: [Path: SpecFile] = [:]
-        let spec = try SpecFile(path: path, cachedSpecFiles: &cachedSpecFiles)
+        let spec = try SpecFile(path: path)
         try self.init(spec: spec)
     }
 
--- a/Sources/ProjectSpec/Project.swift
+++ b/Sources/ProjectSpec/Project.swift
@@ -156,13 +156,7 @@ extension Project {
 
     public init(path: Path) throws {
         var cachedSpecFiles: [Path: SpecFile] = [:]
-        let spec = try SpecFile(filePath: path, basePath: path.parent(), cachedSpecFiles: &cachedSpecFiles)
-        try self.init(spec: spec)
-    }
-
-    public init(path: Path, basePath: Path) throws {
-        var cachedSpecFiles: [Path: SpecFile] = [:]
-        let spec = try SpecFile(filePath: path, basePath: basePath, cachedSpecFiles: &cachedSpecFiles)
+        let spec = try SpecFile(path: path, cachedSpecFiles: &cachedSpecFiles)
         try self.init(spec: spec)
     }
 
--- a/Sources/ProjectSpec/SpecFile.swift
+++ b/Sources/ProjectSpec/SpecFile.swift
@@ -53,14 +53,21 @@ public struct SpecFile {
             dictionary[key] as? Bool ?? (dictionary[key] as? NSString)?.boolValue
         }
     }
-
-    public init(path: Path, basePath: Path? = nil, cachedSpecFiles: inout [Path: SpecFile], variables: [String: String] = [:]) throws {
-        let basePath = basePath ?? path.parent()
+    
+    /// Create a SpecFile for a Project
+    /// - Parameters:
+    ///   - path: The absolute path to the spec file
+    ///   - projectRoot: The root of the project to use as the base path. When nil, uses the parent
+    ///     of the path.
+    public init(path: Path, projectRoot: Path? = nil, variables: [String: String] = [:]) throws {
+        let basePath = projectRoot ?? path.parent()
         let filePath = try path.relativePath(from: basePath)
+        var cachedSpecFiles: [Path: SpecFile] = [:]
         
         try self.init(filePath: filePath, basePath: basePath, cachedSpecFiles: &cachedSpecFiles, variables: variables)
     }
-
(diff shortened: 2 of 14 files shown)


CraigSiemens:
I wouldn't mind some feedback on whether tests should be added to this (I assume they should) and what they would look like. I think it'd either need a new fixture or an existing on to be updated but I don't know the preferred path.

yonaskolb:
Thanks @CraigSiemens, some great changes. Does https://github.com/yonaskolb/XcodeGen/pull/1275 also solve your issue? If so perhaps we can merge that first, and you can add your documentation on top.

In terms of a test case, adding it to `Tests/Fixtures/paths_test.yml` is probably a good place

CraigSiemens:
I tried it out and https://github.com/yonaskolb/XcodeGen/pull/1275 does fix the issue too.

I'm working on rebasing my branch onto master. I did find some improvements for the `cachedSpecFiles` that I'll add into this branch.

Sources/ProjectSpec/Project.swift
```diff
@@ -155,14 +155,7 @@ extension Project: Equatable {
...
     public init(path: Path) throws {
-        var cachedSpecFiles: [Path: SpecFile] = [:]
-        let spec = try SpecFile(filePath: path, basePath: path.parent(), cachedSpecFiles: &cachedSpecFiles)
-        try self.init(spec: spec)
-    }
-
-    public init(path: Path, basePath: Path) throws {
-        var cachedSpecFiles: [Path: SpecFile] = [:]
-        let spec = try SpecFile(filePath: path, basePath: basePath, cachedSpecFiles: &cachedSpecFiles)
+        let spec = try SpecFile(path: path)
```
CraigSiemens, line 158: This is revering a change from https://github.com/yonaskolb/XcodeGen/pull/1275. Internally the init creates the `cachedSpecFiles` so callers don't have to make it.

Sources/ProjectSpec/SpecFile.swift
```diff
@@ -63,26 +79,28 @@ public struct SpecFile {
...
+
         let includes = Include.parse(json: jsonDictionary["include"])
         let subSpecs: [SpecFile] = try includes
             .filter(\.enable)
             .map { include in
-                if let specFile = cachedSpecFiles[filePath] {
-                    return specFile
-                } else {
-                    return try SpecFile(include: include, basePath: basePath, relativePath: relativePath, cachedSpecFiles: &cachedSpecFiles, variables: variables)
-                }
```
CraigSiemens, line 81: `filePath`, which is for the current spec, is not the correct key to use when looking for the `specFile` of an `include` that has it's own path.

CraigSiemens:
@yonaskolb I believe I'm done rebasing/fixing so it's free for a re-review when you have a moment.

I've added in tests to `paths_test.yml` and verified they they failed before any of these changes were added to mater.

I also made a couple minor changes to the usage of `cachedSpecFiles`, I left a reason for some of them above.

yonaskolb:
Thanks @CraigSiemens, I can hopefully get to this soon.

@ma-oli would you mind casting an eye over this too, as you were recently in this area.

ma-oli:
I’m off right now until Monday but I’m happy to have a look when I’m back.

ma-oli:
Hey; sorry about the delay. I'm looking at it now and I'm seeing some issues. I'll keep you posted.

yonaskolb:
This issue has also come in https://github.com/yonaskolb/XcodeGen/issues/1299

Sources/ProjectSpec/SpecFile.swift
```diff
@@ -48,8 +54,11 @@ public struct SpecFile {
         }
     }
 
-    public init(path: Path, cachedSpecFiles: inout [Path: SpecFile], variables: [String: String] = [:]) throws {
-        try self.init(filePath: path, basePath: path.parent(), cachedSpecFiles: &cachedSpecFiles, variables: variables)
+    public init(path: Path, basePath: Path? = nil, cachedSpecFiles: inout [Path: SpecFile], variables: [String: String] = [:]) throws {
+        let basePath = basePath ?? path.parent()
+        let filePath = try path.relativePath(from: basePath)
```
ma-oli, line 64: Using a relative path ( "-r ." for instance) as a project root results in this line throwing an unmatchedAbsolutePath error. We might want to use basePath?.absolute(), either here or in SpecLoader.

ma-oli (changes requested):
I think it's good but this breaks using a relative commit as a project root. Shouldn't be a big deal to fix I think.

Sources/ProjectSpec/SpecFile.swift
CraigSiemens: That's interesting, I'll try and take a look at that.

CraigSiemens:
@ma-oli @yonaskolb I've added the fix for handling relative project paths.

ma-oli (approved):

CraigSiemens:
@yonaskolb Any chance this could get merged since it's got an approval?

yonaskolb:
> @yonaskolb Any chance this could get merged since it's got an approval?

Yep, could you please add a changelog entry. Also does this fix https://github.com/yonaskolb/XcodeGen/issues/1299?

CraigSiemens:
Yeah, it should fix that issue. I've merged in master and updated the changelog.

yonaskolb (approved):
Fantastic, thanks for all your work on this @CraigSiemens 🙏

PR #1268, opened 2022-09-29
adding support for adding files that don't belong to a target

skofgar:
For some projects it might be helpful to add files to a project that don't belong to a specific target. 

For example, `README.md` files or the `project.yml`.

## Known Issues

* Currently I just use `TargetSources` to achieve this capability. If the general direction of this PR is acceptable, then I would - in a separate PR - rename `TargetSources` to `SourcesReference` (or something similar)
* I need to add unit tests for this

### DIFF
--- a/Docs/ProjectSpec.md
+++ b/Docs/ProjectSpec.md
@@ -25,6 +25,7 @@
 - [Scheme](#scheme)
 - [Scheme Template](#scheme-template)
 - [Swift Package](#swift-package)
+- [Files](#files)
 
 ## General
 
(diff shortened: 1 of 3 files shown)


skofgar:
Hi @yonaskolb may I request a review for this PR?

yonaskolb:
Interestingly this is what `fileGroups` is for. The name doesn't suggest it can be used for single files though. Are the options exposed by `TargetSource` required for your use case? If not, perhaps we just deprecate and rename `fileGroups` to `files`, while still supporting `fileGroups` for backwards compatibility.

skofgar:
Oh really 🤦‍♂️, sorry I didn’t realize it. 

I do have a use case where the `include` filter options would be really helpful. 

I’ll look into `fileGroups` more and update this.

PR #1270, opened 2022-09-30
allow multiple spec files to be provided to XcodeGen

skofgar:
**TL;DR: Generate 10 projects in less than 1 second instead of 8 by allowing XcodeGen to accept multiple project files instead of generating them one by one.**

When generating multiple projects it can take a long time. For example, when generating 10 projects by invoking XcodeGen 10 times sequentially via bash, the script takes about `8 seconds`, even with caching turned on. 

## Why is this a problem?

When switching branches and checking out new branches of a repo that contains multiple XcodeGen project files, this causes substantial delays and slows down engineers.

## Investigation

Looking at the profiler, it looks like the slowest part is just starting up XcodeGen. The "initialization" period can take up to `0.5 seconds`.

## Proposed Solution

Allowing XcodeGen to accept multiple spec files, can significantly speed up the generation. In our tests generation of the same 8 projects was less than `3 seconds` without caching and with caching less than `1 second` (sometimes just 0.2 seconds).

## Approach

 * allow specs to be provided as a comma separated list of specs instead of just one

```
xcodegen --spec path/to/first/project.yml,path/to/another/project.yml --use-cache
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 
 - Added support for `enableGPUFrameCaptureMode` #1251 @bsudekum
+- Added ability to generate multiple projects in one XcodeGen launch #1270 @skofgar
 
 ### Fixed
 
--- a/README.md
+++ b/README.md
@@ -130,7 +130,7 @@ This will look for a project spec in the current directory called `project.yml` 
 
 Options:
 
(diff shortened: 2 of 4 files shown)


skofgar:
Hi @yonaskolb may I request a review for this PR?

yonaskolb (approved):
Very cool, thanks @skofgar!

PR #1275, opened 2022-10-11
Fix includes related issues and improve their performances

ma-oli:
This pull request includes the following changes, related to the way XcodeGen processes includes:
- Fix recursive include path when relativePath is not set
- Include projectRoot in include paths
- Use memoization during recursive SpecFiles creation. This fixes the poor performance with lots of recursive includes.


ma-oli:
Gentle reminder?

yonaskolb:
These are great changes, thank you @ma-oli!
Would this also solve the issue that PR https://github.com/yonaskolb/XcodeGen/pull/1262 attempts to tackle?

Would be great if you could add some change logs as well. You can have multiple entries as this fixes multiple things

ma-oli:
I just updated the Changelog as well. I'm not sure if that fixes #1262; it does seem very similar so it's very possible.

yonaskolb (approved):

PR #1288, opened 2022-11-09
Add `--render-markdowns` command flag

MojtabaHs:
This flag generates the `.xcodesamplecode.plist` inside the project file and cause Xcode to render the markdown files with `.md` extension

### DIFF
--- a/Sources/XcodeGenCLI/Commands/GenerateCommand.swift
+++ b/Sources/XcodeGenCLI/Commands/GenerateCommand.swift
@@ -22,6 +22,9 @@ class GenerateCommand: ProjectCommand {
 
     @Flag("--only-plists", description: "Generate only plist files")
     var onlyPlists: Bool
+    
+    @Flag("--render-markdowns", description: "Render markdown files with `.md` extension")
+    var renderMarkdowns: Bool
 
     init(version: Version) {
         super.init(version: version,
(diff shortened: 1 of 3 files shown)


Sources/XcodeGenCLI/Commands/GenerateCommand.swift
```diff
@@ -22,6 +22,9 @@ class GenerateCommand: ProjectCommand {
 
     @Flag("--only-plists", description: "Generate only plist files")
     var onlyPlists: Bool
+    
+    @Flag("--render-markdowns", description: "Render markdown files with `.md` extension")
+    var renderMarkdowns: Bool
```
CraigSiemens, line 27: Rather than being flag to the generate command, I think it would be better if it was an option that could be set in the project spec. That's were all the other options for modifying the generated project are set.

MojtabaHs: I have added this to the spec options too. (check that out please). But the flag is for the CLI, because the render is very likely to override by the developer but should not be changed from the spec options.

MojtabaHs:
Any progress on this?

Sources/XcodeGenCLI/Commands/GenerateCommand.swift
MojtabaHs: @CraigSiemens Can you review it again? Thanks.

malekmatin (approved):

shadiGhelman (approved):

MojtabaHs (commented):
I've double-checked the PR and everything is ok.

Issue #763, opened 2020-01-29
Localized resources are not detected correctly


djbe:
I seem to be hitting the exact same issue as #80, namely that my `Localizable.strings(dict)` files (and `InfoPlist.strings` files) are not detected correctly and seen as loose files:

<img width="283" alt="Screenshot 2020-01-29 at 02 19 00" src="https://user-images.githubusercontent.com/641356/73319861-eb84ab00-423d-11ea-9132-2decc64ec534.png">

As you can see, strings files for storyboards are seen correctly, whereas normal strings files are not seen as language variations of the same file.

The directory structure is correct though:
```yaml
Application
  - Resources
    - Base.lproj
      - Authentication.storyboard
      - ... (only .storyboard files)
    - en.lproj
      - Authentication.strings
      - ... (other .strings files for storyboards)
      - InfoPlist.strings
      - Localizable.strings
      - Localizable.stringsdict
    - fr.lproj
      - Authentication.strings
      - ... (other .strings files for storyboards)
      - InfoPlist.strings
      - Localizable.strings
      - Localizable.stringsdict
```

My project file (mostly, the relevant bits):
```yaml
name: MultiSelector
targets:
  MultiSelector:
    type: application
    platform: iOS
    sources:
      - Application
    settings:
      base:
        INFOPLIST_FILE: Application/Resources/Info.plist
        PRODUCT_NAME: $(TARGET_NAME)
        SWIFT_VERSION: 5.0
```

Also tried using XcodeGen in another project, same issue, which is worse there 'cause that project has +-10 languages.

ikesyo:
From my understanding, according to the implementation of #70, those files must also be in `Base.lproj` directory.

djbe:
From Apple's own docs, they recommend to **not** add Localizable.strings (and such) to `Base.lproj`:
> Add the file to your Xcode project, as described in Adding Languages. Don’t add Localizable.strings to the Base localization when the dialog appears.

https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/MaintaingYourOwnStringsFiles/MaintaingYourOwnStringsFiles.html

So for interface builder files (storyboard, xib, nib) that are localised, the interface file should be in `Base.lproj` and the translation files (strings) in the corresponding language `.lproj`. For simple strings files such as `Localizable.strings`, `InfoPlist.strings` and others, they don't need to be in `Base.lproj`.

[...]

PR #1296, opened 2022-12-05
Fix logic related PBXVariantGroup

takeshi-1000:
Fixed logic in PBXVariantgGroup to meet the requirements of one my project.

- Even if there is no localize file in Base.lproj, recognize localize file as PBXVarientGroup, generate appropriate PBXFileReference, and generate PBXBuildFile for appropriate target.  https://github.com/yonaskolb/XcodeGen/issues/763
- To be able to add a PBXVariantGroup of one target A to another B target
    - Add the path of the file that is the source of the PBXVariantGroup you want to add to source. (Please refer to the fix for Tests/Fixtures/TestProject/project.yml in this PR)

Also Please let us know if you have any other suggestions.
Thanks.

### DIFF
--- a/Sources/ProjectSpec/SourceType.swift
+++ b/Sources/ProjectSpec/SourceType.swift
@@ -11,4 +11,5 @@ public enum SourceType: String {
     case group
     case file
     case folder
+    case variantGroup
 }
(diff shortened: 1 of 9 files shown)


takeshi-1000:
I will check ci problem.

PR #1303, opened 2022-12-24
Permit scheme build stage to be optional

itsthejb:
Hi,

Thanks for the excellent work on this project!

Here's a small tweak, I noticed: currently it's not possible to define schemes with an optional `Build` stage:

```
Parsing project spec failed: Decoding failed at "build": Nothing found
```

In fact, it's entirely legal to define a scheme without build. We're currently looking into this as the only way to allow running of tests in child modules, when those child modules are Swift Packages: it's not possible to create schemes to _build_ the modules, but it is possible to create scheme to _test_, and the modules will be implicitly built.

Screenshot 2022-12-24 at 12 16 03

Regardless of this, the spec should permit everything that's permitted by Xcode, so I made this initial rough PR.

Thanks very much


Sources/ProjectSpec/Scheme.swift
```diff
@@ -8,7 +8,7 @@ public typealias BuildType = XCScheme.BuildAction.Entry.BuildFor
 public struct Scheme: Equatable {
 
     public var name: String
-    public var build: Build
+    public var build: Build!
```
itsthejb, line 11: This should of course should be `Build?`, but changing to a full optional led to many type-inference errors in `SpecLoadingTests` that I wasn't able to easily fix

Tests/Fixtures/test_only.yml
```diff
@@ -0,0 +1,12 @@
...
+  Target:
+    type: application
+    platform: iOS
+    sources:
+      - Target
+schemes:
+  TestOnly:
+    test:
+      targets:
+      - Target
```
itsthejb, line 12: Use case: scheme has only a test action, no build

Issue #1315, opened 2023-01-21
Support cross-import-modules


Iron-Ham:
The [proposal around cross imports](https://gist.github.com/beccadax/8f35216340e829efa5e9c1fc090fda09) has been merged and the feature has been enabled in Xcode by default for some time. 

Let's support developers creating their own cross imports by allowing `.swiftcrossimport` directories and `.swiftinterface` files to move to the same directory which contains `modulemap` files. Although the project has shifted some time since this PR, I think [this](https://github.com/yonaskolb/XcodeGen/pull/346) could be used as a model of how to get this done, as it covers mapping `modulemap` files to the right directory.

Iron-Ham:
Getting some progress here:

1. I'm able to detect `.swiftoverlay` files and place the min the `.framework`. However, they should really be in the same directory as the `.modulemap` _and_ within a `$(PRODUCT_NAME).swiftcrossimport` directory.
2. `swiftinterface` files seem to be ignored completely.

PR #1317, opened 2023-01-21
Support for `swiftcrossimport` folders

Iron-Ham:
## Overview

closes https://github.com/yonaskolb/XcodeGen/issues/1315 

This PR implements support for `swiftcrossimport` by:
- Adding the `swiftcrossimport` folder-type as an extension on `Xcode` such that it's recognized as a wrapper. 
- Adding a special case for the `swiftcrossimport` within `framework` targets to map correctly.

See https://gist.github.com/beccadax/8f35216340e829efa5e9c1fc090fda09#declaring-a-cross-import-overlay for details

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Changed
 
-- Renamed build phase `Embed App Extensions` to `Embed Foundation Extensions` #1310 @casperriboe
+- Renamed build phase `Embed App Extensions` to `Embed Foundation Extensions` to fix Xcode 14 warning #1310 @casperriboe
 
 ### Fixed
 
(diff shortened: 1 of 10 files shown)


yonaskolb (commented):
Thanks @Iron-Ham. While I'm not familiar with cross import folders, this looks ok to me if it's working for you. 
It would be great if you could add an example of one of these files (even if it's empty) to the TestProject fixture https://github.com/yonaskolb/XcodeGen/blob/master/Tests/Fixtures/TestProject/project.yml

yonaskolb:
Great thanks, nearly there
- the diffs that are generated after running the tests need to be checked in
- tests are failing due to a small casing issue in `type: Framework`

yonaskolb (approved):

PR #1329, opened 2023-02-17
Remove SPM configurations warning from ProjectSpec

m-chojnacki:
ProjectSpec states that Swift Packages don't work in projects with configurations other than `Debug` and `Release`. It provides a link to Swift bugtracker to issue marked as duplicate. Original issue that it points to is closed and fixed since 2020 with Xcode 12.

Let's remove misleading note from ProjectSpec.

https://github.com/apple/swift-package-manager/issues/4674


yonaskolb (approved):
Great thanks @m-chojnacki!

PR #1336, opened 2023-03-16
Support for multiple deployment targets with xcode 14

### LINKED ISSUE #1333: Support for multiple deployment targets with Xcode 14?
Hi,
In XCode 14 there is a way to support different platforms with a single target.
In supported destinations you can specify iPhone, iPad, Mac and AppleTV.
Is there a way on the current version to do that or has this been planned for the future?
Thanks

amatig:
Hi, I am Giovanni Amati.
I am new, I have never contributed to this project but because in my company we are using XcodeGen, and we are starting a new app, it was good to have this new feature to support iOS and AppleTV in one single target.

This fixes #1333 , opened by me then I decided to try to solve it.

What I have done:

- Introduced an optional `supportedDestinations` array in target to create the right combination of build settings for multi platform destinations.
- Introduced a new enum `supportedDestination` (iOS, tvOS, macOS, macCatalyst and visionOS)
- Introduced `auto` has platform value that you can use only with supported destinations
- The definition of the platform for  supported destinations is optional, it fallbacks to auto if it is not defined.
- Introduced an optional `destinationFilters` array for sources and dependencies to filter things for platforms
- Introduced an optional `inferDestinationFiltersByPath` boolean, this is a convenience filter that helps you to filter sources (and resources) if their pathes match these patterns `**/<supportedDestination>/*` or `*_<supportedDestination>.swift`
- if `destinationFilters` and `inferDestinationFiltersByPath` are both defined, `inferDestinationFiltersByPath` is ignored
- if `supportedDestinations` is defined and `platform` is an array, an error is raised because we are using to ways to use multi platforms
- Added checks to improve the spec validation
- Added visionOS support
- Added unit tests
- Updated docs / changelog

Ex. of spec

```
targets:
  MyApp:
    type: application
    supportedDestinations: [iOS, tvOS]
    sources:
      - path: ../Project/GLEAP/Sources
        inferDestinationFiltersByPath: true
      - path: ../Project/GLEAP/Storyboards
        destinationFilters: [iOS]
    dependencies:
      - package: Yams
        product: Yams
        destinationFilters: [tvOS]
      - package: Lottie
        product: Lottie
```

Everything works with every bundle so it's possible to do the same for UI tests, 1 single target and if you have different file you can filter them by platform. It is very useful.

This spec becomes:

Screenshot 2023-03-20 at 13 48 39

Screenshot 2023-07-19 at 17 17 25

Many thanks!


amatig:
Hi, it's ready to review when you are free. Many thanks!

yonaskolb:
Thank you for your work here @amatig. Sorry I haven't had the time to go over all these changes yet, just know I haven't forgotten about this

amatig:
> Thank you for your work here @amatig. Sorry I haven't had the time to go over all these changes yet, just know I haven't forgotten about this

Hi, any news?

Sources/ProjectSpec/Target.swift
```diff
@@ -159,13 +162,17 @@ extension Target {
...
-
+        
         var crossPlatformTargets: [String: JSONDictionary] = [:]
 
         for (targetName, target) in targetsDictionary {
-
+            var target = target
+            
+            if let platforms = target["platform"] as? [String], target["supportedPlatforms"] != nil {
+                target["platform"] = platforms.first
```
yonaskolb, line 172: This may be surprising. Could you leave a comment on the intention here?

Sources/ProjectSpec/SupportedPlatforms.swift
```diff
@@ -0,0 +1,37 @@
+import Foundation
+
+public enum SupportedPlatforms: String, CaseIterable {
```
yonaskolb, line 3: Could we rename this enum to the singular case `SupportedPlatform`

Docs/ProjectSpec.md
```diff
@@ -465,6 +467,33 @@ targets:
...
+- `macOS`
+- `macCatalyst`
+
+Note that the definition of supported platforms will skip the `Multi Platform targets` generation.
+
+```yaml
+targets:
+  MyFramework:
+    sources: MyFramework
+    platform: iOS
```
yonaskolb, line 485: Would we even need to specify `platform` in this case? Having them both there is a little confusing

yonaskolb (commented):
Sorry for the delay on this @amatig.
This is a really fantastic PR, thank you for all your hard work here! I really appreciate all the tests. I've left some small comments above but mostly looks good.

I would say too there would probably still be a little bit of confusion for users about when to use `platform` or `supportedPlatforms`.

One thing that would be great to add is an example multi platform app in the TestProject, where you've added the platform files. That project is designed to have every xcodegen feature in it. It doesn't need to do anything, just to build

Docs/ProjectSpec.md
amatig: Hi, I have created a default new project with Xcode 14 and it seems that you need a SDKROOT, a base SDK to use.

SupportedPlatform from the other end helps you to support multiple platform in the same target mixing together target families and other build settings.

Sources/ProjectSpec/Target.swift
amatig: Basically if you are using `supportedPlatforms` and you want to support multiple platforms in the same target we can not allow xcodegen to generate multiple target for each platform if you use `platform` as an array.

So if you by mistake are adding both xcodegen ignores the array in `platform` taking as base platform the first one and it applies the new rules to support multiple platform in the same target.

SettingPresets/SupportedPlatforms/tvOS.yml
```diff
@@ -0,0 +1,2 @@
+SUPPORTED_PLATFORMS: appletvos appletvsimulator
+TARGETED_DEVICE_FAMILY: '3'
```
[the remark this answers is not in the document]
amatig: I have created a fresh new project with Xcode 14 and I looked at all the build settings.

By the current presets it was not possible to achieve the goal, these new presets have parameter that Xcode must have if you start to define multiple platform in the same target.

This is the minimum that I found that mixed together recreates exactly what Xcode does with the build settings.

SettingPresets/SupportedPlatforms/iOS.yml
```diff
@@ -0,0 +1,4 @@
+SUPPORTED_PLATFORMS: iphoneos iphonesimulator
+TARGETED_DEVICE_FAMILY: '1,2'
+SUPPORTS_MACCATALYST: NO
+SUPPORTS_MAC_DESIGNED_FOR_IPHONE_IPAD: YES
```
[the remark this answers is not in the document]
amatig: I saw that Xcode 14 does this by default if you create an iOS app, by default it can run on M1+ in a window that follow the iPad design.

If you chose in `supportedPlatforms` that you want to use macOS or Catalysts this will be disable because it is not possible to support multiple Mac platform.

amatig:
> Sorry for the delay on this @amatig. This is a really fantastic PR, thank you for all your hard work here! I really appreciate all the tests. I've left some small comments above but mostly looks good.
> 
> I would say too there would probably still be a little bit of confusion for users about when to use `platform` or `supportedPlatforms`.
> 
> One thing that would be great to add is an example multi platform app in the TestProject, where you've added the platform files. That project is designed to have every xcodegen feature in it. It doesn't need to do anything, just to build



> Sorry for the delay on this @amatig. This is a really fantastic PR, thank you for all your hard work here! I really appreciate all the tests. I've left some small comments above but mostly looks good.
> 
> I would say too there would probably still be a little bit of confusion for users about when to use `platform` or `supportedPlatforms`.
> 
> One thing that would be great to add is an example multi platform app in the TestProject, where you've added the platform files. That project is designed to have every xcodegen feature in it. It doesn't need to do anything, just to build

Yes I know that I could be confusing but it seems that we need a base platform (the SDK one to use) and then multiple platform that we want to support. The Xcode 14 settings work in this way.

amatig:
Addressed few comments and added TestProject

Sources/ProjectSpec/Target.swift
amatig: I think that people, now, can not use anymore version less than Xcode 14 to upload builds.

So one possibility could renaming `platform` in `basePlatform` or `sdkPlatform` (or something else) and completely remove the possibility to manage it as an array and generate different target per platform.

yonaskolb: Maybe throwing an error is more appropriate? Might make it clearer.  Saying something like you can't have an array of platforms if you define supportedPlatforms

amatig: Yes, I will do that

Tests/Fixtures/TestProject/App_supportedPlatforms/project.yml
```diff
@@ -0,0 +1,32 @@
+name: MyProject
```
yonaskolb, line 1: Thanks for adding this, but can we integrate with `Tests/Fixtures/TestProject/project.yml` as that's the project that gets built on CI to check the generated project is valid. You can even use an `include` in that spec to copy this stuff in

amatig: ah ok sure

Sources/ProjectSpec/Target.swift
```diff
@@ -158,18 +158,16 @@ extension Target: PathContainer {
 
 extension Target {
 
-    static func resolveMultiplatformTargets(jsonDictionary: JSONDictionary) -> JSONDictionary {
+    static func resolveMultiplatformTargets(jsonDictionary: JSONDictionary) throws -> JSONDictionary {
```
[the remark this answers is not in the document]
amatig: ok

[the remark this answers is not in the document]
amatig: Hi, I have a question about this. In `SpecValidation` we already have a `Project` instance that has a property platform that is not an array. It is a simple string because `SpecValidation` phase I think arrives after `resolveMultiplatformTargets` so we convert that in multiple target.
So in the `SpecValidation` I don't know if platform field was an array. Is there a way to understand in `SpecValidation` if we have done a resolveMultiplatformTargets?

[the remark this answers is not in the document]
amatig: Maybe something like this

```
// validate project dictionary
do {
    try specLoader.validateProjectDictionaryWarnings()
} catch {
    warning("\(error)")
}
```

this is the `execute` before to create the project and run, them, the SpecValidation. I could do something like this but that raises errors. Because the problem is that I need the original json to see if platform was an array.

If I do something in SpecValidation we have already parsed the json into a Project and if platform was an array is has been resolved in multiple targets.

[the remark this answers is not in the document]
amatig: Or I can store in the Target a property `generatedMultiTargets: Bool` just to understand if we went in `resolveMultiplatformTargets` and to the error check in `SpecValidation`.

Tests/Fixtures/TestProject/App_supportedPlatforms/project.yml
amatig: Done

bdrelling:
@amatig just wanted to say thanks for your hard work on this, looking forward to having it integrated. 

I am specifically hoping to avoid the problem that occurs currently when frameworks generated with XcodeGen using Xcode 14+ seem to automatically include all supported destinations, even when not specified, like so:

image

Sources/ProjectSpec/Target.swift
```diff
@@ -57,7 +58,8 @@ public struct Target: ProjectTarget {
     public var productName: String
     public var onlyCopyFilesOnInstall: Bool
     public var putResourcesBeforeSourcesBuildPhase: Bool
-
+    public var isResolved: Bool
```
yonaskolb, line 61: Could you add a code comment here for future users. The name isn't very specific

Sources/ProjectSpec/Target.swift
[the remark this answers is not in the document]
yonaskolb: Looks good, thanks

yonaskolb (commented):
Thanks @amatig, looking good, we're getting there!

A couple of comments:
1. @bdrelling makes an interesting point that in the Xcode interface cross platform apps have multiple "Supported Destinations" and use "Destination" terminology instead of "Platform". Renaming `supportedPlatforms` to `supportedDestinations` would match this and remove some of the ambiguity. What do you think of this?
2. On the point of if we still need `platform` if we define `supportedPlatforms`, you mentioned we still need a base SDK. I just created a new cross platform app in Xcode 14.3 and the base SDK is set to `auto`. Is that not something we could replicate?
3. Master now has support for `visionOS`. Ideally this work would support that. How do you feel about adding support for a visionOS destination? It could be a followup PR, though you might encounter some build errors if you update from master now anyway.

amatig:
> Thanks @amatig, looking good, we're getting there!
> 
> A couple of comments:
> 
> 1. @bdrelling makes an interesting point that in the Xcode interface cross platform apps have multiple "Supported Destinations" and use "Destination" terminology instead of "Platform". Renaming `supportedPlatforms` to `supportedDestinations` would match this and remove some of the ambiguity. What do you think of this?
> 
> Screenshot 2023-08-08 at 8 47 18 pm
> 2. On the point of if we still need `platform` if we define `supportedPlatforms`, you mentioned we still need a base SDK. I just created a new cross platform app in Xcode 14.3 and the base SDK is set to `auto`. Is that not something we could replicate?
> 
> Screenshot 2023-08-08 at 8 37 38 pm
> 3. Master now has support for `visionOS`. Ideally this work would support that. How do you feel about adding support for a visionOS destination? It could be a followup PR, though you might encounter some build errors if you update from master now anyway.

Oh yes I can rename it. I keep that name because supportedPlatforms is the one used by XcodeProj. I can look at the auto platform and the other comments. About visionOS we still on Xcode beta but I can have a look.
I am in holidays at the moment but I will be back soon.

amatig:
Hi, I did a couple of things and now I am investigating about "auto" for baseSDK and visionOS.

About "auto" I have a couple of things to ask. 
At the moment, a lot of logics are based on the platform that is mandatory. If I add a new case auto in the enum we are going to lose presets that we are applying now (Platform and Product_Platform) so the result will be a project a little bite more empty where the user needs to configure more build settings stuff in the spec because we can not help the generation.

Is this an issue? Or isn't it better to leave to he user the decision to define one "baseSDK" so we can give him few presets and then he needs to customised the extra platforms things if he wants a multi-platform solution?

Or both? You can put what ever you want iOS, tvOS or auto (or nothing that fallbacks to auto). The user will lose presets only in auto.

What do you think?

amatig:
> Hi, I did a couple of things and now I am investigating about "auto" for baseSDK and visionOS.
> 
> About "auto" I have a couple of things to ask. At the moment, a lot of logics are based on the platform that is mandatory. If I add a new case auto in the enum we are going to lose presets that we are applying now (Platform and Product_Platform) so the result will be a project a little bite more empty where the user needs to configure more build settings stuff in the spec because we can not help the generation.
> 
> Is this an issue? Or isn't it better to leave to he user the decision to define one "baseSDK" so we can give him few presets and then he needs to customised the extra platforms things if he wants a multi-platform solution?
> 
> Or both? You can put what ever you want iOS, tvOS or auto (or nothing that fallbacks to auto). The user will lose presets only in auto.
> 
> What do you think? @yonaskolb

I have an idea. Because the enum of SupportedDestination has a priority of destination I can assign to the target the platform preset and product_preset of the destination with higher priority. So the user can have something if the platform is "auto" or undefined.

Sources/XcodeGenKit/SourceGenerator.swift
```diff
@@ -109,7 +109,23 @@ class SourceGenerator {
...
     }
-
+    
+    private func makePlatformFilters(for path: Path, with filters: [SupportedDestination]?, or inferPlatformFiltersByPath: Bool?) -> [String]? {
+        if let filters = filters, !filters.isEmpty {
+            return filters.map { $0.string }
+        } else if inferPlatformFiltersByPath == true {
+            for supportedDestination in SupportedDestination.allCases {
+                let regex1 = try? NSRegularExpression(pattern: "\\/\(supportedDestination)\\/", options: .caseInsensitive)
+                let regex2 = try? NSRegularExpression(pattern: "\\_\(supportedDestination)\\.swift$", options: .caseInsensitive)
```
yonaskolb, line 119: I'm slightly worried about the performance implications of this, but it's opt in

yonaskolb (commented):
Nice work @amatig

Docs/ProjectSpec.md
```diff
@@ -560,10 +597,11 @@ A dependency can be one of a 6 types:
...
-- [ ] **platformFilter**: **String** - This field is specific to Mac Catalyst. It corresponds to the "Platforms" dropdown in the Frameworks & Libraries section of Target settings in Xcode. Available options are: **iOS**, **macOS** and **all**. Defaults is **all**
+- [ ] **codeSign**: **Bool** - Whether the `codeSignOnCopy` setting is applied when embedding framework. Defaults to true.
+- [ ] **removeHeaders**: **Bool** - Whether the `removeHeadersOnCopy` setting is applied when embedding the framework. Defaults to true.
+- [ ] **weak**: **Bool** - Whether the `Weak` setting is applied when linking the framework. Defaults to false.
+- [ ] **platformFilter**: **String** - This field is specific to Mac Catalyst. It corresponds to the "Platforms" dropdown in the Frameworks & Libraries section of Target settings in Xcode. Available options are: **iOS**, **macOS** and **all**. Defaults is **all**.
+- [ ] **platformFilters**: **[[Supported Destinations](#supported-destinations)]** - List of supported platform destinations this dependency should filter to. Defaults to all supported destinations.
```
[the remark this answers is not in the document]
amatig: ah yes, I can do this. This is ok, as input I can change it but when the value arrives to `XcodeProj` we must use this field name because that project accept that.

Sources/ProjectSpec/Target.swift
```diff
@@ -162,15 +165,17 @@ extension Target {
...
         for (targetName, target) in targetsDictionary {
-
             if let platforms = target["platform"] as? [String] {
-
                 for platform in platforms {
                     var platformTarget = target
+                    
+                    /// This value is set to help us to check, in Target init, that there are no conflicts in the definition of the platforms. We want to ensure that the user didn't define, at the same time,
+                    /// the new Xcode 14 supported destinations and the XcodeGen generation of Multiple Platform Targets (when you define the platform field as an array).
+                    platformTarget["isResolved"] = true
```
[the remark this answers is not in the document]
amatig: Ok

Docs/ProjectSpec.md
```diff
@@ -483,7 +515,9 @@ A source can be provided via a string (the path) or an object of the form:
...
 - [ ] **includes**: **[String]** - A list of global patterns in the same format as `excludes` representing the files to include. These rules are relative to `path` and _not the directory where `project.yml` resides_. If **excludes** is present and file conflicts with **includes**, **excludes** will override the **includes** behavior.
-- [ ] **createIntermediateGroups**: **Bool** - This overrides the value in [Options](#options)
+- [ ] **platformFilters**: **[[Supported Destinations](#supported-destinations)]** - List of supported platform destinations the files should filter to. Defaults to all supported destinations.
+- [ ] **inferPlatformFiltersByPath**: **Bool** - This is a convenience filter that helps you to filter the files if their paths match these patterns `**/<supportedDestination>/*` or `*_<supportedDestination>.swift`. Note, if you use `platformFilters` this flag will be ignored.
```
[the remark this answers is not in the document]
amatig: Yes, ok

Sources/XcodeGenKit/SourceGenerator.swift
amatig: I know but this thing is very helpful

Sources/ProjectSpec/Target.swift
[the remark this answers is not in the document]
amatig: Done

Docs/ProjectSpec.md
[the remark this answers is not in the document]
amatig: Done

Docs/ProjectSpec.md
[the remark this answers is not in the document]
amatig: Done

amatig:
Any news?

FelixLisczyk:
This feature would significantly reduce the number of framework targets in my project. 👍 When can we merge it?

amatig:
> This feature would significantly reduce the number of framework targets in my project. 👍 When can we merge it?

Hope soon, I made the last changes after the review I am waiting for @yonaskolb

yonaskolb (approved):
Thank you for your patience @amatig, and all your great work here. I think this is good to merge now 👍

FelixLisczyk:
@amatig I've noticed that the `supportedDestination` enum doesn't support the `watchOS` platform. Is there any reason for this? I can add it manually in Xcode.

Screenshot 2023-11-09 at 11 51 01

Issue #1283, opened 2022-11-04
bugs in "includes" in "sources"


WeijunDeng:
<img width="147" alt="image" src="https://user-images.githubusercontent.com/7430966/199927196-c4a2d047-1cde-49ed-917c-11f9bbb4cb2b.png">
<img width="330" alt="image" src="https://user-images.githubusercontent.com/7430966/199927281-c2b7bfb9-1e31-411b-9746-3636bb2040b1.png">
<img width="180" alt="image" src="https://user-images.githubusercontent.com/7430966/199927334-251e798e-1c60-48b0-b256-365dc9545d98.png">

There is no .swift file in this folder.
Expected: DO NOT add this .h file to project.

PR #1337, opened 2023-03-19
Fix issue with includes not working when no matches are found

shnhrrsn:
This PR addresses the issue raised in #1283, where a single includes glob pattern with no matches causes the `isIncludedPath` function to ignore the includes restriction.

The problem occurs when the `isIncludedPath` function checks if `includePaths` is empty, which is true if there are no include patterns or if there are no matches for a pattern. To fix this issue, I changed `includePaths` to an optional, using `nil` to indicate that all paths should be allowed.

- Update the `isIncludedPath` function to handle optional `includePaths`.
- Modify the `getSourceChildren` function to accept an optional `includePaths` parameter.
- Adjust the `getSourceFiles` function to work with optional `includePaths`.
- Update the generate function to handle an optional `includePaths`.

Additionally, I added a new test case to verify the correct behavior when no matches are found for an includes pattern.

### DIFF
--- a/Sources/XcodeGenKit/SourceGenerator.swift
+++ b/Sources/XcodeGenKit/SourceGenerator.swift
@@ -384,12 +384,12 @@ class SourceGenerator {
     }
 
     /// Checks whether the path is not in any default or TargetSource excludes
-    func isIncludedPath(_ path: Path, excludePaths: Set<Path>, includePaths: SortedArray<Path>) -> Bool {
+    func isIncludedPath(_ path: Path, excludePaths: Set<Path>, includePaths: SortedArray<Path>?) -> Bool {
         return !defaultExcludedFiles.contains(where: { path.lastComponent == $0 })
             && !(path.extension.map(defaultExcludedExtensions.contains) ?? false)
             && !excludePaths.contains(path)
             // If includes is empty, it's included. If it's not empty, the path either needs to match exactly, or it needs to be a direct parent of an included path.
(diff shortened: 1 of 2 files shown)


Sources/XcodeGenKit/SourceGenerator.swift
```diff
@@ -384,12 +384,12 @@ class SourceGenerator {
     }
 
     /// Checks whether the path is not in any default or TargetSource excludes
-    func isIncludedPath(_ path: Path, excludePaths: Set<Path>, includePaths: SortedArray<Path>) -> Bool {
+    func isIncludedPath(_ path: Path, excludePaths: Set<Path>, includePaths: SortedArray<Path>?) -> Bool {
         return !defaultExcludedFiles.contains(where: { path.lastComponent == $0 })
             && !(path.extension.map(defaultExcludedExtensions.contains) ?? false)
             && !excludePaths.contains(path)
             // If includes is empty, it's included. If it's not empty, the path either needs to match exactly, or it needs to be a direct parent of an included path.
```
yonaskolb, line 391: Is this comment still accurate, or does it need changing?

yonaskolb (commented):
Thanks for investigating and fixing this @shnhrrsn. Could you also please add a changelog entry

yonaskolb (approved):

yonaskolb:
Thanks @shnhrrsn!

Issue #1131, opened 2021-09-27
Duplicate elements of type 'PBXFileElement' were found in a Set


dalemyers:
We are seeing this error regularly:

```
⚙️  Generating plists...
⚙️  Generating project...
Fatal error: Duplicate elements of type 'PBXFileElement' were found in a Set.
This usually means either that the type violates Hashable's requirements, or
```

Our project is quite large, but we have no insight into what is going on here. Re-running xcodegen solves this, so it's likely a race condition somewhere. I've got a PR which adds some debug logging to hopefully help try and track it down: https://github.com/yonaskolb/XcodeGen/pull/1128

guojiubo:
Any updates on this?

vguerci:
Gave a look at this yesterday, as we were having the same issue breaking our CI randomly.

Basically, `XcodeProj`'s `PBXObject` and inherited classes are not respecting `Hashable` requirements. Two different instances (not equal) can have the same hash, which leads randomly to this issue.

Tried to fix that naively, to realise it was more complex, so closed my Fix above.

dungi:
Does someone find a fix for this yet?

dalemyers:
Thankfully my new project reproduced the issue much more frequently and I was able to create a fix: https://github.com/yonaskolb/XcodeGen/pull/1341

dalemyers-flipdish:
Note: There may be multiple reasons for this error, but my fix corrects the one I'm seeing now at least. Try it and see if it fixes your own.

vguerci:
Yes, it should work, this is how I fixed the issues on our side, avoiding the collision, by removing some trailing `/`.

PR #1341, opened 2023-03-23
Sanitize the ends of folder source paths

### LINKED ISSUE #1339: Duplicate group generated inconsistently between runs
I have a project which contains 3 targets. The 3 targets all have the same source folder (this will be fixed in the future, but it's what I've got for now). When xcodegen generates the project, _usually_ it deduplicates those groups correctly and I just see one in Xcode with each file being a member of all 3 targets. Sometimes however, it generates two of those groups. In the pbxproj they contain identical contents, but Xcode only loads one of them. With `createIntermediateGroups: true` this happened every time. With `createIntermediateGroups: false` it happens maybe 80% of the time. 

When checking a diff of the pbxproj between the "good" and "bad" versions, I also noticed a group with the same content but with a `TEMP_` prefix. This one however was not added to the root group. It exists without a parent.

dalemyers-flipdish:
Setting the source: `/foo/bar` is _sometimes_ different from `/foo/bar/` even if `bar` is a folder in both cases. The result of this is that we often run into a race condition where we have two objects with the same hash but different properties. This fixes #1339 and #1131.

### DIFF
--- a/Sources/ProjectSpec/TargetSource.swift
+++ b/Sources/ProjectSpec/TargetSource.swift
@@ -53,7 +53,7 @@ public struct TargetSource: Equatable {
         attributes: [String] = [],
         resourceTags: [String] = []
     ) {
-        self.path = path
+        self.path = (path as NSString).standardizingPath
         self.name = name
         self.group = group
         self.compilerFlags = compilerFlags
@@ -87,7 +87,7 @@ extension TargetSource: ExpressibleByStringLiteral {
 extension TargetSource: JSONObjectConvertible {
 
     public init(jsonDictionary: JSONDictionary) throws {
-        path = try jsonDictionary.json(atKeyPath: "path")
+        path = ((try jsonDictionary.json(atKeyPath: "path")) as NSString).standardizingPath
         name = jsonDictionary.json(atKeyPath: "name")
         group = jsonDictionary.json(atKeyPath: "group")
 
--- a/Sources/ProjectSpec/TargetSource.swift
+++ b/Sources/ProjectSpec/TargetSource.swift
@@ -4,8 +4,18 @@ import PathKit
 
 public struct TargetSource: Equatable {
     public static let optionalDefault = false
+    
+    private var _path: String = ""
 
-    public var path: String
+    public var path: String {
+        get {
+            return _path
+        }
+        set {
+            _path = TargetSource.clean(path: newValue)
+        }
+    }
+    
     public var name: String?
     public var group: String?
     public var compilerFlags: [String]
@@ -48,7 +58,7 @@ public struct TargetSource: Equatable {
         attributes: [String] = [],
         resourceTags: [String] = []
     ) {
-        self.path = path
+        self._path = TargetSource.clean(path: path)
         self.name = name
         self.group = group
         self.compilerFlags = compilerFlags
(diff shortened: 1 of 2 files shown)


Brian-MobileTeam (approved):

Sources/ProjectSpec/TargetSource.swift
```diff
@@ -4,8 +4,18 @@ import PathKit
...
+    
+    private var _path: String = ""
 
-    public var path: String
+    public var path: String {
+        get {
+            return _path
+        }
+        set {
+            _path = TargetSource.clean(path: newValue)
```
vguerci, line 15: What about sanitizing using `didSet`? That would avoid using a backing `_path` variable.

Sources/ProjectSpec/TargetSource.swift
```diff
@@ -62,6 +72,14 @@ public struct TargetSource: Equatable {
         self.attributes = attributes
         self.resourceTags = resourceTags
     }
+    
+    private static func clean(path: String) -> String {
```
vguerci, line 76: Could [standardizingPath](https://developer.apple.com/documentation/foundation/nsstring/1407194-standardizingpath) be a better option than this method?

Sources/ProjectSpec/TargetSource.swift
```diff
@@ -82,7 +100,7 @@ extension TargetSource: ExpressibleByStringLiteral {
 extension TargetSource: JSONObjectConvertible {
 
     public init(jsonDictionary: JSONDictionary) throws {
-        path = try jsonDictionary.json(atKeyPath: "path")
+        _path = TargetSource.clean(path: try jsonDictionary.json(atKeyPath: "path"))
```
vguerci, line 103: Using `didSet` could avoid this line ;)

vguerci (commented):
Nice work! Dropped some comments

Sources/ProjectSpec/TargetSource.swift
dalemyers: Honest answer? I tried and Xcode shouted at me for doing it wrong. It's been about a year since I last wrote any Swift code so I assumed I'd just mis-remembered :D  

I'll take another look.

Sources/ProjectSpec/TargetSource.swift
```diff
@@ -4,8 +4,13 @@ import PathKit
 
 public struct TargetSource: Equatable {
     public static let optionalDefault = false
-
-    public var path: String
+    
+    public var path: String {
+        didSet {
+            path = (path as NSString).standardizingPath
```
yonaskolb, line 10: As far as I remember didSet is not called from the initializer, so you'd need to do this transformation in the `public init(path: String ...)` and `init(jsonDictionary: JSONDictionary)`

yonaskolb (commented):
Thanks @dalemyers-flipdish. If you had the time to add a (previously) failing unit test to https://github.com/yonaskolb/XcodeGen/blob/master/Tests/XcodeGenKitTests/SourceGeneratorTests.swift, that would be fantastic

Sources/ProjectSpec/TargetSource.swift
dalemyers: Ah interesting. I'm sure you saw my comment above about the last time I wrote Swift code. Interestingly, I tested this way and I couldn't repeat it. I wonder if something else is setting this property and that's what's triggering the bug? 

Either way, I'll add to the constructors.

dalemyers-flipdish:
I've added the sanitization to the constructors. Regarding the tests, I'm not really sure what the options are here? I couldn't reproduce this every time no matter what I tried. Personally, I'd rather skip it than add a knowingly flaky test.

yonaskolb (approved):

PR #1353, opened 2023-04-27
Display absolute path for invalidBuildScriptPath and invalidTargetConfigFile error print.

BartoszKosiorek-TomTom:
With previous implementation, the error message was not clear for end user:
`  Spec validation error: Target "Project" has a script "SwiftLint" which has a path that doesn't exist "../../wrong-dir/scripts/run_swiftlint_in_xcode_wrapper.sh"`

With new implementation with absolute path, it is easier to validate what is wrong with path:
`  Spec validation error: Target "Project" has a script "SwiftLint" which has a path that doesn't exist "/Users/SomeUser/dev/fancy-project/wrong-dir/scripts/run_swiftlint_in_xcode_wrapper.sh"`

### DIFF
--- a/Sources/ProjectSpec/SpecValidation.swift
+++ b/Sources/ProjectSpec/SpecValidation.swift
@@ -81,8 +81,9 @@ extension Project {
         for target in projectTargets {
 
             for (config, configFile) in target.configFiles {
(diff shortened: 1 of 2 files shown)


yonaskolb (approved):
Great, thanks @BartoszKosiorek-TomTom!

PR #1357, opened 2023-05-09
add json-schema

Kila2:
add json schema files.


yonaskolb:
Wow looks like a lot of work went into these. How did you create them? 
What is your intention of adding them here? They would serve as a good source of truth for documentation, but at the moment they just duplicate information.
I'm worried about the maintenance work these would create if these weren't generated, or these weren't being used to generate documentation.

Kila2:
> Wow looks like a lot of work went into these. How did you create them? What is your intention of adding them here? They would serve as a good source of truth for documentation, but at the moment they just duplicate information. I'm worried about the maintenance work these would create if these weren't generated, or these weren't being used to generate documentation.

- Except for xcode_settings.json, all the files are handwritten.
- I wanted to have the ability to provide completion, suggestions, and automatic validation when viewing or writing project.yml when using VSCode with the Yaml plugin. Therefore, these files were created based on projectSpec.md. 
- Perhaps the documentation can be regenerated based on this file. Currently, the files cover most cases, but there are still some small parts that have not been added to the json-schema, such as the Test Target Object.

The current completion level is sufficient for my needs. I have submitted a PR to provide additional features.

imWildCat:
> I'm worried about the maintenance work these would create if these weren't generated, or these weren't being used to generate documentation.

That's a valid concern. But currently the configuration structs are usually not `Codable`, which makes it not feasible to generate the JSON schema in the short term.


As we don't usually introduce breaking changes in the format of `project.yml`, it is not a bad idea to merge handwritten schema to this repo.

imWildCat (approved):

PR #1366, opened 2023-06-21
Add ability to sort local Packages group

mthole:
In my team's project, we primarily use the local SPM packages organized into "layers", with a directional dependency graph. In such a set up, it's super helpful to have the ordering of the layers in the Xcode organizer reflect the logical ordering of the layers.  This PR adds that functionality, expanding the `groupOrdering` configuration to be applied to the local package ordering as well.

### DIFF
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -586,21 +586,40 @@ public class PBXProjGenerator {
 
     func sortGroups(group: PBXGroup) {
         // sort children
-        let children = group.children
-            .sorted { child1, child2 in
-                let sortOrder1 = child1.getSortOrder(groupSortPosition: project.options.groupSortPosition)
-                let sortOrder2 = child2.getSortOrder(groupSortPosition: project.options.groupSortPosition)
+        let children: [PBXFileElement]
+        if group.name == "Packages",
+           let packagesOrdering = project.options.groupOrdering.first(where: { $0.pattern == "Packages" }) {
+            
+            var sortedPackages = [PBXFileElement]()
+            var unsortedPackages = Array(group.children)
+            for packageName in packagesOrdering.order {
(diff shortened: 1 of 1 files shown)


Sources/XcodeGenKit/PBXProjGenerator.swift
```diff
@@ -586,21 +586,40 @@ public class PBXProjGenerator {
 
     func sortGroups(group: PBXGroup) {
         // sort children
-        let children = group.children
-            .sorted { child1, child2 in
-                let sortOrder1 = child1.getSortOrder(groupSortPosition: project.options.groupSortPosition)
-                let sortOrder2 = child2.getSortOrder(groupSortPosition: project.options.groupSortPosition)
+        let children: [PBXFileElement]
+        if group.name == "Packages",
+           let packagesOrdering = project.options.groupOrdering.first(where: { $0.pattern == "Packages" }) {
```
mthole, line 591: This is a bit fragile-- other suggestions welcome!

jflan-dd: Another option could be to treat any folder as "group" for the purposes of sorting in [`setupGroupOrdering`](https://github.com/yonaskolb/XcodeGen/blob/372f20fe5a3823dc398f52cb6cc1af5003d452f7/Sources/XcodeGenKit/PBXProjGenerator.swift#L611-L656) with something like this:
```swift
    public func setupGroupOrdering(group: PBXGroup) {
        let groupOrdering = project.options.groupOrdering.first { groupOrdering in
            // existing logic ...
        }

        func isGroupOrFolder(_ element: PBXFileElement) -> Bool {
            element is PBXGroup || (element as? PBXFileReference)?.lastKnownFileType == "folder"
        }
        
        if let order = groupOrdering?.order {
            let files = group.children.filter { !isGroupOrFolder($0) }
            var groups = group.children.filter { isGroupOrFolder($0) }
            
            // existing logic ...
```

Issue #1362, opened 2023-05-29
`generate` command changes `objectVersion` of `.pbxproj` to an older one


KacperCzapp:
When running `xcodegen generate`, in `.pbxproj` file the `objectVersion` keeps being reverted to 51 (from 54 which newest Xcode sets by itself, for example when adding new files). How to fix this?

Environment:
- XcodeGen 2.35.0
- Xcode 14.3 (14E222b)

yonaskolb:
fixed by https://github.com/yonaskolb/XcodeGen/pull/1368

PR #1368, opened 2023-06-26
Updates project object version to support `BuildIndependentTargetsInParallel` setting

leonardorock:
Hi, I've been trying to implement the XcodeGen to a project and the generated project would always end up with a warning about the new "Parallelize build for command-line builds", after some investigation I've discovered that the attribute key that sets this value is `BuildIndependentTargetsInParallel`, and without updating the `object version` of the `.pbxproj` file this setting is completely ignored by Xcode.

This also would fix a open issue #1362 

## Discussion

I believe that this setting should be default from now on since that's the default setting when creating a new App from Xcode.
This PR doesn't set this value as default but allows it to be overwritten and interpreted by Xcode.

## Setting "Parallelize build for command-line builds"

To enable the Parallelize build for command-line builds configuration add this to the `project.yml` file
```yaml
attributes:
  BuildIndependentTargetsInParallel: YES
```

The **_project_** configuration section should have the "Parallelize build for command-line builds" checked ✅ 

Screenshot 2023-06-26 at 4 52 35 PM

## Environment
- XcodeGen Version: 2.35.0
- Xcode Version: 14.3.1


leonardorock:
Thank you @yonaskolb, the `CHANGELOG.md` has been updated as requested.

yonaskolb (approved):

PR #1376, opened 2023-07-24
Update README for Xcode version

chessbyte:
Showing Xcode 11 in the README in 2023 does not instill confidence in the project.

Xcode 11 was released in 2019 and the latest dot version updated in 2020.  I am assuming that XcodeGen supports Xcode 14.x

Additionally, as you do for Swift Compatibility, it would be helpful to the community if the project shows what Xcode versions it works with.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ A project would be created with 2 connected targets, with all the required confi
 
 ## Installing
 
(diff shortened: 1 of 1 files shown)


yonaskolb (approved):
A lot of maintenance relies on contributors such as yourself @chessbyte, so thank you

PR #1377, opened 2023-07-26
Don't add `PBXContainerItemProxy` when reusing `PBXReferenceProxy`

liamnichols:
Fixes https://github.com/yonaskolb/XcodeGen/pull/1354#issuecomment-1650099512

In an unrelated test, I spotted that the recent changes in #1354 were mistakenly creating orphaned `PBXContainerItemProxy` objects in the project file. This change updates the fixture tests to reproduce the issue and applies a patch (check out the individual commits to see the impact).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,10 +5,11 @@
 ### Added
 
 - Added `scheme.enableGPUValidationMode` #1294 @LouisLWang
-  
+
 ### Fixed
 
 - Fix external dependencies from being removed by Xcode #1354 @OdNairy
+- Stop creating orphaned object references when reusing references to external dependencies #1377 @liamnichols
 
 ## 2.35.0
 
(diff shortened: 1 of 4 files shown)


liamnichols:
On a side note, I became aware of this because we use R.swift in our project, which uses XcodeEdit, which has a validator to check for orphaned object references in the project file:

https://github.com/tomlokhorst/XcodeEdit/blob/cd466d6e8c5ffd2f2b61165d37b0646f09068e1e/Sources/XcodeEdit/AllObjects.swift#L142-L176

I wonder if it would be beneficial/possible for XcodeGen to perform a similar validation 🤔 Either just in the unit tests, or maybe even at runtime

yonaskolb (commented):
Nice one, thank you @liamnichols! 🙏  Could you please add a CHANGELOG fix entry as well

yonaskolb (approved):

PR #1374, opened 2023-07-16
Add support for adding build tool plugins to targets

BarredEwe:
# Short description 📝
Support for [**Build Tool Plugins**](https://github.com/apple/swift-package-manager/blob/main/Documentation/Plugins.md#build-tool-plugins) has been added here. A detailed discussion can be read here: https://github.com/yonaskolb/XcodeGen/issues/1290

Screenshot 2023-07-16 at 13 18 29

# Details ⚙️
### API

To use plugins, you need to specify in your target which plugin you want to connect.

```yaml
targets:
  App:
    buildToolPlugins:
      - plugin: PrefirePlaybookPlugin
        package: Prefire
```

 And don't forget to connect the package.

```yaml
packages:
  Prefire:
    url: https://github.com/BarredEwe/Prefire
    from: 1.3.0
```

### What is generated

1. Added `PBXTargetDependency`:
```xml
E157C6348B8AD6A28C706801 /* PBXTargetDependency */ = {
	isa = PBXTargetDependency;
	productRef = DC47EF1BFBBD751E3C1C95E3 /* PrefirePlaybookPlugin */;
};
```
2. Added Plugin information. An additional prefix (`plugin:`) is added:
```xml
DC47EF1BFBBD751E3C1C95E3 /* PrefirePlaybookPlugin */ = {
	isa = XCSwiftPackageProductDependency;
	package = 41BBDD09D038D66F59D14D11 /* XCRemoteSwiftPackageReference "Prefire" */;
	productName = "plugin:PrefirePlaybookPlugin";
};
```

### What is being validated

If you add a plugin that references a package that is not in the project:

```yaml
packages:
  Prefire:
    url: https://github.com/BarredEwe/Prefire
    from: 1.3.0
targets:
  App:
    buildToolPlugins:
      - plugin: PrefirePlaybookPlugin
        package: SomeName # Must be `Prefire`
```

There will be an error:

`Spec validation error: Plugin PrefirePlaybookPlugin has invalide package reference SomeName `

PR #1378, opened 2023-07-26
Update CI Workflows to also test Xcode 14.3.1

liamnichols:
I noticed that the CI checks only ran on Xcode 12 and 13 using macOS 11 so thought that it might be good to update this to have a sense-check using Xcode 14 as well.


Tests/Fixtures/TestProject/carthage_dynamic.xcconfig
```diff
@@ -1 +1,6 @@
 #include "xcode12_and_13_workaround.xcconfig"
+
+IPHONEOS_DEPLOYMENT_TARGET=14.0
+MACOSX_DEPLOYMENT_TARGET=10.15
+TVOS_DEPLOYMENT_TARGET=14.0
+WATCHOS_DEPLOYMENT_TARGET=7.0
```
liamnichols, line 6: Xcode 14 stopped setting default values for these settings which usually only results in warnings, but in this instance the builds were failing entirely. I came to this solution from https://stackoverflow.com/a/75729977

.github/workflows/ci.yml
```diff
@@ -4,22 +4,22 @@ on:
...
     - uses: actions/checkout@master
-    - name: Set Xcode
-      run: |
-        echo "Available Xcode versions:"
-        ls /Applications | grep Xcode
-        echo "Choosing Xcode_${{ matrix.xcode }}.app"
-        sudo xcode-select -s /Applications/Xcode_${{ matrix.xcode }}.app
-        xcodebuild -version
-        swift --version
-        swift package --version
```
liamnichols, line 22: They do the same thing, but `DEVELOPER_DIR` is slightly easier as it avoids us having to make a sudo call to `xcode-select`. 

Happy to revert this if preferred though, I originally did it when I was debugging an issue as I thought it might have been related (it wasn't).

yonaskolb: No this is great, much better. I saw it recently too in https://github.com/yonaskolb/XcodeGen/pull/1374/files

yonaskolb (approved):

liamnichols:
Thanks for the review @yonaskolb.. I just fixed the build error after i renamed a file in the last commit 🤦

yonaskolb (approved):

PR #1380, opened 2023-08-03
Add support to disable "Thread Performance Checker"

piellarda:
Lets you disable the ''Thread Performance Checker" in the run action diagnostics "Runtime API Checking" section.
Screenshot 2023-08-03 at 12 09 33


yonaskolb (commented):
Thanks @piellarda. Could you please also add an entry in CHANGELOG.md

yonaskolb (approved):

PR #1382, opened 2023-08-11
Update ProjectSpec.md to use true/false for boolean default values

kaseken:
## Changes

- Updates ProjectSpec.md.  Uses true or false instead of yes or no, for boolean default values in the Build Script options.
  - I confirmed that these values are equal to the default values defined [here](https://github.com/yonaskolb/XcodeGen/blob/master/Sources/ProjectSpec/BuildScript.swift#L5-L7) in the source code.


yonaskolb (approved):

yonaskolb:
Thanks @kaseken

PR #1395, opened 2023-08-30
Supports specifying multiple package products

simonbs:
Adds support for specifying multiple products for when adding a package dependency to a target.

With this change dependencies that were previously written as

```yaml
dependencies:
- package: FooFeature
  product: FooDomain
- package: FooFeature
  product: FooUI
```

can now be written as

```yaml
dependencies:
- package: FooFeature
  products: 
  - FooDomain
  - FooUI
```

This makes it easier to add multiple products of a single package.

### DIFF
--- a/Docs/ProjectSpec.md
+++ b/Docs/ProjectSpec.md
@@ -645,7 +645,7 @@ targets:
 ```
 
 **Package dependency**
-- [ ] **product**: **String** - The product to use from the package. This defaults to the package name, so is only required if a Package has multiple libraries or a library with a differing name
+- [ ] **product**: **String** - The product to use from the package. This defaults to the package name, so is only required if a Package has multiple libraries or a library with a differing name. Use this over `products` when you want to define different linking options per product.
 - [ ] **products**: **String** - A list of products to use from the package. This can be used when depending on multiple products from a package.
 
 ```yaml
--- a/Docs/ProjectSpec.md
+++ b/Docs/ProjectSpec.md
@@ -646,6 +646,7 @@ targets:
 
 **Package dependency**
 - [ ] **product**: **String** - The product to use from the package. This defaults to the package name, so is only required if a Package has multiple libraries or a library with a differing name
+- [ ] **products**: **String** - A list of products to use from the package. This can be used when depending on multiple products from a package.
 
 ```yaml
 packages:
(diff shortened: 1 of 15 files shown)


Docs/ProjectSpec.md
```diff
@@ -663,6 +664,21 @@ targets:
...
+packages:
+  FooFeature:
+    path: Packages/FooFeature
+targets:
+  App:
+    dependencies:
+      - package: FooFeature
+        products:
+          - FooDomain
+          - FooUI
```
freddi-kit, line 679: How can we handle if we want to do below case?

- FooDomain: embed = true
- FooUI: embed = false

simonbs: My thinking was that people would use `product` when needing fine-grained control and `products` (with an s) when products are to be added the same way.

```yaml
dependencies:
- package: FooFeature
  product: FooDomain
  embed: true
- package: FooFeature
  product: FooUI
  embed: false
```

Alternatively, products should be objects as shown below while also supporting products as plain strings. If both aren't supported, the benefit of `products` become too little.

So both of these would be valid:

```yaml
dependencies:
- package: FooFeature
  products:
  - name: FooDomain
    embed: true
  - name: FooUI
    embed: false
```

```yaml
dependencies:
- package: FooFeature
  products:
  - FooDomain
  - FooUI
```

simonbs:
> 👍
> 
> Please add some test case under Fixtures, like: https://github.com/yonaskolb/XcodeGen/blob/master/Tests/Fixtures/SPM/project.yml

Sure! I'v added that in https://github.com/yonaskolb/XcodeGen/pull/1395/commits/be92f16155eed8dbe4a6e1f68f1fe847be131852.

Docs/ProjectSpec.md
yonaskolb: @simonbs the `embed` property is part of the Dependency type, so the list of objects would break that relationship.
I think having to split out into seperate `product` is fine. Would like to see that perhaps mentioned in the docs under `product` like "Useful if you want to define different linking options per product. "

Docs/ProjectSpec.md
```diff
@@ -646,6 +646,7 @@ targets:
 
 **Package dependency**
 - [ ] **product**: **String** - The product to use from the package. This defaults to the package name, so is only required if a Package has multiple libraries or a library with a differing name
```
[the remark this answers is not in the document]
yonaskolb: I think we have to keep it around due to it's ability to set different options related to the dependency itself, like you mention below

yonaskolb:
Looks good! Thanks @simonbs!

Docs/ProjectSpec.md
simonbs: @yonaskolb That's a good idea. I've added it in https://github.com/yonaskolb/XcodeGen/pull/1395/commits/9785749ad19a107de8f4fb1a7184c0bd5c4a0c12.

yonaskolb (approved):
Thanks @simonbs!

PR #1400, opened 2023-09-07
Allow lookup template variables from globalTemplateAttributes

vikage:
For example, I have an iOS project with a network extension target. But I have some partners (Google, Facebook, etc.) and have Xcode configurations for each partner separately.
I have to create multiple xcodegen spec files to configure Google and Facebook.
Network extension reuse for every partner, then I create a file to reuse for every partner. But currently, xcodegen doesn't support a mechanism to import a config partner name from the xcodegen spec to the network extension spec.
Then I created a PR to allow xcodegen to lookup template attributes from `globalTemplateAttributes` spec.
Demo project for this situation: https://github.com/vikage/DemoXcodeGenGlobalTemplateAttribute

package-tunnel-target.yml
```
targetTemplates:
  PacketTunnelTemplate:
    platform: iOS
    type: app-extension
    settings:
      CODE_SIGN_IDENTITY: Apple Development
    dependencies:
      - sdk: NetworkExtension.framework
    configFiles:
      Debug: Partners/${partner}/PacketTunnel-Debug.xcconfig
      Release: Partners/${partner}/PacketTunnel-Release.xcconfig
    settings:
      CODE_SIGN_IDENTITY: Apple Development
    sources:
      - PacketTunnel

targets:
  PacketTunnel:
    templates:
      - PacketTunnelTemplate
 ```

google-project.yml
```
name: Demo

options:
  
globalTemplateAttributes:
  partner: Google

include:
  - packet-tunnel-target.yml

targets:
  Demo:
    type: application
    platform: iOS
    deploymentTarget: "10.0"
    sources:
      - DemoXcodeGenGlobalTemplateAttribute
 ```

### DIFF
--- a/Sources/ProjectSpec/Template.swift
+++ b/Sources/ProjectSpec/Template.swift
@@ -66,6 +66,10 @@ private func resolveTemplates(jsonDictionary: JSONDictionary, templateStructure:
             if let templateAttributes = reference["templateAttributes"] as? [String: String] {
                 reference = reference.expand(variables: templateAttributes)
             }
+
+            if let globalVariables = jsonDictionary["globalTemplateAttributes"] as? [String: String] {
+                reference = reference.expand(variables: globalVariables)
+            }
         }
         baseDictionary[referenceName] = reference
     }


vikage:
@yonaskolb Please help me review this PR

elliot-vu:
@yonaskolb Please help me review this PR. Thanks

vikage:
@yonaskolb Please help me review this PR. Thanks

yonaskolb:
I can see the utility in this @vikage. Got a few notes:
- does the existing functionality of referencing environment variables in the spec also solve your issue? eg `Config/${VERSION}/Debug.xcconfig` and `VERSION=new xcodegen generate`
- If this were added I can see this being useful just as globally accessible values, as opposed to restricting to just target or scheme templates. What do you think?
- If this were moved forward, it would definitely need some unit tests, and an example in Tests/Fixtures/TestProject/project.yml

vikage:
> I can see the utility in this @vikage. Got a few notes:
> 
> * does the existing functionality of referencing environment variables in the spec also solve your issue? eg `Config/${VERSION}/Debug.xcconfig` and `VERSION=new xcodegen generate`
> * If this were added I can see this being useful just as globally accessible values, as opposed to restricting to just target or scheme templates. What do you think?
> * If this were moved forward, it would definitely need some unit tests, and an example in Tests/Fixtures/TestProject/project.yml

@yonaskolb 
- Currently, exporting a variable on the command line could fix my issue. But I think it will be great if we have it in the yml config file.
- If this was added, we could access variables globally without any restrictions.

yonaskolb:
Yes happy for you to move it forward, with the ability to reference these values anywhere and not just within templates

vikage:
@yonaskolb Please help me take a look

vikage:
@yonaskolb Please review, thanks

Issue #251, opened 2018-02-26
Replace Make file with Beak file


yonaskolb:
https://github.com/yonaskolb/Beak

PR #1402, opened 2023-09-14
Issue #251 I have replace the MakeFile with the BeakFile

Cretus-iosDev:
#251


yonaskolb:
Hi @Cretus-iosDev. Is this a work in progress? The BeakFile doesn't bring over any of the functionality the Makefile had

Cretus-iosDev:
> Hi @Cretus-iosDev. Is this a work in progress? The BeakFile doesn't bring over any of the functionality the Makefile had

i'll make the changes and re-generate the PR

yonaskolb:
@Cretus-iosDev, appreciate the enthusiasm, but at this point I don't think migrating from make to beak is worth it. It seemed interesting at the time because I had recently built beak, but I think make serves the purpose better here, and if we used beak we'd still need a wrapper to control that dependency. I'll close that 5 year old issue

PR #1405, opened 2023-09-28
Typo in ProjectSpec.md

ahmedk92:
schems -> schemes

It ain't much, but it's honest work 😄


yonaskolb (approved):
Every bit helps, thanks @ahmedk92 🙏

PR #1410, opened 2023-10-25
Changed copy with move in writeXcodeProject. Resolves #1409

### LINKED ISSUE #1409: XcodeGen stops generating schemes when using 'preAction' in a project with lots of schemes
### Problem Description
My application has over 150 schemes, and I'm attempting to run XcodeGen as a "preAction" in a scheme to simplify the project file to include only the necessary files. Each scheme is targeted for a different client, and each client has their own set of files for customizing the app's appearance. Originally, I had these files in different targets, but due to performance issues in Xcode, I want to keep the project file as lightweight as possible.

The issue is that when running a scheme, in most cases, XcodeGen stops before completing the project generation, and only a few schemes are actually created. I've logged the output to a log file, and all it shows is:

```
⚙️  Generating plists...
⚙️  Generating project...
⚙️  Writing project...
```

### Additional Information
- **XcodeGen Version:** 2.37.0

### Steps to Reproduce
1. Create a project.yml with more than 150 schemes.
2. Set up a "preAction" in all schemes to run XcodeGen.
3. Execute a scheme.

### Expected Behavior
I expect XcodeGen to successfully generate all schemes and finish building the project without interruptions.

### Current Behavior
XcodeGen stops before completing the project generation

arkonxu:
Just changed copy with move in FileWrite: writeXcodeProject in order to fix an issue that I opened before:

[https://github.com/yonaskolb/XcodeGen/issues/1409](https://github.com/yonaskolb/XcodeGen/pull/url)


arkonxu:
I opened an issue that I could resolve with these changes

[https://github.com/yonaskolb/XcodeGen/issues/1409](url)

Sources/XcodeGenKit/FileWriter.swift
```diff
@@ -16,12 +16,11 @@ public class FileWriter {
         let tempPath = try Path.processUniqueTemporary() + "XcodeGen"
         try? tempPath.delete()
         if projectPath.exists {
-            try projectPath.copy(tempPath)
+            try projectPath.move(tempPath)
```
yonaskolb, line 19: If the project is open, this would yank the project file out from underneath Xcode, which would complain no?

arkonxu: Thank you for your concern! I've tested the changes multiple times, as I mentioned in the issue description, both by using the scheme preAction and from the terminal. It appears that there are no issues with the project file in my environment, and I haven't encountered any complains in Xcode. If you have specific test cases or scenarios in mind that you'd like me to investigate, please feel free to share them!

yonaskolb:
Any luck with this one @arkonxu? I don't have any tips, but I do have the frustration of how Xcode chooses to resolve swift packages when the project changes, and have always wondered if there was a way around that.

PR #1415, opened 2023-11-07
Remove the prefix v for the semantic version if necessary

coffmark:
## Describe your changes
- Remove the prefix v for the semantic version in SPM if necessary, as in v2.2.0.
  - Libraries that include the prefix v in tags are [Quick](https://github.com/Quick/Quick/releases), [Nimble](https://github.com/Quick/Nimble/releases), [socket.io-client-swift](https://github.com/socketio/socket.io-client-swift/releases)

## Issue ticket number and link
- None. (as far as I've seen)

## Checklist before requesting a review
- [x] I have performed a self-review of my code
- [x] I have added tests.

Thanks!

### DIFF
--- a/Docs/ProjectSpec.md
+++ b/Docs/ProjectSpec.md
@@ -1228,7 +1228,7 @@ Swift packages are defined at a project level, and then linked to individual tar
 ### Remote Package
 
 - [x] **url**: **URL** - the url to the package
-- [x] **version**: **String** - the version of the package to use. It can take a few forms:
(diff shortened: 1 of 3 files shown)


yonaskolb:
@coffmark can you help me understand the issue? How does SPM manage this when you reference a dependency with and without `v` when the tag contains a `v`?

coffmark:
Screenshot 2024-04-10 10 02 08

In SPM, it outputs an error when the tag contains a v.
I'm beginning to feel that my PR are not right.
Thank you for your comments 🙏

PR #1421, opened 2023-12-04
Support for Strings Catalogs (Xcode 15)

nicolasbosi95:
Xcode 15 introduced [String Catalogs](https://developer.apple.com/documentation/Xcode/localizing-and-varying-text-with-a-string-catalog) (`.xcstrings` files), which intend to replace the now legacy `.strings` and `stringsdict` files.

These changes add support to strings catalogs in the following ways:

- String catalogs are added in the Resources build phase.
- It is specified that the content type of string catalogs is `text.json.xcstring`.
- Locales detected in a string catalog are treated as `knownRegions` of the project. This is because string catalogs do not reside in dedicated per-locale `.lproj` directories, so detecting `knownRegions` exclusively based on existing `.lproj` directories is no longer reliable.

### DIFF
--- a/Sources/XcodeGenKit/SourceGenerator.swift
+++ b/Sources/XcodeGenKit/SourceGenerator.swift
@@ -455,6 +455,7 @@ class SourceGenerator {
 
         let createIntermediateGroups = targetSource.createIntermediateGroups ?? project.options.createIntermediateGroups
         let nonLocalizedChildren = children.filter { $0.extension != "lproj" }
+        let stringCatalogChildren = children.filter { $0.extension == "xcstrings" }
 
         let directories = nonLocalizedChildren
             .filter {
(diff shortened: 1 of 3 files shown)


Tests/Fixtures/TestProject/project.yml
```diff
@@ -163,6 +163,7 @@ targets:
         resourceTags:
           - tag1
           - tag2
+      - String Catalogs/Localizable.xcstrings
```
MojtabaHs, line 166: This seems not following the naming pattern.

MojtabaHs (commented):
Thanks for the contribution.
I'm suggesting to add more documentation about this change and also, review the naming pattern for the arguments.

yonaskolb (approved):
Fantastic, thank you @nicolasbosi95!

PR #1424, opened 2023-12-10
Update Rainbow version

nysander:
It is not possible now to use as dependencies both Genesis and XcodeGen due to conflicting Rainbow package versions (Genesis has higher one (4.0.0) and XcodeGen has lower (3.0.0)


yonaskolb (approved):
Thanks @nysander

PR #1434, opened 2023-12-31
Create swift.yml

rmtkcjs:


yonaskolb:
Hi @rmtkcjs, was this created for a certain reason? We have a workflow here https://github.com/yonaskolb/XcodeGen/blob/master/.github/workflows/ci.yml

PR #1452, opened 2024-02-20
Fix typo in README

tylermilner:
"...which lets **out** output the resolved spec..." --> "...which lets **one** output the resolved spec..."

Could also change to "...which **outputs** the resolved spec..."


yonaskolb:
Outputs sounds good to me 👍

PR #1454, opened 2024-02-22
Add support for relative paths for exclusions

dalemyers:
Includes already support relative paths, but excludes don't. This means if you did something like

```
targets:
  Foo:
    sources:
      path: ../Foo
      excludes:
        - ../Foo/Bar
```

then it wouldn't work. This fix corrects that issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Next Version
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


dalemyers:
Weird. The tests all passed locally. Maybe a weird caching thing. I'll take a look when I get a moment.

PR #1466, opened 2024-04-11
Include folder (SPM packages) in group sorting logic

jflan-dd:
Add ability to sort SPM packages within a project by allowing groupOrdering to apply to both groups and folders.

Same motivation as https://github.com/yonaskolb/XcodeGen/pull/1366


jflan-dd:
@yonaskolb what do you think of this change?

For our project we consider it pretty important to be able to sort the local SPM packages. Do you think combining them with the group sorting logic makes sense, or should they have their own dedicated logic somewhere?

yonaskolb (approved):
Looks good to me. Thanks @jflan-dd

PR #1477, opened 2024-05-26
Add Glob support for dictionary includes

Lutzifer:
This change enables the use of Glob expressions (e.g. Sources/*/Package.yml) when including a yml file:

```
include:
  - path: Sources/*/Package.yml
  - path: Sources/*/Tests.yml
  - path: Sources/*/UITests.yml
```

### DIFF
--- a/Sources/ProjectSpec/SpecFile.swift
+++ b/Sources/ProjectSpec/SpecFile.swift
@@ -1,6 +1,7 @@
 import Foundation
 import JSONUtilities
 import PathKit
+import XcodeGenCore
 import Yams
 
 public struct SpecFile {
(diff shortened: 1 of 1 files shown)


Sources/ProjectSpec/SpecFile.swift
```diff
@@ -26,22 +27,50 @@ public struct SpecFile {
...
+        
+        private init?(path: Path, dictionary: JSONDictionary) {
+            self.path = path
+            relativePaths = Self.resolveBoolean(dictionary, key: "relativePaths") ?? Include.defaultRelativePaths
+            enable = Self.resolveBoolean(dictionary, key: "enable") ?? Include.defaultEnable
+        }
+        
+        private static func includes(from array: [Any], basePath: Path) -> [Include] {
+            array.flatMap { entry -> [Include] in
+                if let string = entry as? String, let include = Include(any: string) {
```
yonaskolb, line 60: If I'm reading this right only includes defined as a dictionary will be glob enabled, whereas plain strings won't?

yonaskolb (commented):
Would love to see some tests for this in `SpecLoadingTests`

Sources/ProjectSpec/SpecFile.swift
Lutzifer: yeah, I wanted to keep the changes as isolated to my use case as possible.
But I can have a look at extending it if you want to

Lutzifer:
> SpecLoadingTests

I implemented a very basic test for this

Sources/ProjectSpec/SpecFile.swift
Lutzifer: started working on this, but still ironing out some stuff

Lutzifer: implemented

Lutzifer:
@yonaskolb Anything I can do here?

Sources/ProjectSpec/SpecFile.swift
yonaskolb: If I understand, I see you've added support for a single string include, but not arrays of strings (where this line is). In breaking 3.0 version, support for these simpler strings could probably be dropped in the future.

Also ideally the glob logic would only be in one place where currently it is split across `parse(json:)` and `includes(from:,basePath:)`

PR #1485, opened 2024-07-02
Remove GraphViz feature

### LINKED ISSUE #1462: Deintegration of GraphViz
Repository GraphViz is read only and no longer maintained. Will be this dependency removed from project? Do we have any data of how much is this feature used?

### LINKED ISSUE #1483: Build is broken on Xcode 16 Beta 1 due to GraphViz
Currently, XcodeGen can't build with Xcode 16 Beta 1.

```
/Users/giginet/Library/Developer/Xcode/DerivedData/XcodeGen-ccjxvauqmliuzaaqjslkfzkifxrg/SourcePackages/checkouts/GraphViz/Sources/GraphViz/Edge.swift:6:27 'ID' is only available in macOS 10.15 or newer
```

The following errors were caused by [GraphViz](https://github.com/SwiftDocOrg/GraphViz). However it is no longer maintained and has already become a public archive.

This dependency is only used in the `dump` command, so we can easily drop the feature. Therefore, I think it's better to remove this dependency for future maintainability. How do you think?

By the way, I guess this error is caused by a compiler bug. Therefore, it may be resolved by Swift 6 release.

giginet:
closes #1483 closes #1462

I found XcodeGen can't build with Xcode 16 and Swift 6 compilers due to GraphViz.

It might be caused by Swift compiler's bug so it will be fixed until the stable release.
However, GraphViz is no longer maintained, so I suggest dropping this dependency for future maintainability.

This PR drops `dump --type graphviz` feature.


yonaskolb (approved):

freddi-kit (approved):
I found there is graphviz in https://github.com/yonaskolb/XcodeGen/blob/f51719ce29f835eb866a6e05f633a4d55f66c0bb/Tests/Fixtures/SPM/SPM.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved#L24 but not related so much 😂

giginet:
I checked this test fixture. It depends on XcodeGenKit itself, so it's better to update them for the current situation.

PR #1489, opened 2024-07-26
Update package version

### LINKED ISSUE #1490: Error building XcodeGen as a dependency
When XcodeGen used as a dependency in the project with same dependencies package definition became invalid.
For example on the screenshot I have errors with project wich depends on the Xcodegen and Yams.
Screenshot 2024-07-26 at 15 19 08

0111b:
When XcodeGen used as a dependency in the project with same dependencies package definition became invalid.
This PR fixes this issue.

Screenshot 2024-07-26 at 15 19 08

Fixes #1490

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,26 +8,26 @@ jobs:
     name: Xcode ${{ matrix.xcode }}
     strategy:
       matrix:
-        xcode: ["14.3.1"]
+        xcode: ["15.3"]
         include:
-        - xcode: "14.3.1"
-          macos: macos-13
+          - xcode: "15.3"
+            macos: macos-14
     env:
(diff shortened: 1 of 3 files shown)


yonaskolb (commented):
What's the swift version of the package you're importing XcodeGen into? We can probably update the swift version in the xcodegen package to something like 5.9 at this point

0111b:
@yonaskolb  I`m using exactly Swift 5.9. I didn't think about backward compatibility.
Should I create a separate `Package@swift-5.9.swift` ?

yonaskolb:
At this point I'm fine with dropping older versions, so no need for `Package@swift-5.9.swift` Can you make that change here. I'll handle the release notes after merging

0111b:
@yonaskolb I've bumped swift tools version in the package manifest

0111b:
It looks like CI failed due API rate limit issue

yonaskolb (approved):
Thanks @0111b

PR #1492, opened 2024-08-07
Make asyncMap to prevent from race-condition

### LINKED ISSUE #1491: Build failure due to concurrency issue since Xcode 16 Beta 5
Screenshot 2024-08-07 at 13 16 59

Since Xcode 16 Beta 5, the compiler raises error async tasks that may cause race conditions even if in Swift 5 mode.

Currently, the codebase of XcodeGen is legacy; it almost never uses Swift Concurrency.

I'll open the PR to fix this.

giginet:
closes #1491
[Build failure due to concurrency issue since Xcode 16 Beta 5 · Issue #1491 · yonaskolb/XcodeGen](https://github.com/yonaskolb/XcodeGen/issues/1491)

The current main can't build with Xcode 16 Beta 5.
Because, `UnsafeMutableBudderPointer` is not a thread-safe, so we can't pass it to `Sendable` closure even if compiled in Swift 5 mode.

This PR resolves this issue.

This change may cause the performance, but I have no idea how much not using `ContiguousArray` will affect performance

### DIFF
--- a/Sources/XcodeGenCore/ArrayExtensions.swift
+++ b/Sources/XcodeGenCore/ArrayExtensions.swift
@@ -2,14 +2,12 @@ import Foundation
 
 public extension Array {
 
-   func parallelMap<T>(transform: (Element) -> T) -> [T] {
-       var result = ContiguousArray<T?>(repeating: nil, count: count)
-       return result.withUnsafeMutableBufferPointer { buffer in
-           DispatchQueue.concurrentPerform(iterations: buffer.count) { idx in
-               buffer[idx] = transform(self[idx])
-           }
-           return buffer.map { $0! }
+    func parallelMap<T>(transform: @Sendable (Element) -> T) -> [T] {
+       var buffer: [T] = []
+       DispatchQueue.concurrentPerform(iterations: count) { idx in
+           buffer[idx] = transform(self[idx])
        }
+       return buffer
    }
 }
 
--- a/Sources/XcodeGenCore/ArrayExtensions.swift
+++ b/Sources/XcodeGenCore/ArrayExtensions.swift
@@ -3,7 +3,7 @@ import Foundation
 public extension Array where Element: Sendable {
 
     func parallelMap<T>(transform: @Sendable (Element) -> T) -> [T] {
-      var buffer = Array<T?>(repeating: nil, count: count)
+      var buffer = ContiguousArray<T?>(repeating: nil, count: count)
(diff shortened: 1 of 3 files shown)


Sources/XcodeGenCore/Glob.swift
```diff
@@ -89,12 +89,13 @@ public class Glob: Collection {
         }
 
         let patterns = behavior.supportsGlobstar ? expandGlobstar(pattern: adjustedPattern) : [adjustedPattern]
-        
+
+        let isIncludingFiles = includeFiles
```
giginet, line 93: This closure made `@Sendable` closure, so it's better to avoid passing mutable variables

Sources/XcodeGenCore/ArrayExtensions.swift
```diff
@@ -1,15 +1,13 @@
...
-               buffer[idx] = transform(self[idx])
-           }
-           return buffer.map { $0! }
-       }
+    func parallelMap<T>(transform: @Sendable (Element) -> T) -> [T] {
+      var buffer = ContiguousArray<T?>(repeating: nil, count: count)
+      DispatchQueue.concurrentPerform(iterations: count) { idx in
+        buffer[idx] = transform(self[idx])
+      }
+      return buffer.map { $0! }
```
freddi-kit, line 10: tab/space is not unified

Sources/XcodeGenCore/ArrayExtensions.swift
```diff
@@ -1,15 +1,13 @@
...
-       return result.withUnsafeMutableBufferPointer { buffer in
-           DispatchQueue.concurrentPerform(iterations: buffer.count) { idx in
-               buffer[idx] = transform(self[idx])
-           }
-           return buffer.map { $0! }
-       }
+    func parallelMap<T>(transform: @Sendable (Element) -> T) -> [T] {
+      var buffer = ContiguousArray<T?>(repeating: nil, count: count)
+      DispatchQueue.concurrentPerform(iterations: count) { idx in
+        buffer[idx] = transform(self[idx])
```
giginet, line 8: This implementation is dangerous, it may cause a race-condition...

giginet:
This PR seems to pass the test and builds, but I'm concerned that this change is breaking thread-safe... 🤔 

The best solution is to replace the modern concurrency mechanism, but it's tough to introduce Swift Concurrency to this project.

giginet:
This issue seems to be fixed on Beta 6

PR #1498, opened 2024-08-25
Respect relativePaths in local package paths

### LINKED ISSUE #1497: Local package path used in an included file does not respect `relativePaths: true`
If I define a local package in an included file, the path of the package is evaluated relative to the main file, regardless of the `relativePaths` setting used in the `include` directive. If I use a path relative to the included file as the local package path, I get a `Spec validation error: Invalid local package`.

juri:
Before this a local package defined in an included file would not respect relativePaths: true.

This fixes #1497.

This is a breaking change because this changes how paths are resolved.

### DIFF
--- a/Sources/ProjectSpec/Project.swift
+++ b/Sources/ProjectSpec/Project.swift
@@ -241,6 +241,7 @@ extension Project: PathContainer {
             .object("aggregateTargets", AggregateTarget.pathProperties),
             .object("schemes", Scheme.pathProperties),
             .object("projectReferences", ProjectReference.pathProperties),
+            .object("packages", SwiftPackage.pathProperties)
         ]
     }
 }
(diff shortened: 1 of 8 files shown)


Sources/ProjectSpec/Project.swift
```diff
@@ -241,6 +241,7 @@ extension Project: PathContainer {
             .object("aggregateTargets", AggregateTarget.pathProperties),
             .object("schemes", Scheme.pathProperties),
             .object("projectReferences", ProjectReference.pathProperties),
+            .object("packages", SwiftPackage.pathProperties)
```
yonaskolb, line 244: Could you add support for `localPackages` as well

yonaskolb (commented):
Thank you @juri!
Could you also please add a test by adding a package under `packages` and also `localPackages` in tests/fixtures/paths_test and add an expectation to `expands directories` test in SpecLoadingTests

yonaskolb (approved):
Thanks @juri

PR #1513, opened 2024-10-14
[swift] Fix Swift 6 warnings

marcosgriselli:
Add backwards compatible Swift 6 (Swift 5 language mode) fixes to protocol conformance and non-sendable type warnings


marcosgriselli:
@yonaskolb can we get this one merged too?

yonaskolb (approved):
Thanks @marcosgriselli. Those protocol conformance warnings are generally valid, but in these cases I don't think it's likely they will be issues in the future.

PR #1514, opened 2024-10-14
Use `FOUNDATION_FRAMEWORK` on linux

marcosgriselli:
The `linux` job started failing after `ubuntu-latest` was updated
from: https://github.com/actions/runner-images/blob/ubuntu22/20240908.1/images/ubuntu/Ubuntu2204-Readme.md
to: https://github.com/actions/runner-images/blob/ubuntu22/20241006.1/images/ubuntu/Ubuntu2204-Readme.md

https://github.com/swiftlang/swift-foundation/blob/main/Sources/FoundationEssentials/PropertyList/PlistDecoder.swift#L173

"The openStep format is unsupported on this platform"

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -2,6 +2,12 @@
 
 import PackageDescription
 
+func swiftSettings() -> [SwiftSetting] {
+    return [
+        .define("FOUNDATION_FRAMEWORK", .when(platforms: [.linux])),
+    ]
+}
+
 let package = Package(
     name: "XcodeGen",
     platforms: [.macOS(.v10_13)],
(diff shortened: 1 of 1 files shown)


yonaskolb:
Ah, I was wondering about this. Did this not work, hence the close?

marcosgriselli:
@yonaskolb yea took a bit of a blind guess there before actually trying to build on linux. My assumption at the moment is that with the `ubuntu-latest` runner update from Swift 5.10.1 to Swift 6.0.1 we got the new swift foundation framework by default and it's likely failing to validate the plist xml files correctly [here](https://github.com/swiftlang/swift-foundation/blob/main/Sources/FoundationEssentials/PropertyList/PlistDecoder.swift#L298) and marking them as [`openStep`](https://github.com/swiftlang/swift-foundation/blob/main/Sources/FoundationEssentials/PropertyList/PlistDecoder.swift#L35) which doesn't seem to be supported in Linux.

PR #1517, opened 2024-10-24
Skip failing test on Linux until upstream fix

marcosgriselli:
The linux CI job started failing after ubuntu-latest was updated
from: [actions/runner-images@ubuntu22/20240908.1/images/ubuntu/Ubuntu2204-Readme.md](https://github.com/actions/runner-images/blob/ubuntu22/20240908.1/images/ubuntu/Ubuntu2204-Readme.md)
to: [actions/runner-images@ubuntu22/20241006.1/images/ubuntu/Ubuntu2204-Readme.md](https://github.com/actions/runner-images/blob/ubuntu22/20241006.1/images/ubuntu/Ubuntu2204-Readme.md)

The error message was "The openStep format is unsupported on this platform" which comes from the new [swift-foundation implementation](https://github.com/swiftlang/swift-foundation/blob/60506f3af4ea2e27cc9cb80c97f9624ea43a9e2a/Sources/FoundationEssentials/PropertyList/PlistDecoder.swift#L166)

That functionality has now been back-ported to the state it was on Swift 5.10 on this PR https://github.com/swiftlang/swift-foundation/pull/1002 so Linux test should pass whenever `ubunutu-latest` includes a swift version with the fix.

Skipping is not the most elegant solution but it's better than a 100% red CI since the tool will already fail for Swift 6+ in Linux and there's no way to workaround the problem in foundation.

### DIFF
--- a/Tests/XcodeGenKitTests/SourceGeneratorTests.swift
+++ b/Tests/XcodeGenKitTests/SourceGeneratorTests.swift
@@ -10,10 +10,7 @@ import TestSupport
 class SourceGeneratorTests: XCTestCase {
 
     func testSourceGenerator() throws {
-        #if os(Linux) && swift(<6.0.2)
-            // https://github.com/swiftlang/swift-foundation/pull/1002
-            throw XCTSkip("Skipping test on Linux until PropertyListDecoder issues are fixed.")
-        #endif
+        try skipIfNecessary()
         describe {
 
             let directoryPath = Path("TestDirectory")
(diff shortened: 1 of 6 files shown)


Tests/XcodeGenKitTests/SourceGeneratorTests.swift
```diff
@@ -9,7 +9,11 @@ import TestSupport
 
 class SourceGeneratorTests: XCTestCase {
 
-    func testSourceGenerator() {
+    func testSourceGenerator() throws {
+        #if os(Linux) && swift(<6.0.2)
```
marcosgriselli, line 13: Swift version is constrained to ensure it tries to run again when a new release is out.

marcosgriselli:
@yonaskolb let me know if you don't want to be skipping these. The alternative here is to run CI from a Linux container that has an older Swift version without this issue.

yonaskolb:
Great resolution @marcosgriselli, thank you

yonaskolb (approved):

PR #1519, opened 2024-11-07
Run preGenCommand before checking the cache

dalemyers:
The preGenCommand may generate files, change paths, etc. If we run this after checking the cache, it never gets the chance to run if the cache hasn't changed. Running it before checking the cache lets us make these changes and _then_ we can check the cache to see if we need to regenerate or not.


dalemyers:
@yonaskolb Hey, any chance of a review? This one is blocking us. Thanks!

yonaskolb (approved):
Looks good, thanks @dalemyers

PR #1531, opened 2025-02-10
add rule to copy app extensions ('.appex') to product's Plugins dir

iljaiwas:
Add a default build rule to put app extensions (.appex) into the product's "PlugIns" directory, instead of the "Resources" directory, where they won't be found by the system.


yonaskolb (approved):
Thanks @iljaiwas

PR #1534, opened 2025-03-09
Support relative paths for fileGroups in includes

shnhrrsn:
Adds support for handling paths in `fileGroups` as relative when included from subfiles.


yonaskolb (approved):
Thank you @shnhrrsn

Issue #1505, opened 2024-09-20
Updated project format for Xcode 16


CraigSiemens:
Xcode 16 has introduced a new version of the project format. I believe it's mainly to support folder references 

From the Xcode 16 release notes
https://developer.apple.com/documentation/xcode-release-notes/xcode-16-release-notes#Project-Management

> Minimize project file changes, and avoid version control conflicts with buildable folder references.
> 
> Convert an existing group to a buildable folder with the “Convert to Folder” context menu item in the Project Navigator. Buildable folders only record the folder path into the project file without enumerating the contained files. This minimizes diffs to the project when files are added and removed, and avoids source control conflicts with your team.
> 
> To use a folder as an opaque copiable resource, the default behavior before Xcode 16, uncheck the “Build Folder Contents” option in the File Inspector. (123729918)

In theory this should allow for faster project generation when using folder references since xcodegen won't need to enumerate all the files in a source folder.

It appears like `XcodeProj` has already been updated to support the new types in the file format.
- https://github.com/tuist/XcodeProj/pull/827

erneestoc:
🙏🏻 🙏🏻 🙏🏻

bartosz-pawelczyk:
@yonaskolb

piotrsmajek:
@yonaskolb

esnssr:
still no support for Xcode 16 folder format?

funclosure:
would like to have the Xcode 16 Buildable Folder support indeed, 🙏

yonaskolb:
Ok, I will try to find some time this week

r0texx:
@yonaskolb 
🙏🏻 🙏🏻 🙏🏻

esnssr:
> Ok, I will try to find some time this week

@yonaskolb thank you so much! ❤️🙏🏼

yonaskolb:
Ok, so basic support has been added in this PR https://github.com/yonaskolb/XcodeGen/pull/1541
Please give it a spin and report back there about any issues. This will need some more work to support anything but the most basic integration.

If you have mint then you can run `mint run yonaskolb/xcodegen@synced_folder` to test it out. Be aware this includes changes from master of which there are a couple of breaking fixes to relative paths in included files, so check the changelog. I'll do a release without this feature now

yonaskolb:
Released in 2.44.0

PR #1541, opened 2025-04-15
Synced folders

yonaskolb:
This adds support for Xcode 16's synchronized folders, which depending where you look have a few different names:
- synchronized root groups
- synchronized folders
- folder references
- buildable folders

For our purposes I've chosen the name `syncedFolder`, looking for feedback on this.

This initial push supports the following:
- [x] a new `syncedFolder` source type
- [x] a new `defaultSourceDirectoryType` project spec option of the same type that can be used to override the default directory type of `group`

At the moment only top level synced folders are supported. Xcode also includes a `PBXFileSystemSynchronizedBuildFileExceptionSet` which has not been added yet here either. 
There is a load of complexity that has been added to XcodeGen over the years to sources around relative paths, custom parent groups, includes and excludes, mixed target phases, and various other options. In this first pass anything beyond the basics is unsupported for these special folders.

Please test this in your own projects and give feedback about what is broken and how this should be expanded

### DIFF
--- a/Sources/XcodeGenKit/SourceGenerator.swift
+++ b/Sources/XcodeGenKit/SourceGenerator.swift
@@ -9,6 +9,7 @@ struct SourceFile {
     let fileReference: PBXFileElement
     let buildFile: PBXBuildFile
     let buildPhase: BuildPhaseSpec?
+    var synchronizedRootGroup: PBXFileSystemSynchronizedRootGroup?
 }
 
 class SourceGenerator {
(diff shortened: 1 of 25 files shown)


esnssr:
thank you so much!
i will give this a try this week, hopefully😅, and come back with feedback :)

esnssr:
for the _very_ basic use case, which is just "make all of my project's groups -> buildable folders" it works as expected, whether by specifying `syncedFolder` for `defaultSourceDirectoryType`, or on the source's path `type`.

i just tried this on a very small project i have and it worked without any issues whatsoever.

i am yet to try to mix and match with other types and see how it goes...
but so far, it gets the job done ❤️🚀

carlosefonseca:
This is looking pretty cool! Thank you! 🚀 

You said "At the moment only top level synced folders are supported.", that's why I can't (yet) specify a group as a parent of a `syncedFolder`, right?

noefroidevaux:
Thank you for the PR ❤️
Is the PR scheduled to be merged before the next release? Is there anything I can help with﻿?
[﻿﻿I tested it](https://github.com/yonaskolb/XcodeGen/issues/1505#issuecomment-2915966114), and it seems to be working great! 🚀

mredig:
Would love an ETA on this merge :D

Sources/XcodeGenKit/SourceGenerator.swift
```diff
@@ -9,6 +9,7 @@ struct SourceFile {
     let fileReference: PBXFileElement
     let buildFile: PBXBuildFile
     let buildPhase: BuildPhaseSpec?
+    var synchronizedRootGroup: PBXFileSystemSynchronizedRootGroup?
```
kirillyakimovich, line 12: `PBXFileSystemSynchronizedRootGroup` is a subclass of `PBXFileElement`, so maybe we should not introduce new var here?

Just change addition to `PBXProjGenerator`
from `let synchronizedRootGroups = sourceFiles.compactMap { $0.synchronizedRootGroup }` to
`let synchronizedRootGroups = sourceFiles.compactMap { $0. fileReference as? PBXFileSystemSynchronizedRootGroup }`

kirillyakimovich: https://github.com/yonaskolb/XcodeGen/pull/1557

yonaskolb:
I've fixed the cache from checking all children within synced folders, which should add one of the big benefits of this feature, which is cache not getting invalidated when files are added or removed within a synced folder.

On a project I tested this with I'm getting a `duplicate Info.plist` error (surprised no one else has mentioned it), which I've attempted to fix here, but it needs an upstream fix https://github.com/yonaskolb/XcodeGen/pull/1560

clausjoergensen:
Looks like `xcassets` folders are no longer included in the build targets, thus breaking pretty badly

PR #1543, opened 2025-04-19
Feature: Custom Working Directory

georgenavarro:
I wanted the ability to set the current working directory in the run action:

Screenshot 2025-04-18 at 9 17 34 PM

XcodeProj already supports this so I just added two new properties to `Scheme.Run` so that they can be passed to `XCScheme.LaunchAction` during the generation process.

### DIFF
--- a/Sources/ProjectSpec/Scheme.swift
+++ b/Sources/ProjectSpec/Scheme.swift
@@ -132,6 +132,7 @@ public struct Scheme: Equatable {
         public static let disableThreadPerformanceCheckerDefault = false
         public static let debugEnabledDefault = true
         public static let enableGPUValidationModeDefault = true
+        public static let useCustomWorkingDirectoryDefault = false
 
         public var config: String?
         public var commandLineArguments: [String: Bool]
@@ -153,6 +154,8 @@ public struct Scheme: Equatable {
         public var storeKitConfiguration: String?
         public var customLLDBInit: String?
         public var macroExpansion: String?
+        public var customWorkingDirectory: String?
+        public var useCustomWorkingDirectory: Bool
 
         public init(
             config: String? = nil,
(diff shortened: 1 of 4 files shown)


Sources/ProjectSpec/Scheme.swift
```diff
@@ -523,6 +530,8 @@ extension Scheme.Run: JSONObjectConvertible {
         }
         customLLDBInit = jsonDictionary.json(atKeyPath: "customLLDBInit")
         macroExpansion = jsonDictionary.json(atKeyPath: "macroExpansion")
+        customWorkingDirectory = jsonDictionary.json(atKeyPath: "customWorkingDirectory")
+        useCustomWorkingDirectory = jsonDictionary.json(atKeyPath: "useCustomWorkingDirectory") ?? Scheme.Run.useCustomWorkingDirectoryDefault
```
yonaskolb, line 534: We'll need to write these values in `toJSONValue` as well

Sources/XcodeGenKit/SchemeGenerator.swift
```diff
@@ -355,6 +355,8 @@ public class SchemeGenerator {
             selectedDebuggerIdentifier: selectedDebuggerIdentifier(for: schemeTarget, run: scheme.run),
             selectedLauncherIdentifier: selectedLauncherIdentifier(for: schemeTarget, run: scheme.run),
             askForAppToLaunch: scheme.run?.askForAppToLaunch,
+            customWorkingDirectory: scheme.run?.customWorkingDirectory,
+            useCustomWorkingDirectory: scheme.run?.useCustomWorkingDirectory ?? Scheme.Run.useCustomWorkingDirectoryDefault,
```
yonaskolb, line 359: We could drop the `useCustomWorkingDirectory` property entirely and use a non nil value of customWorkingDirectory for this. What are your thoughts? Do you have need to set these separately?

georgenavarro: No, I don't have that need so I'll apply your suggestions  👍🏼

georgenavarro:
@yonaskolb I've applied your suggestions and added a check to make sure that `useCustomWorkingDirectory` is properly generated when the `customWorkingDirectory` is present.

PR #1546, opened 2025-05-10
Handle major.minor for SPM

RomanPodymov:


Sources/ProjectSpec/Project.swift
```diff
@@ -193,7 +193,19 @@ extension Project {
         attributes = jsonDictionary.json(atKeyPath: "attributes") ?? [:]
         include = jsonDictionary.json(atKeyPath: "include") ?? []
         if jsonDictionary["packages"] != nil {
-            packages = try jsonDictionary.json(atKeyPath: "packages", invalidItemBehaviour: .fail)
+            packages = try jsonDictionary.json(atKeyPath: "packages", invalidItemBehaviour: .custom({ error in
```
yonaskolb, line 196: Unless I'm missing something, the fix could live directly in `SwiftPackage.VersionRequirement(jsonDictionary: JSONDictionary)`

RomanPodymov: Hello @yonaskolb 
It's possible since I'm not very familiar with the project)

yonaskolb:
This isn't strictly a bug in XcodeGen but rather in how YAML handles number literals. Doesn't mean it's not worth supporting numbers as well though

PR #1547, opened 2025-05-19
Add validation to ensure that settings.configs values are dictionaries, in order to prevent misuse

Ryu0118:
## Motivation
In our team, we previously encountered cases where `settings.configs` was misused by passing non-dictionary values.  
However, because no error was raised, so the issue went unnoticed.

## Proposed Solution
This PR adds validation to ensure that all values in `settings.configs` are dictionaries.  
For example, the following invalid configuration will now throw a `SpecParsingError.invalidConfigsFormat` during parsing:

```yaml
settings:
  configs:
    SOME_FLAG: true
```
If any non-dictionary values are found, a `SpecParsingError.invalidConfigsFormat` is thrown during parsing.

By introducing this validation, such misconfigurations can now be detected early.

### DIFF
--- a/Sources/ProjectSpec/Project.swift
+++ b/Sources/ProjectSpec/Project.swift
@@ -173,7 +173,17 @@ extension Project {
         let jsonDictionary = Project.resolveProject(jsonDictionary: jsonDictionary)
 
         name = try jsonDictionary.json(atKeyPath: "name")
-        settings = jsonDictionary.json(atKeyPath: "settings") ?? .empty
+
+        do {
+            settings = try jsonDictionary.json(atKeyPath: "settings")
+        } catch let specParsingError as SpecParsingError {
+            // Rethrow SpecParsingError to prevent misuse of settings.configs
+            throw specParsingError
+        } catch {
+            // Ignore any errors other than SpecParsingError
+            settings = .empty
+        }
+
         settingGroups = jsonDictionary.json(atKeyPath: "settingGroups")
             ?? jsonDictionary.json(atKeyPath: "settingPresets") ?? [:]
         let configs: [String: String] = jsonDictionary.json(atKeyPath: "configs") ?? [:]
--- a/Sources/ProjectSpec/Settings.swift
+++ b/Sources/ProjectSpec/Settings.swift
@@ -28,7 +28,23 @@ public struct Settings: Equatable, JSONObjectConvertible, CustomStringConvertibl
             groups = jsonDictionary.json(atKeyPath: "groups") ?? jsonDictionary.json(atKeyPath: "presets") ?? []
             let buildSettingsDictionary: JSONDictionary = jsonDictionary.json(atKeyPath: "base") ?? [:]
             buildSettings = buildSettingsDictionary
-            configSettings = jsonDictionary.json(atKeyPath: "configs") ?? [:]
+
+            guard let configSettings = jsonDictionary["configs"] as? JSONDictionary else {
+                configSettings = [:]
+                return
+            }
+
+            let invalidConfigKeys = configSettings.reduce(into: Set<String>()) { partialResult, config in
+                if !(config.value is JSONDictionary) {
+                    partialResult.insert(config.key)
+                }
+            }
+
+            guard invalidConfigKeys.isEmpty else {
+                throw SpecParsingError.invalidConfigsFormat(keys: invalidConfigKeys)
+            }
+
+            self.configSettings = try jsonDictionary.json(atKeyPath: "configs")
         } else {
             buildSettings = jsonDictionary
             configSettings = [:]
--- a/Sources/ProjectSpec/SpecParsingError.swift
+++ b/Sources/ProjectSpec/SpecParsingError.swift
@@ -15,6 +15,7 @@ public enum SpecParsingError: Error, CustomStringConvertible {
     case unknownBreakpointActionType(String)
     case unknownBreakpointActionConveyanceType(String)
     case unknownBreakpointActionSoundName(String)
+    case invalidConfigsFormat(keys: Set<String>)
 
     public var description: String {
         switch self {
@@ -46,6 +47,8 @@ public enum SpecParsingError: Error, CustomStringConvertible {
             return "Unknown Breakpoint Action conveyance type: \(type)"
         case let .unknownBreakpointActionSoundName(name):
             return "Unknown Breakpoint Action sound name: \(name)"
+        case let .invalidConfigsFormat(keys):
+            return "The value for \"\(keys.sorted().joined(separator: ", "))\" in configs must be a dictionary"
         }
     }
 }
--- a/Tests/ProjectSpecTests/InvalidConfigsFormatTests.swift
+++ b/Tests/ProjectSpecTests/InvalidConfigsFormatTests.swift
@@ -1,0 +1,35 @@
+import ProjectSpec
+import Testing
+import TestSupport
+import PathKit
+
+struct InvalidConfigsFormatTests {
+    @Test("throws invalidConfigsFormat error for non-dictionary configs entries")
+    func testNonDictionaryConfigsEntries() throws {
+        let path = fixturePath + "invalid_configs_value_non_dict.yml"
+        let expectedError = SpecParsingError.invalidConfigsFormat(keys: ["invalid_key0", "invalid_key1"])
+
+        #expect(throws: EquatableErrorBox(expectedError)) {
+            try perform(path: path)
+        }
+    }
+
+    private func perform(path: Path) throws {
+        do {
+            _ = try Project(path: path)
+        } catch let error as SpecParsingError {
+            throw EquatableErrorBox(error)
+        } catch {
+            throw error
+        }
+    }
+
+    // SpecParsingError does not conform to Equatable, so we wrap its description here
+    private struct EquatableErrorBox: Error, Equatable {
+        let description: String
+
+        init<E: Error & CustomStringConvertible>(_ error: E) {
+            self.description = error.description
+        }
+    }
+}
--- a/Tests/Fixtures/invalid_configs_value_non_dict.yml
+++ b/Tests/Fixtures/invalid_configs_value_non_dict.yml
@@ -1,0 +1,8 @@
+name: InvalidConfigsNonDict
+
+settings:
+  configs:
+    invalid_key0: value0
+    debug:
+      valid_key: value1
+    invalid_key1: value2


Tests/ProjectSpecTests/InvalidConfigsFormatTests.swift
```diff
@@ -0,0 +1,35 @@
...
+
+struct InvalidConfigsFormatTests {
+    @Test("throws invalidConfigsFormat error for non-dictionary configs entries")
+    func testNonDictionaryConfigsEntries() throws {
+        let path = fixturePath + "invalid_configs_value_non_dict.yml"
+        let expectedError = SpecParsingError.invalidConfigsFormat(keys: ["invalid_key0", "invalid_key1"])
+
+        #expect(throws: EquatableErrorBox(expectedError)) {
+            try perform(path: path)
+        }
```
giginet, line 14: How about getting error as the return value and checking this.
https://developer.apple.com/documentation/testing/require(throws:_:sourcelocation:performing:)-7n34r

Then you can get rid of `EquatableBox`.

Sources/ProjectSpec/Project.swift
```diff
@@ -173,7 +173,17 @@ extension Project {
         let jsonDictionary = Project.resolveProject(jsonDictionary: jsonDictionary)
 
         name = try jsonDictionary.json(atKeyPath: "name")
-        settings = jsonDictionary.json(atKeyPath: "settings") ?? .empty
+
+        do {
+            settings = try jsonDictionary.json(atKeyPath: "settings")
+        } catch let specParsingError as SpecParsingError {
```
freddi-kit, line 179:
```suggestion
        } catch specParsingError is SpecParsingError {
```

`is` is enough if there is no typed throw

Sources/ProjectSpec/Project.swift
```diff
@@ -173,7 +173,17 @@ extension Project {
...
+            // Rethrow SpecParsingError to prevent misuse of settings.configs
```
freddi-kit, line 180:
```suggestion
            // Re-throw `SpecParsingError` to prevent the misuse of settings.configs.
```

Sources/ProjectSpec/Project.swift
```diff
@@ -173,7 +173,17 @@ extension Project {
...
+            throw specParsingError
+        } catch {
+            // Ignore any errors other than SpecParsingError
```
freddi-kit, line 183:
```suggestion
            // Ignore all errors except `SpecParsingError`
```

Sources/ProjectSpec/SpecParsingError.swift
```diff
@@ -46,6 +47,8 @@ public enum SpecParsingError: Error, CustomStringConvertible {
             return "Unknown Breakpoint Action conveyance type: \(type)"
         case let .unknownBreakpointActionSoundName(name):
             return "Unknown Breakpoint Action sound name: \(name)"
+        case let .invalidConfigsFormat(keys):
+            return "The value for \"\(keys.sorted().joined(separator: ", "))\" in configs must be a dictionary"
```
freddi-kit, line 51:
```suggestion
            return "The value for \"\(keys.sorted().joined(separator: ", "))\" must be set in `configs`"
```

Tests/ProjectSpecTests/InvalidConfigsFormatTests.swift
```diff
@@ -0,0 +1,35 @@
+import ProjectSpec
+import Testing
+import TestSupport
+import PathKit
+
+struct InvalidConfigsFormatTests {
```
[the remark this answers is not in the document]
freddi-kit: To do it, we need to remove Xcode 15.4 check in CI

Sources/ProjectSpec/SpecParsingError.swift
```diff
@@ -15,6 +15,7 @@ public enum SpecParsingError: Error, CustomStringConvertible {
     case unknownBreakpointActionType(String)
     case unknownBreakpointActionConveyanceType(String)
     case unknownBreakpointActionSoundName(String)
+    case invalidConfigsFormat(keys: Set<String>)
```
freddi-kit, line 18: Please explain more accurate error, like `invalidConfigsMappingFormat `

Sources/ProjectSpec/Settings.swift
```diff
@@ -28,7 +28,23 @@ public struct Settings: Equatable, JSONObjectConvertible, CustomStringConvertibl
...
+                if !(config.value is JSONDictionary) {
+                    partialResult.insert(config.key)
+                }
+            }
+
+            guard invalidConfigKeys.isEmpty else {
+                throw SpecParsingError.invalidConfigsFormat(keys: invalidConfigKeys)
+            }
+
+            self.configSettings = try jsonDictionary.json(atKeyPath: "configs")
```
freddi-kit, line 47: How about having a function named `validateMappingStyleInConfig`?

giginet:
Could you also update `CHANGELOG.md`?

Tests/ProjectSpecTests/InvalidConfigsFormatTests.swift
Ryu0118: It is only available from Swift 6.1, so it can not use it.

> If you use this macro with a Swift compiler version lower than 6.1, it doesn’t return a value.

Sources/ProjectSpec/Project.swift
Ryu0118: When using `catch is SpecParsingError ` form, the error instance is not available within the block.
So, I’ll keep using let specParsingError as SpecParsingError.

Tests/ProjectSpecTests/InvalidConfigsFormatTests.swift
giginet: You can use an old API. https://developer.apple.com/documentation/testing/require(_:sourcelocation:performing:throws:)

Sources/ProjectSpec/Settings.swift
```diff
@@ -28,14 +28,32 @@ public struct Settings: Equatable, JSONObjectConvertible, CustomStringConvertibl
...
     }
 
+    private static func validateMappingStyleInConfig(jsonDictionary: JSONDictionary) throws -> [String: Settings] {
+        guard let configSettings = jsonDictionary["configs"] as? JSONDictionary else {
+            return [:]
+        }
+
+        let invalidConfigKeys = Set(
+            configSettings.filter { !($0.value is JSONDictionary) }
+            .map { $0.key }
```
giginet, line 47:
```suggestion
            .map(\.key)
```

Sources/ProjectSpec/Settings.swift
```diff
@@ -28,14 +28,32 @@ public struct Settings: Equatable, JSONObjectConvertible, CustomStringConvertibl
...
+
+            self.configSettings = try Self.validateMappingStyleInConfig(jsonDictionary: jsonDictionary)
         } else {
             buildSettings = jsonDictionary
             configSettings = [:]
             groups = []
         }
     }
 
+    private static func validateMappingStyleInConfig(jsonDictionary: JSONDictionary) throws -> [String: Settings] {
```
giginet, line 40: This method is named `validate` but it returns a subset of the given JSONDictionary, it's a bit confusing. 

How about `extractValidConfigs(from jsonDictionary: JSONDictionary)`?

giginet, line 40: Could you add doccomment?

Tests/ProjectSpecTests/InvalidConfigsFormatTests.swift
```diff
@@ -0,0 +1,18 @@
+import ProjectSpec
+import Testing
+import TestSupport
+import PathKit
+
+struct invalidConfigsMappingFormatTests {
+    @Test("throws invalidConfigsMappingFormat error for non-dictionary configs entries")
+    func testNonDictionaryConfigsEntries() throws {
```
Ryu0118, line 8:
```Suggestion
    @Test("throws invalidConfigsMappingFormat error for non-mapping configs entries")
    func testNonMappingConfigsEntries() throws {
```

giginet (approved):

freddi-kit (approved):

Ryu0118 (commented):
Currently, the following YAML is correctly detected as invalid:

```yaml
settings:
  configs:
    invalid_key: value
```

However, the following YAML is **not** detected as invalid:

```yaml
targets:
  invalid_target:
    settings:
      configs:
        invalid_key: value
```

This needs to be fixed.

Tests/ProjectSpecTests/InvalidConfigsFormatTests.swift
```diff
@@ -0,0 +1,44 @@
+import ProjectSpec
+import Testing
+import TestSupport
+import PathKit
+
+struct invalidConfigsMappingFormatTests {
+    let invalidConfigsFixturePath: Path = fixturePath + "invalid_configs"
+
+    @Test("throws invalidConfigsMappingFormat for non-mapping settings.configs entries at root level")
```
giginet, line 9: Could you make this to parameterize tests with each fixture.

Tests/ProjectSpecTests/InvalidConfigsFormatTests.swift
```diff
@@ -0,0 +1,44 @@
+import ProjectSpec
+import Testing
+import TestSupport
+import PathKit
+
+struct invalidConfigsMappingFormatTests {
```
giginet, line 6: I wonder it's better to test using `settingsGroup` https://github.com/yonaskolb/XcodeGen/blob/master/Docs/ProjectSpec.md#setting-groups

Sources/ProjectSpec/ValidSettingsExtractor.swift
```diff
@@ -0,0 +1,22 @@
+import Foundation
+import JSONUtilities
+
+/// A helper for extracting and validating the `Settings` object from a JSON dictionary.
+struct ValidSettingsExtractor {
```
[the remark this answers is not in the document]
giginet: I feel it's just enough `BuildSettingsParser`

Tests/ProjectSpecTests/InvalidConfigsFormatTests.swift
```diff
@@ -0,0 +1,43 @@
...
```
freddi-kit, line 6:
```suggestion
@Suite
struct invalidConfigsMappingFormatTests {
```

Do we need it? Or do you remove it for other reason?

Ryu0118: even if we don't apply the `@Suite`, `swift-testing` recognizes this as test suites.

> The use of the @Suite attribute is optional; types are recognized as test suites even if they do not have the @Suite attribute applied to them.

ref: https://developer.apple.com/documentation/testing/suite(_:_:)

yonaskolb:
And thank you for reviewing @freddi-kit and @giginet 🙏

PR #1548, opened 2025-05-20
Update CI equipments and drop Xcode 15 support

giginet:
Updated CI equipment to support Xcode 16.3 and dropped Xcode 15.4 from CI stacks.

Swift Testing are introduced since Xcode 16.0. We are using Spectre as a test framework, however, we want to replace them into Swift Testing gradually.

However, Xcode 15 doesn't support Swift Testing yet.
It prevents introducing Swift Testing on #1547.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,16 +4,11 @@ on:
   pull_request: {}
 jobs:
   run:
-    runs-on: ${{ matrix.macos }}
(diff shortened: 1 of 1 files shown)


freddi-kit (approved):

yonaskolb (approved):

giginet:
Thank you for the quick approval!

PR #1550, opened 2025-05-22
Address Sanitizer options in run/test schemes

hi-kumar:
Introduces new options to enable or configure various sanitizers in XcodeGen schemes. Specifically, it adds support for the following options in both the Run and Test actions of a scheme:
- Address Sanitizer
- ASan Stack Use After Return
- Thread Sanitizer
- Undefined Behavior (UB) Sanitizer

Sample result of enabling just Address-Sanitizer in a generated project:
run-scheme
test-scheme


yonaskolb (approved):
Thanks @hi-kumar

PR #1565, opened 2025-07-21
Map parallelizable true to parallelization all (#1564)

### LINKED ISSUE #1564: Test target `parallelizable` behaviour changed in 2.44.0
After updating from `2.43.0` to `2.44.0` we noticed our tests behaving differently (and failing). We narrowed it down to the scheme being generated differently when `parallelizable: true` is set on the test target.

## Expected Behaviour

The test target parallelization used to be "Enabled"

Image

## Actual Behaviour

The test target parallelization is now "Swift Testing only"

Image

CraigSiemens:
- Map `parallelizable` true to `parallelization` all to match the behaviour from Xcodeproj 8.24.3.

- Fixes #1564

### DIFF
--- a/Sources/XcodeGenKit/SchemeGenerator.swift
+++ b/Sources/XcodeGenKit/SchemeGenerator.swift
@@ -260,7 +260,7 @@ public class SchemeGenerator {
             
             return XCScheme.TestableReference(
                 skipped: testTarget.skipped,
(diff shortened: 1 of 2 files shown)


yonaskolb (approved):
Thanks @CraigSiemens. If someone needs the fine grained option, that can be added to XcodeGen later

PR #1568, opened 2025-07-23
Add Readme_CN Documentation

JamesCCoder:
1. Added README_CN.md Chinese translation of the original README.md.
2. Included a language note at the top of README.md, linking to the Chinese version.


yonaskolb:
Thank you for your effort here @JamesCCoder, but I can’t take on the maintenance burden to keep this up to date. And the only thing worse than missing documentation is incorrect documentation. People will have to use their browser or AI to translate the English version into whatever language they rely on. So with thanks, I’ll be closing this

PR #1570, opened 2025-07-24
Update ArtifactBundleGen to 0.0.8

georgenavarro:
I noticed that the artifact bundles in the release assets had multiple versions of XcodeGen instead of only the version that was just released. After some investigation I deduced that it was a bug in ArtifactBundleGen, so I submitted a [bugfix](https://github.com/freddi-kit/ArtifactBundleGen/pull/10). This PR updates to the new version with that fix.


yonaskolb (approved):
Thank you @georgenavarro. I'll update the artifact on the latest release as well

PR #1576, opened 2025-08-13
Use package name for local package fileref name

dongcarl:
Otherwise, if you git clone XcodeGen under a different directory name (say `foo`), running `swift test` will result in a diff in `Tests/Fixtures/SPM/SPM.xcodeproj/project.pbxproj` that changes the `PBXFileReference` name from `XcodeGen` to `foo`


dongcarl:
Is anything else needed here for a review?

Issue #1602, opened 2026-03-05
Synced folders: `includes` silently ignored + no deduplication across targets


AlexNsbmr:
## Summary

When using `defaultSourceDirectoryType: syncedFolder` with multi-target projects, two issues prevent proper cross-target file membership:

1. **`includes` is silently ignored for synced folders** — only `excludes` is implemented
2. **No deduplication of synced folders across targets** — same path in multiple targets creates duplicate `PBXFileSystemSynchronizedRootGroup` entries

## Environment

- XcodeGen: 2.45.2
- Xcode: 16+
- macOS: Sequoia

### Directory structure

```
DummyApp/
├── project.yml
├── App/
│   └── Source/
│       ├── Feature/
│       │   └── FeatureView.swift
│       └── Shared/
│           ├── SharedModel.swift
│           └── AnotherModel.swift
├── AppTests/
│   └── AppTests.swift
└── AppExtension/
    └── ExtensionFile.swift
```

### project.yml

```yaml
name: DummyApp
options:
  defaultSourceDirectoryType: syncedFolder
  createIntermediateGroups: true
  deploymentTarget:
    iOS: 17.0

targets:
  DummyApp:
    type: application
    platform: iOS
    sources: [App]
    settings:
      PRODUCT_BUNDLE_IDENTIFIER: com.example.dummy

  DummyTests:
    type: bundle.ui-testing
    platform: iOS
    sources:
      - AppTests
      - path: App
        includes:
          - Source/Shared/SharedModel.swift
    dependencies:
      - target: DummyApp

  DummyExtension:
    type: app-extension
    platform: iOS
    sources:
      - AppExtension
      - path: App
        includes:
          - Source/Shared/AnotherModel.swift
    settings:
      PRODUCT_BUNDLE_IDENTIFIER: com.example.dummy.extension
```

### Result

Running `xcodegen generate` produces:

- **2 "App" entries** in the project navigator (1 `PBXFileSystemSynchronizedRootGroup` per target referencing `App/`)
- **`includes` silently ignored** — all files in `App/` are included in DummyTests and DummyExtension instead of just the specified files
- **No `membershipExceptions`** generated in the pbxproj

## Issue 1: `includes` silently ignored for synced folders

**Expected:** Only `SharedModel.swift` is included in DummyTests target.  
**Actual:** ALL files in `App/` are included (the `includes` is silently ignored).

### Root cause

[...]

PR #1604, opened 2026-03-06
Fix issue 1602 - Synced folders: includes silently ignored + no deduplication across targets

4brunu:
Fixes https://github.com/yonaskolb/XcodeGen/issues/1602 — two bugs affecting syncedFolder source type:

  - includes were silently ignored: When a TargetSource with type: syncedFolder specified includes, they
   had no effect. The fix computes exception paths by walking the directory tree and adding any
  files/folders not matched by the includes patterns as exceptions on the
  PBXFileSystemSynchronizedRootGroup.
  - No deduplication across targets: When the same synced folder path was referenced by multiple
  targets, a new PBXFileSystemSynchronizedRootGroup object was created each time, resulting in duplicate
   entries. The fix caches created groups by path in syncedGroupsByPath and reuses them, merging
  explicitFolders from all referencing targets.

  Changes

  - SourceGenerator.swift: Added syncedGroupsByPath cache to deduplicate synced root groups across
  targets. Added syncedFolderExceptions(for:at:) to compute exception paths from includes/excludes
  patterns.
  - PBXProjGenerator.swift: Minor supporting change.
  - SourceGeneratorTests.swift: Added tests covering both fixes.

@AlexNsbmr can you please check if this PR fixes your issue?

Thanks

### DIFF
--- a/Tests/XcodeGenKitTests/SourceGeneratorTests.swift
+++ b/Tests/XcodeGenKitTests/SourceGeneratorTests.swift
@@ -415,6 +415,82 @@ class SourceGeneratorTests: XCTestCase {
                 try expect(exceptions.contains("Nested/b.swift")) == false
             }
 
+            $0.it("excludes entire subdirectory as single exception when no files in it are included") {
+                let directories = """
+                Sources:
+                  - a.swift
+                  - ExcludedDir:
+                    - x.swift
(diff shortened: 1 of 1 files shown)


yonaskolb (approved):
Looks good, thanks @4brunu!

PR #1607, opened 2026-03-11
Fix synced folder issues: configFiles duplicate, group insertion, and directory-level membershipExceptions

AlexNsbmr:
## Description

3 related fixes for `defaultSourceDirectoryType: syncedFolder` support.

### 1. configFiles creating duplicate groups

When using `configFiles` that reference paths inside a synced folder source, a duplicate `PBXGroup` is created alongside the `PBXFileSystemSynchronizedRootGroup`.

```yaml
options:
  defaultSourceDirectoryType: syncedFolder

configFiles:
  Debug: Sources/Config/debug.xcconfig
  Release: Sources/Config/release.xcconfig

targets:
  App:
    sources: [Sources]
```

**Root cause:** `getContainedFileReference()` always creates a `PBXGroup` hierarchy via `getGroup()`, regardless of whether the file lives inside a synced folder. Same class of issue as #1602, but for `configFiles`.

**Fix:** Return a file reference with `sourceTree: .sourceRoot` when the path is inside a synced folder, skipping group creation entirely.

### 2. Root group duplication for paths inside synced folders

When multiple targets reference a synced folder path (e.g., extension targets with `type: syncedFolder`), `getGroup()` inserts the group into `rootGroups`, causing a duplicate entry in the project navigator.

**Fix:** Skip `rootGroups` insertion when the path is inside an existing synced folder.

### 3. Directory-level membershipExceptions not working

When using `includes` on a synced folder source, `findExceptions` adds directory names to `membershipExceptions` instead of individual file paths. Xcode does **not** recursively exclude directory contents from membership exceptions, leading to:

- **"Multiple commands produce Info.plist"** build errors (e.g., `Supporting Files/` directory in exceptions doesn't exclude `Supporting Files/Info.plist`)
- **Wrong target membership** for files in non-included directories (e.g., files incorrectly added to extension targets)

```yaml
# Example: extension target cherry-picking files from shared synced folder
NotificationServiceExtension:
  sources:
    - path: Sources
      type: syncedFolder
      includes:
        - Notifications/NotificationPayload.swift
```

**Fix:** Recurse into non-included directories in `findExceptions` to list individual file paths instead of directory names.

## Tests

- Existing test: `does not create duplicate group for configFiles inside synced folder`
- Existing tests for synced folder `includes` cover the membershipExceptions behavior

## Environment

- XcodeGen: 2.45.3
- Xcode: 16
- macOS: Sequoia

### DIFF
--- a/Sources/XcodeGenKit/SourceGenerator.swift
+++ b/Sources/XcodeGenKit/SourceGenerator.swift
@@ -208,7 +208,7 @@ class SourceGenerator {
         let parentPath = path.parent()
 
         guard !isInsideSyncedFolder(path: path) else {
-            return getFileReference(path: path, inPath: project.basePath, sourceTree: .sourceRoot)
+            return getFileReference(path: path, inPath: basePath, sourceTree: .sourceRoot)
         }
 
         let fileReference = getFileReference(path: path, inPath: parentPath)
@@ -293,7 +293,7 @@ class SourceGenerator {
     /// Checks the project spec directly because configFiles are resolved before target sources
     /// populate `syncedGroupsByPath`.
     private func isInsideSyncedFolder(path: Path) -> Bool {
-        let relativePath = (try? path.relativePath(from: project.basePath)) ?? path
+        let relativePath = (try? path.relativePath(from: basePath)) ?? path
         return project.targets.contains { target in
             target.sources.contains { source in
                 let type = source.type ?? (project.options.defaultSourceDirectoryType ?? .group)
--- a/Sources/XcodeGenKit/SourceGenerator.swift
+++ b/Sources/XcodeGenKit/SourceGenerator.swift
@@ -200,11 +200,12 @@ class SourceGenerator {
         let createIntermediateGroups = project.options.createIntermediateGroups
 
         let parentPath = path.parent()
-        let fileReference = getFileReference(path: path, inPath: parentPath)
 
         guard !isInsideSyncedFolder(path: path) else {
-            return fileReference
+            return getFileReference(path: path, inPath: project.basePath, sourceTree: .sourceRoot)
         }
+
+        let fileReference = getFileReference(path: path, inPath: parentPath)
 
         let parentGroup = getGroup(
             path: parentPath,
@@ -368,7 +369,7 @@ class SourceGenerator {
             groupReference = addObject(group)
             groupsByPath[path] = groupReference
 
-            if isTopLevelGroup {
+            if isTopLevelGroup && !isInsideSyncedFolder(path: path) {
                 rootGroups.insert(groupReference)
             }
         }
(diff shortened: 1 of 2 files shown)


AlexNsbmr:
@4brunu or @yonaskolb could you take a look on my PR when you have some time?

Sources/XcodeGenKit/SourceGenerator.swift
```diff
@@ -283,6 +289,19 @@ class SourceGenerator {
         }
     }
 
+    /// Whether the given path falls inside a target source configured as a synced folder.
+    /// Checks the project spec directly because configFiles are resolved before target sources
+    /// populate `syncedGroupsByPath`.
+    private func isInsideSyncedFolder(path: Path) -> Bool {
```
yonaskolb, line 295: I'm concerned about the performance implications of such a loop

Sources/XcodeGenKit/SourceGenerator.swift
```diff
@@ -206,7 +206,13 @@ class SourceGenerator {
         let createIntermediateGroups = project.options.createIntermediateGroups
 
         let parentPath = path.parent()
+
+        guard !isInsideSyncedFolder(path: path) else {
+            return getFileReference(path: path, inPath: project.basePath, sourceTree: .sourceRoot)
```
yonaskolb: we have a new `basePath` getter that may be a better value, as xcodegen can be run in a different directory than the project base path.

AlexNsbmr: You're right, I did't see it.
I just changed with `basePath`, it's way more cleaner

Sources/XcodeGenKit/SourceGenerator.swift
AlexNsbmr: Fair point. In practice it's pretty small I think because it is only called from two cold paths (`getContainedFileReference` for configFiles and the `getGroup` root insertion guard), not per-file.

But I can easily cache it with a lazy, it could look like this:

  ```swift
  private lazy var syncedFolderPrefixes: Set<String> = {
      var prefixes = Set<String>()
      for target in project.targets {
          for source in target.sources {
              let type = source.type ?? (project.options.defaultSourceDirectoryType ?? .group)
              if type == .syncedFolder {
                  prefixes.insert(source.path + "/")
              }
          }
      }
      return prefixes
  }()

  private func isInsideSyncedFolder(path: Path) -> Bool {
      let relativePath = (try? path.relativePath(from: basePath)) ?? path
      return syncedFolderPrefixes.contains { relativePath.string.hasPrefix($0) }
  }
```

Want me to go with that?

yonaskolb: It's ok. As you mention - not in a hot path 👍

yonaskolb (approved):
Thanks @AlexNsbmr

PR #1609, opened 2026-03-19
fix: include SettingPreset in Swift package

maikelvdh:
SPM is copying the symlink `Sources/XcodeGenKit/SettingPresets -> ../../SettingPresets/` as-is into the bundle, creating a broken symlink effectively. Therefore turning the symlink directive around which should make it work for all supported approaches as Mint/Make seem to follow symlink by usage of `cp -R`.


yonaskolb:
Which issue are you trying to solve? In what scenario is the current setup broken?

maikelvdh:
> Which issue are you trying to solve? In what scenario is the current setup broken?

The issue is that the installed SPM package through usage of mise (ie. https://mise.jdx.dev/dev-tools/backends/spm.html) results incorrect the `SettingPresets`. Effectively mise is using following https://github.com/jdx/mise/blob/main/src/backend/spm.rs#L190 to build the package with Swift.

maikelvdh:
@yonaskolb any feedback on above? 🙇

PR #1616, opened 2026-04-16
Fix membershipExceptions for localized files in synced folders

AlexNsbmr:
## Description

- Localized files (`.strings` inside `.lproj` directories) in synced folders were not properly excluded via `membershipExceptions` when using `includes` filters
- Xcode expects a `/Localized:` prefix format (e.g., `/Localized: Resources/AppShortcuts.strings`) for variant-group membership exceptions, but XcodeGen was generating individual per-language paths (e.g., `Resources/de.lproj/AppShortcuts.strings`)
- These per-language paths were silently ignored by Xcode, causing localized resources to leak into targets that should have excluded them

## Fix

In `configureMembershipExceptions()`, paths containing `.lproj` directory components are now transformed into the `/Localized:` variant-group format that Xcode expects.

**Before** (ignored by Xcode):
```
Resources/de.lproj/AppShortcuts.strings,
Resources/fr.lproj/AppShortcuts.strings,
```

**After** (honored by Xcode):
```
"/Localized: Resources/AppShortcuts.strings",
```

## How to reproduce

1. Create a project with `defaultSourceDirectoryType: syncedFolder`
2. Add a target that references a synced folder with `includes` filtering only specific files
3. The synced folder contains `.lproj` directories with `.strings` files
4. After generation, `.strings` files appear as members of the target despite not being in `includes`
5. This causes build warnings like "This phrase is not used in any App Shortcut" for `AppShortcuts.strings` being processed by unrelated targets

### DIFF
--- a/Sources/XcodeGenKit/PBXProjGenerator.swift
+++ b/Sources/XcodeGenKit/PBXProjGenerator.swift
@@ -1491,6 +1491,7 @@ public class PBXProjGenerator {
         var exceptions: Set<String> = Set(
             sourceGenerator.syncedFolderExceptions(for: targetSource, at: syncedPath)
                 .compactMap { try? $0.relativePath(from: syncedPath).string }
+                .map { syncedFolderLocalizedMembershipException($0) }
         )
 
         for infoPlistPath in Set(infoPlistFiles.values) {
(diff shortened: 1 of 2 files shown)


AlexNsbmr:
@yonaskolb Would you mind taking a look at this PR when you get a chance? Thanks in advance!

PR #1618, opened 2026-04-16
Fix: Exclude default ignored files and xcconfigs from synced folder membership

4brunu:
This PR fixes an issue where system files (like `.DS_Store`) and configuration files (like `.xcconfig`) were being incorrectly added to target membership when using the `syncedFolder` source type.

In `SourceGenerator.swift`, the `isDefaultExcluded` check was remove, because it was causing default excluded files to be skipped instead of being added to the `membershipExceptions` set.

Also added unit tests to verify that this works.


4brunu:
@yonaskolb could you please help to merge and release this fix? Thanks :)

4brunu:
@yonaskolb could you please take a look at this PR? Thanks

PR #1626, opened 2026-06-29
Add Swift Package Registry (SE-0292) configuration support

bdolewski:
## Motivation

Swift Package Registry (SE-0292) dependencies are resolved by SwiftPM through a `registries.json` configuration file. Xcode has no project-level (pbxproj) representation for registry packages, so today there is no way to make an XcodeGen-generated project resolve `.package(id:)` dependencies from a registry without hand-placing `registries.json` after every `xcodegen generate` (and re-placing it whenever the project is regenerated).

This PR lets XcodeGen own that configuration.

## What it does

Adds a top-level `registries` key to `project.yml`:

```yaml
registries:
  default: https://tuist.dev/api/registry/swift
  scopes:
    acme:
      url: https://packages.example.com/artifactory/api/swift/swift-registry
      supportsAvailability: false
```

XcodeGen serializes this into the generated project's `project.xcworkspace/xcshareddata/swiftpm/configuration/registries.json` in SwiftPM's schema (sorted keys, unescaped slashes, empty `authentication`, `version: 1`). A registry value may be a bare URL string or a `{ url, supportsAvailability }` mapping.

Dependencies themselves are still declared the existing, supported way: a local Swift package whose `Package.swift` uses `.package(id:)`, referenced from `project.yml` as a `local:` package. XcodeGen does not write registry references into the `.xcodeproj`, because Xcode has no pbxproj representation for them. Authentication credentials are never written (configure them via `.netrc`/keychain as usual).

## Why not a pbxproj registry reference?

I verified on Xcode 26.5 that a project-level `XCRemoteSwiftPackageReference` with `kind = registry` is silently ignored by `xcodebuild` (resolves zero packages, no error), consistent with Apple Feedback FB14699700 ("not supported in Xcode directly"). The supported mechanism is `registries.json` plus SwiftPM resolution, which is what this PR generates.

## Validation

- `swift test` passes, including new parsing and serialization tests.
- End to end: a generated iOS app consuming a dependency from the public Tuist registry (`https://tuist.dev/api/registry/swift`) builds with vanilla `xcodebuild build -destination 'generic/platform=iOS Simulator'` (BUILD SUCCEEDED); the dependency is fetched from the registry, not from git.

## Changes

- New `registries` project spec model with parsing and serialization.
- `FileWriter` writes `registries.json` after the project is written (no-op when unset or unchanged).
- Docs (`Docs/ProjectSpec.md`) and a CHANGELOG entry.

## Notes

- No breaking changes; `registries` is optional.
- No new dependencies and no change to tuist/XcodeProj.
- Possible follow-ups (separate PRs): `mirrors.json` generation; SCM-to-registry transformation scheme/build settings.

This change was developed with assistance from Claude Code (Anthropic). Happy to adjust the `registries` key name or shape to your preference.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Next Version
 
+### Added
+- Add `registries` to generate a Swift Package Registry (SE-0292) `registries.json` into the project's SwiftPM configuration, so generated projects can resolve `.package(id:)` dependencies from a registry @bdolewski
+
 ## 2.45.4
 
 ### Fixed
--- a/Docs/ProjectSpec.md
+++ b/Docs/ProjectSpec.md
@@ -51,6 +51,7 @@ You can also use environment variables in your configuration file, by using `${S
   - [Test Plan](#test-plan)
 - [Scheme Template](#scheme-template)
 - [Swift Package](#swift-package)
+- [Swift Package Registries](#swift-package-registries)
   - [Remote Package](#remote-package)
   - [Local Package](#local-package)
 - [Project Reference](#project-reference)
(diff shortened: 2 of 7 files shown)


wkornewald:
I think this won't work when building an XCFramework. The wrapper project is just a dependency and the xcodeproj itself doesn't contain any source code in this case, so the resulting XCFramework would be basically empty. Also, I tested at least with Xcode 26.4.1 and when setting up the dependencies in the UI this adds proper XCRemoteSwiftPackageReference and XCSwiftPackageProductDependency sections and a registries.json. With those, the xcodebuild archive command works without any problems. @bdolewski are you sure your pbxproj was really complete and you haven't overlooked anything in your test? Or maybe only 26.5 has a bug that doesn't exist in 26.4.1 and possibly 26.6?

PR #1642, opened 2026-08-22
Add scheme build architecture overrides

### LINKED ISSUE #1524: Add support for `buildArchitectures` of Scheme's Build action
- https://developer.apple.com/documentation/xcode-release-notes/xcode-15_3-release-notes
  - > Schemes provide a new “Override Architectures” build option, which controls the set of architectures that will be built for all targets in the workspace, including Swift packages. The recommended option (and default for new schemes) is “Match Run Destination”. Full details are available by clicking the information icon next to the “Override Architectures” setting on the scheme build options sheet in Xcode. (66146584)
- https://github.com/tuist/XcodeProj/releases/tag/8.22.0
  - > Add `buildArchitectures` to XCScheme+BuildAction

arhxam:
## Summary

- expose Xcode's scheme-level architecture override through `build.buildArchitectures`
- support `matchRunDestination`, `universal`, and `useTargetSettings` in parsing, encoding, generation, and documentation
- preserve XcodeGen's existing `useTargetSettings` behavior when the option is omitted

Resolves #1524.

## Test plan

- `swift test` (76 tests passed)
- `swift test --filter SchemeGeneratorTests/testSchemes`
- `swift test --filter SpecLoadingTests`
- generated a project with all three modes and verified the resulting scheme XML uses `Automatic`, `All`, and no override attribute, respectively

This pull request was prepared with assistance from OpenAI Codex. I reviewed and tested the changes locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Added
 - Added support for Swift package `traits` on remote and local package references, so generated projects preserve the configured trait selections #1629 @philprime
+- Added scheme build architecture overrides #1642 @arhxam
 
 ### Changed
 - Targets in the generated project now follow the declaration order from the source spec (Xcode sidebar, `xcodebuild -list` output). Previously they were always sorted alphabetically. Applies to both YAML and JSON specs. Declaration order is now also preserved for targets whose `platform`/`name` come from a target template and for targets whose key is a `${VARIABLE}`. #1619 @mirkokg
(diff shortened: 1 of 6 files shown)


yonaskolb:
Thanks @arhxam! I've pushed a couple of commits on top of this:

- **Default is now `matchRunDestination`**. This matches Xcode's default for new schemes since Xcode 15.3, and fixes problems like the SwiftUI previews linking issue mentioned in #1524. Set `buildArchitectures: useTargetSettings` to keep the previous behaviour.
- **Target schemes**: `buildArchitectures` can now also be set in a target's `scheme`, so schemes generated from targets can opt out of the new default too.
- **Docs**: each value is documented along with its name in Xcode's scheme editor ("Match Run Destination", "Universal", "Use Target Settings").
- **Changelog**: moved the entry into a new `Next Version` section, and added a `Changed` entry for the new default.
- **Fixtures**: regenerated, so generated schemes now include `buildArchitectures = "Automatic"`.

The second commit only restores a fixture file I accidentally included in the first one.

PR #1644, opened 2026-08-31
Document how to add capability

### LINKED ISSUE #1574: Capabilities - In-App Purchase
Hi! 
I'm struggling with adding the "In-App Purchase" capability to the project file. 
As far as I know it's not the feature we can add to `.entitlements` file. 
```
<key>com.apple.developer.in-app-purchase</key>
<true/>
```
☝️ this doesn't work

I tried adding it to `project.yml` config file
``` 
capabilities:
   inAppPurchase: {}
```
☝️ also doesn't work

Any ideas?

Hokila:
## Summary
- document how to investigate an Xcode capability by inspecting the generated project and entitlements changes
- add the confirmed In-App Purchase `StoreKit.framework` dependency example
- clarify that XcodeGen does not provide a general-purpose capability abstraction

Resolves #1574

## Validation
- `git diff --check`
- `swift test`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Change Log
 
+## Unreleased
+
+### Documentation
+- Document how to add the In-App Purchase capability #1644 @Hokila
+
 ## 2.46.0
 
 ### Added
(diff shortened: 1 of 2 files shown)


yonaskolb:
Thanks @Hokila

